PR #3, opened 2015-01-15
architecture changes to dropwizard guidelines

sqreek:
architecture changes to dropwizard guidelines
using HalBuilder
create comments entry point
most important: add banner

question:
What kind of pojo file we want to use?

muttable or immutable?
with builder (inside, or outside in factory) or without builder

check UriBuilder <- its hack, after few hours, when im looking at it right now, i start to dislike it.

ArticleResource -> Creation response with HalBuilder is preaty easy, and we dont event need to have pojo here.


src/main/java/com/wikia/pandora/PandoraApplication.java
```diff
@@ -1,48 +1,56 @@
...
 public class PandoraApplication extends Application<PandoraConfiguration> {
-    public static void main(String[] args) throws Exception {
-        new PandoraApplication().run(args);
-    }
 
-    @Override
-    public String getName() {
-        return "Pandora";
-    }
+  private static RepresentationFactory representationFactory;
```
drsnyder: Why can't this be created in `initialize` or `run` and then passed to the resources that need it? Why does this need to be a static member of the `PandoraApplication`?

src/main/java/com/wikia/pandora/service/ArticleService.java
```diff
@@ -1,29 +1,17 @@
...
-import java.io.IOException;
-import java.util.Map;
+import com.wikia.pandora.core.Article;
+import com.wikia.pandora.core.Comment;
 
-public class ArticleService {
+import java.util.List;
 
-    private final MercuryGateway mercuryGateway;
+public interface ArticleService {
```
drsnyder: I think this is a good idea. It seems like part of what you are suggesting is that we more strictly program to abstractions over concrete class.

src/main/java/com/wikia/pandora/service/mercury/MercuryService.java
```diff
@@ -0,0 +1,12 @@
+package com.wikia.pandora.service.mercury;
+
+public class MercuryService {
+
+  protected MercuryGateway mercuryGateway;
```
drsnyder: Should all of these be `final`?

src/main/java/com/wikia/pandora/util/UriBuilder.java
```diff
@@ -0,0 +1,59 @@
...
+          .path(method)
+          .build(path)
+          .getPath();
+      return uri;
+    } catch (Exception e) {
+      e.printStackTrace();
+
+    }
+    return "";
+  }
```
drsnyder: It seems like you are eschewing the magic provided by `Halarious` and it's decorators in favor of one we build our own. Do you see anything wrong with building the `self` URI manually?

src/main/resources/banner.txt
```diff
@@ -0,0 +1,17 @@
...
+       W:::::W:::::W   W:::::W:::::W         I::::I    K:::::K K:::::K     I::::I       A:::::::::::::::::::::A               A:::::::::::::::::::::A       P::::P              I::::I
+        W:::::::::W     W:::::::::W          I::::I  KK::::::K  K:::::KKK  I::::I      A:::::AAAAAAAAAAAAA:::::A             A:::::AAAAAAAAAAAAA:::::A      P::::P              I::::I
+         W:::::::W       W:::::::W         II::::::IIK:::::::K   K::::::KII::::::II   A:::::A             A:::::A           A:::::A             A:::::A   PP::::::PP          II::::::II
+          W:::::W         W:::::W          I::::::::IK:::::::K    K:::::KI::::::::I  A:::::A               A:::::A         A:::::A               A:::::A  P::::::::P          I::::::::I
+           W:::W           W:::W           I::::::::IK:::::::K    K:::::KI::::::::I A:::::A                 A:::::A       A:::::A                 A:::::A P::::::::P          I::::::::I
+            WWW             WWW            IIIIIIIIIIKKKKKKKKK    KKKKKKKIIIIIIIIIIAAAAAAA                   AAAAAAA     AAAAAAA                   AAAAAAAPPPPPPPPPP          IIIIIIIIII
```
drsnyder: How could we have forgotten that!

src/main/java/com/wikia/pandora/service/mercury/MercuryArticlesService.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Map;
+
+public class MercuryArticlesService extends MercuryService implements ArticleService {
+
+  public MercuryArticlesService(MercuryGateway mercuryGateway) {
+    super(mercuryGateway);
+
+  }
+
+  public Article getArticleByTitle(String wikia, String title) {
```
drsnyder: :+1: Good call on changing the name. I wasn't fond of `lookup` either.

src/main/java/com/wikia/pandora/service/mercury/MercuryArticlesService.java
```diff
@@ -0,0 +1,74 @@
...
+      article.setTitle((String) detailsMap.get("title"));
+      article.setContent((String) articleMap.get("content"));
+      return article;
+    } catch (IOException e) {
+      throw new NotFoundException("No such article");
+    }
+  }
+
+  @Override
+  public List<Comment> getArticleComments(String wikia, String title) {
```
drsnyder: I favor erring on the side of immutable as a rule rather than an exception. In this case I can't see why a caller would need to modify the maps or lists returned from MediaWiki.

drsnyder:
Unless I'm mistaken, there are three main themes here.

 1. Use HalBuilder over Halarious and construct the response using `withBean`. Does the HalBuilder Representation abstraction support classes with final properties instead of getters?
 2. Program to abstractions as a rule
 3. Immutable objects-- both beans and lists, sets, maps

For my part, I'm neutral on HalBuilder vs Halarious. I think we would both agree that decoupling the representation from the `*Resource` classes is where we want to go it's just how to get there.

I would support both 2 and 3.

/cc @nandy-andy @pchojnacki @nmonterroso

src/main/resources/banner.txt
nandy-andy: I blame Dropwizard - it yells at you when you don't implement any health checks but stays silent when you forget about a banner :smirk:

nandy-andy:
I agree :smile_cat: And AFAIK @sqreek started working on decoupling, so probably we will see it here soon ;) I will apply those rules in mobile config service today.

Something connected to both branches: why are we using Mercury API written in PHP? Shouldn't we "copy" solutions from there and send requests to MediaWiki? Of course in future we might replace the MW API calls and abstraction would most likely help us with it.

build.gradle
```diff
@@ -21,6 +21,16 @@ dependencies {
     compile 'io.dropwizard:dropwizard-client:0.7.1'
     compile 'ch.halarious:halarious-core:0.11'
 
+    //HAL builder
+    compile 'com.theoryinpractise:halbuilder-api:4.0.1'
+    compile 'com.theoryinpractise:halbuilder-core:4.0.3'
+    compile 'com.theoryinpractise:halbuilder-json:4.0.2'
+    compile 'com.theoryinpractise:halbuilder-standard:4.0.1'
```
nandy-andy: If you leave only this package it will download the three above as well. So, `halbuilder-standard` and `halbuilder-jaxrs` are enough.

But I might be wrong and maybe it is a good practice to be more explicit here. It also gives us possibility to manually set versions.

src/main/java/com/wikia/pandora/service/mercury/MercuryGateway.java
```diff
@@ -0,0 +1,154 @@
...
+
+  private static final Logger logger = LoggerFactory.getLogger(MercuryGateway.class);
+
+  public static final String
+      DEFAULT_MERCURY_COMMENT_REQUEST_FORMAT =
+      "http://%s.wikia.com/api/v1/Mercury/ArticleComments?title=%s";
+
+  public static final String
+      DEFAULT_MERCURY_ARTICLE_REQUEST_FORMAT =
+      "http://%s.wikia.com/api/v1/Mercury/Article?title=%s";
```
nandy-andy: Let's use http://www.wikia.com/api/v1/#!/Articles

src/main/java/com/wikia/pandora/service/mercury/MercuryGateway.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Map;
+
+///TODO refactor this class
+public class MercuryGateway {
+
+  private static final Logger logger = LoggerFactory.getLogger(MercuryGateway.class);
+
+  public static final String
+      DEFAULT_MERCURY_COMMENT_REQUEST_FORMAT =
+      "http://%s.wikia.com/api/v1/Mercury/ArticleComments?title=%s";
```
nandy-andy: Just for reference: [MercuryApi uses Nirvana controller](https://github.com/Wikia/app/blob/dev/extensions/wikia/MercuryApi/MercuryApiController.class.php#L183) and [its method](https://github.com/Wikia/app/blob/dev/extensions/wikia/ArticleComments/modules/ArticleCommentsController.class.php#L148) to get comments and then [it processes the response](https://github.com/Wikia/app/blob/dev/extensions/wikia/MercuryApi/models/MercuryApi.class.php#L150).

src/main/java/com/wikia/pandora/util/UriBuilder.java
sqreek: its come from jersey.
more manually -> more maintance
more maintance -> more sadness

UriBuilder is good, for making URI, but this aproach is a little too risky to handle it.
For example, this function can be called exacly from method with entry point
I will fix this later

sqreek:
@drsnyder 
Why we don't want use getters?

Program to abstractions as a rule
+1

Immutable objects-- both beans and lists, sets, maps
+1

drsnyder:
> Why we don't want use getters?

It's fine. It just seems unnecessary if your object is immutable. If you have `public final foo;` you don't need a getter. This is how I'm used to programming in clojure. I'm not sure yet how it maps to Java but it's something I want to explore. It has been my experience that when you can't mutate objects they tend to be a lot easier to reason about.

I expect that a lot of what we will be doing in Pandora will lend itself well to this convention. We'll mostly be doing data transformation from one interface/data format to another.

src/main/java/com/wikia/pandora/service/mercury/MercuryGateway.java
drsnyder: It's my understanding that Mercury has a different render type. They elide the images in the article payload and the include them separately in a list so that they can lazy load them.

I imagine this will be an input that we'll have to support-- that is we need to allow the caller to specify the render type on request.

sqreek:
With no red/yellow lights

PR #6, opened 2015-01-26
Mercury api replacement

sqreek:
Added mediawiki-fluent-api module
Added mercury-api module
Added new domains item
Added Category Service

Pls check: 
Mediawiki gateway for using mediawiki-fluent-api
mercury-api for custom data aproach, and how to used other resources for making embeded
core.domains, is new domains is ok? Article/WithDescriptionWithContent/

not finished yet!


mercury-api/src/main/java/com/wikia/pandora/core/domains/MercuryArticle.java
```diff
@@ -0,0 +1,44 @@
+package com.wikia.pandora.core.domains;
```
drsnyder: Should this be singular `domain`?

mercury-api/src/main/java/com/wikia/pandora/core/domains/MercuryArticle.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+public class MercuryArticle {
+
+  private String content;
```
drsnyder: `private final String`?

mercury-api/src/main/java/com/wikia/pandora/core/domains/MercuryArticleData.java
```diff
@@ -0,0 +1,44 @@
+package com.wikia.pandora.core.domains;
+
+import java.util.List;
+
+public class MercuryArticleData {
```
drsnyder: The naming seems confusing or too similar to the `MercuryArticle` above.

mercury-api/src/main/java/com/wikia/pandora/resources/HalMercuryResource.java
```diff
@@ -0,0 +1,56 @@
...
+import com.theoryinpractise.halbuilder.api.RepresentationFactory;
+import com.wikia.pandora.api.service.ArticleService;
+import com.wikia.pandora.core.util.UriBuilder;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+
+@Path("{wikia}/MercuryArticles")
```
drsnyder: I don't think this belongs as a separate resource. Mercury "articles" differ from desktop articles, but they aren't a separate resource but a different representation or transformation of the same resource.

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mediawiki/MediawikiArticleService.java
```diff
@@ -1,25 +1,59 @@
...
+import com.wikia.pandora.core.domains.ArticleWithDescription;
 import com.wikia.pandora.core.domains.Comment;
+import com.wikia.pandora.core.domains.builder.PojoBuilderFactory;
+
+import org.apache.commons.lang.NotImplementedException;
 
 import java.util.List;
 
-public class MediawikiArticleService implements ArticleService {
+public class MediawikiArticleService extends MediawikiService implements ArticleService {
```
drsnyder: It seems odd to me to have a service under the gateway. When I think of service I think of the glue between different gateways or data sources.

drsnyder:
This PR should be reduced to only include the changes to add the Mercury gateway. In general PR's should be closer to atomic (when possible) then wide and broad changes. 

I know that can be difficult at this stage but we should strive for smaller change sets.

mercury-api/src/main/java/com/wikia/pandora/core/domains/MercuryArticle.java
sqreek: true

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mediawiki/MediawikiArticleService.java
sqreek: Every Mediawiki*Service have a gateway, so MediawikiService is a base class for all Mediawiki*Services with standard gateway inside

drsnyder:
We also need to start including the [Javadocs](http://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s7-javadoc) on classes and methods.

drsnyder:
I think I have been struggling to find the words to communicate the approach that I would like for us to take. Although I'm not a disciple of extreme programming I think some of the ideas and values (they call them rules) that came from the system are good.

For example, one of the rules is [Simplicity is the Key](http://www.extremeprogramming.org/rules/simple.html). From that rule comes the recommendation to *always do the simplest thing that could possibly work*. I think this is a question that we need to keep asking ourselves throughout this effort.

In this case, what is the simplest thing that will work for adding a Mercury gateway? Let's try to find that while also adding tests and documentation. When it's working, we can come back and refactor if necessary.

mediawiki-fluent-api/src/main/java/com/wikia/mwapi/MWApi.java
```diff
@@ -0,0 +1,183 @@
...
+  private String list;
+
+
+  protected MWApi() {
+    this(HttpClientBuilder.create().build());
+  }
+
+  public MWApi(HttpClient httpClient) {
+    this.httpClient = httpClient;
+    baseUrl = "http://%s.wikia.com/api.php?%s";
```
drsnyder: Should this be injected (optionally with a sane default)? I think it's likely to change once we have varnish in front of MW internally.

mediawiki-fluent-api/src/main/java/com/wikia/mwapi/MWApi.java
```diff
@@ -0,0 +1,183 @@
...
+  @Override
+  public OptionChoose rvContent() {
+    rvProp = CONTENT;
+    return this;
+  }
+
+  private String buildUrl() {
+    URIBuilder b = null;
+    try {
+      b = new URIBuilder(String.format("http://%s.wikia.com/api.php", wikia));
```
drsnyder: Should this have `baseUrl`?

mediawiki-fluent-api/src/main/java/com/wikia/mwapi/MWApi.java
```diff
@@ -0,0 +1,183 @@
...
+  public ApiResponse get() {
+    ApiResponse apiResponse = null;
+    String url = buildUrl();
+
+    HttpUriRequest request = new HttpGet(url);
+    try {
+      HttpResponse response = httpClient.execute(request);
+      ObjectMapper mapper = new ObjectMapper();
+      apiResponse =
+          mapper.readValue(response.getEntity().getContent(), ApiResponse.class);
```
drsnyder: How does this work? `ApiResponse` doesn't have any setters.

pandora-examples/src/main/java/com/wikia/pandora/examples/PandoraApplication.java
```diff
@@ -35,10 +40,19 @@ public void run(PandoraConfiguration configuration, Environment environment) {
     final PandoraHealthCheck healthCheck = new PandoraHealthCheck();
     environment.healthChecks().register("pandora", healthCheck);
 
-    ServiceFactory serviceFactory = new MercuryServiceFactory(configuration, environment);
+//    ServiceFactory serviceFactory = new MercuryServiceFactory(configuration, environment);
```
drsnyder: The `MercuryServiceFactory` should be the default. Eventually I think we will need both in the `ArticleService`. We'll probably be combining data from the two end points.

pandora-core/src/main/java/com/wikia/pandora/core/domain/builder/ArticleBuilder.java
```diff
@@ -1,12 +1,12 @@
...
+import com.wikia.pandora.core.domain.Article;
 
 public class ArticleBuilder {
 
   private Integer id;
   private String title;
-  private String content;
+  private Integer ns;
 
   private ArticleBuilder() {
```
drsnyder: nitpick: Personally, I prefer builders that are tied into the classes that they build e.g. having `private Article(Builder builder)` with the `public static final class Builder` as part of the same class ([example](https://github.com/drsnyder/dropdora/blob/master/src/main/java/com/wikia/pandora/api/Article.java)). Our POJOs are going to have a lot of elements in their constructors if we go this way.

pandora-core/src/main/java/com/wikia/pandora/core/domain/Article.java
```diff
@@ -1,16 +1,16 @@
...
 
 public class Article {
 
   private final Integer id;
   private final String title;
-  private final String content;
+  private final Integer ns;
 
-  public Article(Integer id, String title, String content) {
+  public Article(Integer id, String title, Integer ns) {
```
drsnyder: I suppose one way to support the separate class approach for the builder would be to have this class accept the builder as the parameter instead of the individual elements.

drsnyder:
Sorry for all the nitpicky comments. My primary concerns I center around the file expansion due to the POJOs and their builders. IMO the builder is there to allow you to create an immutable object while also performing an [extract parameter object](http://www.refactoring.com/catalog/introduceParameterObject.html). The way things are now, you end up editing 3 files to add a field to the response. I think I would rather lower that number and simplify the constructors even it it means a larger POJO (as in [this example](https://jlordiales.wordpress.com/2012/12/13/the-builder-pattern-in-practice/) ) that includes the builder as a static class inside.

mercury-api/src/main/java/com/wikia/pandora/resources/HalMercuryResource.java
sqreek: Yes, this new representation only. 
I Created MercuryArticles for separete them from Articles path

pandora-core/src/main/java/com/wikia/pandora/core/domain/Article.java
sqreek: This is the fastest way to create builder from resources(InteliJ, Builder Construct plugin) we have for this moment.
If U add new field to class, just delete constructor, and Builder class, and generate them again.
It will be nice to have one parameter constructor, but In my opinion we want to add new field with small effort.

pandora-core/src/main/java/com/wikia/pandora/core/domain/builder/ArticleBuilder.java
sqreek: same answer like in previous comment:

This is the fastest way to create builder from resources(InteliJ, Builder Construct plugin) we have for this moment.
If U add new field to class, just delete constructor, and Builder class, and generate them again.
It will be nice to have one parameter constructor, but In my opinion we want to add new field with small effort.

pandora-examples/src/main/java/com/wikia/pandora/examples/PandoraApplication.java
sqreek: I'm planning to create Placeholder for share what resources are in the system between Resources, but it still need some grooming

pandora-core/src/main/java/com/wikia/pandora/core/domain/Article.java
pchojnacki: I think since hopefully most of our POJOs will be hopefully constructed automatically. Then adding builders to them makes it more complicated and 2nd they stop being Plain Old Java Objects if they have builder inside.

pchojnacki:
@drsnyder 
IMO us building builders for everything is little bit counter productive. But if we do it then keeping it in separate files, Is generally the Java way.

An alternative would be using something like this:
https://github.com/mkarneim/pojobuilder

pchojnacki:
This has spent too much time in review already. Since we have to push forward. I'll merge and we can address any leftover problems with separate pull requests.
so ++LGTM

PR #7, opened 2015-01-27
Mobile configuration service merge

nandy-andy:
Together with the Service/API Team we decided to merge Java services repos with our main pandora repo. This is a pull request in which it is happening.

**Why?**
It is a common trend and seems as good practice followed by "huge players" - we do not want to stay behind.

Second reason is to create libraries which will be useful for many other services. Once written in the repo will be easy to find and re-use.

We found out in the early state of services (Mobile Configuration Service and MercuryApi Service) keeping everything in one repository will be really helpful while refactoring code.

**Are you serious?!**
Yes, we are. But we are also aware of one downside of this solution: 1GB+ huge repository which needs to be cloned in example when new developer joins us. However, right now this repository is rather small and we want to give it a try. If we got to the size issue we will re-think this approach.

ping @drsnyder @jcellary @sqreek @nmonterroso @pchojnacki

### DIFF
--- a/mobile-config-service/.gitignore
+++ b/mobile-config-service/.gitignore
@@ -1,0 +1,11 @@
+/.idea
+/.gradle
+/logs
+/build
+*.jar
+*.class
+*.iml
+gradlew
+gradlew.bat
+settings.gradle
+mobile-config.yml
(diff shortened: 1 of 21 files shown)


drsnyder:
lgtm

jcellary:
Ok I'm assuming we can merge it? I will be committing my own changes to mobile config shortly, so I'll merge this first so that we don't get lost in PRs.

drsnyder:
:+1:

PR #9, opened 2015-01-29
Platform 789 create article resource

sqreek:
Added ArticleResource.
Added MediawikiArticle Service
Extend MediawikiGateway for necessary functionality
Extend Medawiki-fluent-api for necessary functionality


drsnyder:
Please provide a [description](https://github.com/Wikia/guidelines/blob/master/Process/PullRequests.md) :).

pandora-core/src/main/java/com/wikia/pandora/core/domain/Revision.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.Date;
+
+public class Revision {
+
+  private int revId;
+  private int parentId;
+  private String user;
+  private Date timestamp;
+  private String comment;
```
drsnyder: `private final String`?

pandora-core/src/main/java/com/wikia/pandora/core/domain/Media.java
```diff
@@ -3,4 +3,19 @@
 
 public class Media {
 
+  private int ns;
+  private String title;
```
drsnyder: `private final *`?

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mediawiki/MediawikiArticleService.java
```diff
@@ -73,4 +78,81 @@ public Comment getComment(String wikia, String title, Long commentId) {
     }
     return articleList;
   }
+
+  @Override
+  public List<Category> getArticleCategories(String wikia, String title) {
+    ApiResponse apiResponse = this.getGateway().getArticleCategories(wikia, title);
+    List<Category> categoryList = new ArrayList<>();
+    if (apiResponse.getQuery().getFirstPage() != null
```
drsnyder: I think we should have `getFirstPage()` return `Optional<String>` so we can avoid Nulls. I'm wondering if we should make that a guideline.

pandora-resources/src/main/java/com/wikia/pandora/resources/HalArticleResource.java
```diff
@@ -153,24 +188,57 @@ public Object getArticleUsers(@PathParam("wikia") String wikia,
...
         representationFactory.newRepresentation(uri.build(wikia, title));
-
+    List<User> users = articleService.getArticleContributors(wikia, title);
+    for (User user : users) {
+      representation.withLink("user",
+                              getLinkToUser(wikia, user.getName()),
+                              null,
+                              user.getName(),
+                              null,
+                              null);
```
drsnyder: The null blocks look odd here. Should this be a wrapper function that takes 3 parameters?

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mediawiki/MediawikiArticleService.java
```diff
@@ -73,4 +78,81 @@ public Comment getComment(String wikia, String title, Long commentId) {
...
+            .withUser(mwRevision.getUser())
+            .build();
+        revisions.add(revision);
+      }
+    }
+    return revisions;
+  }
+
+  @Override
+  public List<com.wikia.pandora.core.domain.User> getArticleContributors(String wikia, String title) {
```
drsnyder: Is the fully qualified path (`com.wikia.pandora.core.domain.User`) necessary?

drsnyder:
Other than minor comments, lgtm.

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mediawiki/MediawikiArticleService.java
sqreek: Yes becuse there is User from mediawiki-fluent-api, and there is a User from pandora-core

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mediawiki/MediawikiArticleService.java
sqreek: We need to talk about it, are we want to is it anywhere like a standard, or not.
If yes, then I will refactor this on other pull request

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mediawiki/MediawikiArticleService.java
pchojnacki: Since its all the same codebase could we rename one of the users? Just a thought since I dislike fully qualified paths

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mediawiki/MediawikiArticleService.java
pchojnacki: Im a little hesitant to return Optional<> everywhere. But it might  be that this concept is just new for me. I think we could revisit Optional<Guideline> in the future

pandora-resources/src/main/java/com/wikia/pandora/resources/HalArticleResource.java
sqreek: Created wrapper for this

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mediawiki/MediawikiArticleService.java
sqreek: When we put Mediawiki-fluent-api it will be two different things

pchojnacki:
LGTM

PR #10, opened 2015-01-29
fix MW test 

sqreek:
Previous test only check is mockito working correctly.

I suggest to mock HttpClientOnly

/cc @drsnyder


mediawiki-fluent-api/src/test/java/com/wikia/mwapi/MWApiTest.java
```diff
@@ -22,27 +19,25 @@ public void testBuildUrl() {
...
-    Mockito.when(mwApi.wikia(Mockito.anyString())).thenCallRealMethod();
-    Mockito.when(mwApi.queryAction()).thenCallRealMethod();
-    Mockito.when(mwApi.titles(Mockito.anyString())).thenCallRealMethod();
-
-    InputStream jsonStream = new ByteArrayInputStream(
-        FixtureHelpers.fixture("fixtures/kermit-the-frog-title.json").getBytes(StandardCharsets.UTF_8)
-    );
-    Mockito.when(mwApi.handleMWRequest(Mockito.anyString())).thenReturn(jsonStream);
-
-    ApiResponse response = mwApi
```
drsnyder: @sqreek do we really need to test the nuts and bolts of `HttpClient`? I think it's reasonable at this point to assume that it will do it's job. Its work is encapsulated [here](https://github.com/Wikia/pandora/blob/Mercury_Api_Replacement/mediawiki-fluent-api/src/main/java/com/wikia/mwapi/MWApi.java#L103).

mediawiki-fluent-api/src/test/java/com/wikia/mwapi/HttpClientMock.java
```diff
@@ -0,0 +1,109 @@
+package com.wikia.mwapi;
```
drsnyder: These mocked stubs shouldn't be necessary. This is why tools like Mockito exist-- so you don't have to create these classes. By using Mockito you can also confirm that the calls were made as expected.

drsnyder:
How is this different from using a mocking library/framework?

mediawiki-fluent-api/src/test/java/com/wikia/mwapi/MWApiTest.java
```diff
@@ -22,27 +19,25 @@ public void testBuildUrl() {
...
-    MWApi mwApi = Mockito.mock(MWApi.class);
-    Mockito.when(mwApi.get()).thenCallRealMethod();
-    Mockito.when(mwApi.wikia(Mockito.anyString())).thenCallRealMethod();
-    Mockito.when(mwApi.queryAction()).thenCallRealMethod();
-    Mockito.when(mwApi.titles(Mockito.anyString())).thenCallRealMethod();
-
-    InputStream jsonStream = new ByteArrayInputStream(
-        FixtureHelpers.fixture("fixtures/kermit-the-frog-title.json").getBytes(StandardCharsets.UTF_8)
-    );
-    Mockito.when(mwApi.handleMWRequest(Mockito.anyString())).thenReturn(jsonStream);
```
drsnyder: This is testing more than that mockito works. One way to read this is "assuming that `HttpClient` does what we expect does `MWApi` behave as expected and give us the title and the id.

mediawiki-fluent-api/src/test/java/com/wikia/mwapi/MWApiTest.java
sqreek: We dont test HttpClient stuff, only MwApi, the correct way from gathering parameters, building url and get answer.
Its Assume HttpClient response correct values if url was corrected

mediawiki-fluent-api/src/test/java/com/wikia/mwapi/HttpClientMock.java
sqreek: Yes,  it can be mocked by Mockito (HttpClient, HttpEntity, HttpResponse)

mediawiki-fluent-api/src/test/java/com/wikia/mwapi/MWApiTest.java
sqreek: This is testing only Syntax what function return, becuse every get will returning excpected Json

sqreek:
Closed becuse of https://github.com/Wikia/pandora/pull/11

PR #11, opened 2015-01-29
Fix the MWApi test by only mocking injected dependencies

drsnyder:
This approach means that only the injected dependency is mocked which is preferred. The only issue with this approach is the inability to compare two `HttpUriRequest` objects with the same constructor input. Mockito provides a work around which allows you to capture the input and use that for verification purposes. There are [other ways to do this](http://docs.mockito.googlecode.com/hg/1.9.5/org/mockito/ArgumentMatcher.html) but I think this is the cleanest and simplest.

/cc @sqreek

### DIFF
--- a/mediawiki-fluent-api/src/test/java/com/wikia/mwapi/MWApiTest.java
+++ b/mediawiki-fluent-api/src/test/java/com/wikia/mwapi/MWApiTest.java
@@ -2,6 +2,12 @@ package com.wikia.mwapi;
 
 import com.wikia.mwapi.domain.ApiResponse;
 
+import org.apache.http.HttpEntity;
+import org.apache.http.HttpResponse;
+import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpUriRequest;
+import org.mockito.ArgumentCaptor;
 import org.mockito.Mockito;
 import io.dropwizard.testing.FixtureHelpers;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


mediawiki-fluent-api/src/test/java/com/wikia/mwapi/MWApiTest.java
```diff
@@ -27,25 +33,37 @@ public void testBuildUrl() {
 
   @Test
   public void testGet() throws IOException {
```
pchojnacki: Could we change it to more descriptive name?

mediawiki-fluent-api/src/test/java/com/wikia/mwapi/MWApiTest.java
```diff
@@ -27,25 +33,37 @@ public void testBuildUrl() {
...
+    String wikia = "muppet";
+    String title = "Kermit the Frog";
+    String url = MWApi.createBuilder()
+        .wikia(wikia)
+        .queryAction()
+        .titles(title)
+        .url();
+    HttpUriRequest request = new HttpGet(url);
+
+    HttpClient httpClient = Mockito.mock(HttpClient.class);
```
pchojnacki: What about extracting this moch creation to helper method?

mediawiki-fluent-api/src/test/java/com/wikia/mwapi/MWApiTest.java
```diff
@@ -27,25 +33,37 @@ public void testBuildUrl() {
...
-        .titles("Kermit the Frog")
+        .titles(title)
         .get();
     assertEquals("Kermit the Frog", response.getQuery().getFirstPage().getTitle());
     assertEquals(50, response.getQuery().getFirstPage().getPageId());
-  }
 
+    ArgumentCaptor<HttpUriRequest> argument = ArgumentCaptor.forClass(HttpUriRequest.class);
+    Mockito.verify(httpClient).execute(argument.capture());
+    assertEquals(url, argument.getValue().getURI().toString());
```
pchojnacki: this assertion seems to only test this in the end:
MWApi.createBuilder(httpClient).wikia(wikia).queryAction().titles("Kermit the Frog").titles(title) == MWApi.createBuilder(httpClient).wikia(wikia).queryAction().titles("Kermit the Frog").titles(title)

drsnyder: I know it's silly. Object equality doesn't work as you would expect with `HttpUriRequest`. If you pass `request` here `Mockito.when(httpClient.execute(request)).thenReturn(httpResponse);` the mock does not behave as expected.

drsnyder:
Thanks for the feedback. It should be covered in 01a15f8.

pchojnacki:
LATM
A - Awesome

PR #13, opened 2015-02-01
Cosher HTTP errors

nandy-andy:
Mobile configuration service was returning so far blank pages with HTTP 400 and [HTTP 404](https://github.com/Wikia/pandora/compare/cosher-http-errors?expand=1#diff-a38472d5b4469546002d00ed0fbe868eL42) statuses in headers. [According to our API guidelines](https://github.com/Wikia/guidelines/tree/master/APIDesign#error-handling) it is supposed to return proper HTTP code statuses, `application/problem+json` type and an JSON object with `title` and `details`.

Changes below implements mentioned requirements.

//cc @drsnyder @jcellary @nmonterroso @sqreek @pchojnacki

### DIFF
--- a/mobile-config-service/src/main/java/com/wikia/mobileconfig/exceptions/ConfigurationNotFoundException.java
+++ b/mobile-config-service/src/main/java/com/wikia/mobileconfig/exceptions/ConfigurationNotFoundException.java
@@ -1,0 +1,19 @@
+package com.wikia.mobileconfig.exceptions;
+
+import com.sun.jersey.api.Responses;
+import com.wikia.mobileconfig.core.Problem;
+
+import javax.ws.rs.core.Response;
+
+public class ConfigurationNotFoundException extends MobileConfigException {
+  static String EXCEPTION_MESSAGE_FORMAT = "Configuration for %s on %s could not be found";
+
+  public ConfigurationNotFoundException(String platform, String appTag) {
+    super(
+        Response.status(Responses.NOT_FOUND)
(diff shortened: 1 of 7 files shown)


pchojnacki:
LGTM

jcellary:
I vote for moving this to pandora core, to have a consistent error handling base for all projects not just mobile config service, as there is nothing service specific about it.

mobile-config-service/src/main/java/com/wikia/mobileconfig/exceptions/ConfigurationNotFoundException.java
```diff
@@ -0,0 +1,19 @@
...
+public class ConfigurationNotFoundException extends MobileConfigException {
+  static String EXCEPTION_MESSAGE_FORMAT = "Configuration for %s on %s could not be found";
+
+  public ConfigurationNotFoundException(String platform, String appTag) {
+    super(
+        Response.status(Responses.NOT_FOUND)
+            .entity(new Problem("Not found", String.format(EXCEPTION_MESSAGE_FORMAT, appTag, platform)))
+            .type(RESPONSE_TYPE)
+            .build()
+    );
```
drsnyder: We may not want to mix the exception with the representation. One approach would be to have a standard set of exceptions then use a `com.sun.jersey.spi.container.ContainerResponseFilter` to toggle the response code and use a `MessageBodyWriter` to handle the formatting.

drsnyder: Created https://wikia-inc.atlassian.net/browse/SERVICES-17 to address this.

PR #14, opened 2015-02-02
Fix some compilation of nirvana-api, and remove nirvana-api for time being until compilation issues are resolved fully

pchojnacki:
1) when undefined MAVEN_EXEC causes exception
Also after fixing 1) the task seems to be stuck in perpetually so I decided to disable this task for the time being


sqreek:
LGTM
(G)-great

sqreek:
lgtm

PR #15, opened 2015-02-02
Rename pandora-examples to pandora-service and cleanup gradle

pchojnacki:
Done some necessary Gradle config cleanup
Rename pandora examples to hopefully something more telling.


sqreek:
lgtm

PR #16, opened 2015-02-02
Code organization

sqreek:
Move Resources from Pandora-resources to pandora-service,
Move Domains from Pandora-core to pandora-service
Move Services from pandora-gateways to pandora-service
Disable Mercury-api project from gradle
Disable pandora-resources project from gradle


pchojnacki:
I wonder why mercury api is disabled. But other than that it LGTM

pchojnacki:
I added back mercury-api dependency.
so now this LGTM :)

PR #19, opened 2015-02-03
Upgrade to Dropwizard 0.8.0-rc2 and introduce integration tests separation

pchojnacki:
Dropwizard 0.7.1 uses old jersey libraries which caused me a lot headache when trying to use some existing consul libraries.
My proposal is to upgrade to version 0.8.0-rc2 since Dropwizard is 'mostly' a glue between many existing libraries and in my opinion something tagged as RC-2 is generally considered stable.

Documentation for 0.8.0-rc2 can be found at
http://dropwizard.io/0.8.0-rc2/docs/manual/testing.html

@jcellary @drsnyder @nandy-andy @nmonterroso @sqreek

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,4 +6,8 @@ subprojects {
     apply plugin: 'java'
     sourceCompatibility = 1.8
     version = "0.1.0"
+
+    task unitTest(type: Test) {
+        exclude '**/integration/**'
+    }
 }
--- a/README.md
+++ b/README.md
@@ -9,12 +9,22 @@ about this project that’s a good place to start.
 
 ## Usage
 
+Launching unit tests:
+```bash
+./gradlew unitTest
+```
+
+Launching all tests:
+```bash
+./gradlew test
+```
+
 To run the server:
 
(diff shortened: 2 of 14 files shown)


build.gradle
```diff
@@ -6,4 +6,8 @@ subprojects {
     apply plugin: 'java'
     sourceCompatibility = 1.8
     version = "0.1.0"
+
+    task unitTest(type: Test) {
+        exclude '**/integration/**'
```
sqreek: This should be in another PR or in description

sqreek:
Its LGTM.
Please add description about you setup integration test in this pull request too.

nmonterroso:
LGTM :ship:

drsnyder:
> Please add description about you setup integration test in this pull request too.

+1. Please update the README with instructions on how to run unit and integration tests. We should also propose a convention. For example, where should the `integration` be added to the package? Near the tail?

It looks like the `test` task now tests everything while `unitTest` excludes anything with an integration directory in the path.

PR #20, opened 2015-02-03
Refactor Mediawiki-fluent-api using decorator pattern

sqreek:
Refactor Mediawiki-fluent-api to be more fluent using decorator pattern
It's easier now for adding new functionality 
Code is more readable.
Not all of mediawiki api is covered, but now it is very easy to extend it, (look at MWApiRevisions)

Alternative for fluent aproach is to keep everything in one class, and filter next step only by interfaces

Documentation of Mediawiki api https://en.wikipedia.org/w/api.php?action=help&modules=query


mediawiki-fluent-api/src/main/java/com/wikia/mwapi/MWApi.java
```diff
@@ -17,204 +13,50 @@
...
-  private String query;
-  private String[] titles;
-  private String format;
-  private final List<String> prop;
-
-  private HttpClient httpClient;
-  private Integer rvLimit;
-  private String rvProp;
-  private String list;
+  private static final String DEFAULT_BASEURL = "http://%s.wikia.com/api.php";
```
pchojnacki: This should be more like http://%s/api.php
so it will work on:
jedipedia.de etc...
and even wikipedia itself

mediawiki-fluent-api/src/main/java/com/wikia/mwapi/MWApi.java
```diff
@@ -17,204 +13,50 @@
...
+    builder.addParameter("action", getAction().toString());
+    builder.addParameter("format", getFormat().toString());
   }
 
   @Override
-  public OptionChoose categories() {
-    prop.add(CATEGORIES);
-    return this;
+  protected URIBuilder createURIBuilder() throws URISyntaxException {
+    return new URIBuilder(String.format(DEFAULT_BASEURL, getWikia()));
```
pchojnacki: as mentioned before getWikia is not so simple.
For now we should think more/only in terms of hostnames. Other ideas of addressing specific wikis should be handled in the future.

mediawiki-fluent-api/src/main/java/com/wikia/mwapi/MWApi.java
```diff
@@ -17,204 +13,50 @@
...
-  private String[] titles;
-  private String format;
-  private final List<String> prop;
-
-  private HttpClient httpClient;
-  private Integer rvLimit;
-  private String rvProp;
-  private String list;
+  private static final String DEFAULT_BASEURL = "http://%s.wikia.com/api.php";
+  private final HttpClient httpClient;
```
pchojnacki: Since we already use Jersey, then I think going with Jersey-cllient might be more preferable. Since it natively supports async operation.
See here:
https://jersey.java.net/documentation/latest/client.html

for jersey 1.8
https://jersey.java.net/apidocs/1.8/jersey/com/sun/jersey/api/client/package-summary.html

pchojnacki:
It seems too complicated when doing simple things. 
The amount of classes this Pull request introrducees is staggering.
And half of them are just one method long.

This is not a good pattern.

pchojnacki:
I can take a second look after the classes from decorator/query are gone.
But for me this PR is No-Go nu uh

pchojnacki:
The same for all those enums.

pchojnacki:
Although enums would be more acceptable if they were Inner.
I get that it is meant to be generic client. But honestly, the MW api is so extensive that I doubt that even 10x more code will handle all the cases. And we really right now need to read  from 3 or 5 urls, with few parameters that would be handled in 1/100th of the code. 
Also In future or even right now, it might be better to expose smaller single responsibility endpoints and connect to them. Than to build complete solutions based on not so tested MW api.

pchojnacki:
Anyway apart of my dislike for over complication I think that this code would work. Generally :)
Although we really have to answer the question how much will we use MW API directly and how much we will use through custom Nirvana endpoints. @drsnyder ?

IMO this area is heavily over invested and we still have a lot work elsewhere, and I reaaaly would not like for this pull request to be blocker for any of that.

drsnyder:
I agree that it's overly complicated. Our usage of the MW API will be light for the foreseeable future. We'll be doing a lot more with the Nirvana API in phase 1.

Let's keep the basics of the MW API in place. I think it would be better to hold off on adding more functionality until we really need it.

owend:
I haven't had time yet but I'd like to compare to this approach: https://github.com/ivanceras/fluentsql  

We adapted that java library to PHP for our fluent sql api, and it uses a few nice "java-isms" such as static classes that make the code relatively clean and readable.

sqreek:
this fluentsql is actualy just expose all properties. It is not fluent at all. For example it allow to create SQL.WHERE(something).SELECT(something).
Here is some info about more correct way: http://blog.jooq.org/2012/01/05/the-java-fluent-api-designer-crash-course/

pchojnacki:
OK IMO YMMV but  LGTM :+1:

PR #21, opened 2015-02-03
[PLATFORM-776] Mobile config service translations, app list and images

jcellary:
Added support for translations and endpoints for app list and simple images as requested by the mobile team.

Translations are performed when the config contains anything in the format _SOME_TEXT_ (underscore at the beginning and end of key, let me know if you have suggestions for a better naming convention which is compatible with json)

Translations are simply loaded from local resource json file. In the future we would possibly want to reload them dynamically when they change, but currently as discussed with mobile team there are no use cases for that, so the resource is only read on startup.

Responses from the mobile apps django service are cached for 1 hour. Currently the app list is a simple forward of the data received from apps service, they will be changing it and we'll need to include proper filtering, but this will be done later on.

Images are read from ceph and simply returned as a binary data (thumbnailer is not involved). Again current use cases are very limited, so I kept it simple, I think we should enable caching on fastly level to improve performance on icon requests, so I didn't built any ceph result caching on the service level.

Both unit tests and integration tests provided (for now integration tests still run on each built, will change that in future PRs).

pinging @pchojnacki @sqreek @drsnyder @nmonterroso @nandy-andy @aniaMu


mobile-config-service/src/main/java/com/wikia/mobileconfig/core/AppList.java
```diff
@@ -0,0 +1,19 @@
+package com.wikia.mobileconfig.core;
+
+import java.util.List;
+
+public class AppList {
+
+  private List apps;
```
drsnyder: Minor: We have been encouraging immutable POJOs.

mobile-config-service/src/main/java/com/wikia/mobileconfig/resources/AppListResource.java
```diff
@@ -0,0 +1,51 @@
...
+import com.wikia.mobileconfig.core.AppList;
+import com.wikia.mobileconfig.exceptions.MobileConfigException;
+import com.wikia.mobileconfig.service.AppsListService;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+
+@Path("/appList/platform/{platform}/")
```
drsnyder: Having a verb in the URL looks odd. A `GET` on this resource will end up listing them anyway. I would suggest removing the `List` part. 

Also, does the `/app` have any special meaning here?

mobile-config-service/src/main/java/com/wikia/mobileconfig/resources/ImageResource.java
```diff
@@ -0,0 +1,44 @@
...
+        @PathParam("filename") String filename
+    ) throws java.io.IOException, URISyntaxException {
+
+        byte[] imageBytes = this.imageService.getImage(filename);
+
+        if (imageBytes == null || imageBytes.length == 0) {
+            throw new ImageNotFoundException(filename);
+        }
+
+        String contentType = "image/" + Files.getFileExtension(filename);
```
drsnyder: Minor: This is probably fine for now. FYI, we use [this library](http://tika.apache.org/) in vignette to detect the content type.

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/AppsDeployerList.java
```diff
@@ -55,10 +69,31 @@ public AppsDeployerList(Environment environment, MobileConfigConfiguration confi
...
+
+    long diffInSeconds = TimeUnit.MILLISECONDS.toSeconds(new Date().getTime() - appsListUpdateTime.getTime());
+    if (appsList == null || diffInSeconds > CACHE_RESULT_SEC) {
+      appsList = requestAppsList();
+      appsListUpdateTime = new Date();
+    }
+
+    return appsList;
+  }
+
```
drsnyder: It's probably ok for now, but this logic (to fetch the listings over HTTP) should probably be in a gateway namespace.

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/ConfigurationServiceBase.java
```diff
@@ -1,13 +1,55 @@
...
+          }
+        } else {
+          translateNode(langCode, entry.getValue());
+        }
+      }
+    }
+  }
+
+  protected void translateConfiguration(MobileConfiguration config, String langCode) {
+
```
drsnyder: Extra spacing?

mobile-config-service/src/main/java/com/wikia/mobileconfig/resources/AppListResource.java
aniaMu: Or renaming to apps/platform/{platform}

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/AppsDeployerList.java
```diff
@@ -68,6 +103,12 @@ public Boolean isValidAppTag(String appTag) throws IOException {
   }
 
   @Override
+  public List<HashMap<String, Object>> getAppList(String platform) throws IOException {
+    List<HashMap<String, Object>> list = this.getAppsList();
+    return list;
```
aniaMu: just ```return this.getAppsList();```

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/AppsListService.java
```diff
@@ -9,5 +11,7 @@
 
   public Boolean isValidAppTag(String appTag) throws IOException;
 
+  public List<HashMap<String, Object>> getAppList(String platform) throws IOException;
```
aniaMu: methods defined in interfaces are pubic by default so you can skip the modifier

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/ImageService.java
```diff
@@ -0,0 +1,74 @@
...
+
+    try {
+      image = this.executeHttpRequest(createCephUrl(filename));
+    } catch (IOException e) {
+      MobileConfigApplication.logger.error(
+          String.format("Exception caught when requesting Image %s:%s", filename, e.toString())
+      );
+    }
+
+    return image;
```
drsnyder: I think it would be nice to avoid `null` here by using [`Optional`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Optional.html). Then the caller can check [`isPresent`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Optional.html#isPresent()) before attempting to read the value.

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/ConfigurationServiceBase.java
```diff
@@ -1,13 +1,55 @@
...
+    result.put(key, translation);
+    return result;
+  }
+
+  protected boolean isTranslationKey(String key) {
+    return key.length() >=3 && key.charAt(0) == '_' && key.charAt(key.length() - 1) == '_';
+  }
+
+  protected void translateNode(String langCode, Object node) {
+    if (node instanceof Map) {
```
aniaMu: we should try to avoid using ```instanceof```

mobile-config-service/src/main/java/com/wikia/mobileconfig/utils/Translator.java
```diff
@@ -0,0 +1,84 @@
...
+        if (instance == null) {
+          instance = new Translator();
+        }
+      }
+    }
+    return instance;
+  }
+
+  private void loadTranslations(String langCode) {
+
```
drsnyder: Extra white space? Same below in `getTranslationMap`.

mobile-config-service/src/test/java/com/wikia/mobileconfig/integration/ConfigIntegrationTest.java
```diff
@@ -0,0 +1,58 @@
...
+import io.dropwizard.testing.junit.DropwizardAppRule;
+
+import static org.fest.assertions.api.Assertions.assertThat;
+
+public class ConfigIntegrationTest {
+  @ClassRule
+  public static final DropwizardAppRule<MobileConfigConfiguration>
+      RULE = new DropwizardAppRule<MobileConfigConfiguration>(
+        MobileConfigApplication.class,
+        "mobile-config.yml"
```
drsnyder: Does this require the `"mobile-config.yaml"` to be present in the working directory?

mobile-config-service/src/main/java/com/wikia/mobileconfig/resources/AppListResource.java
nandy-andy: Or even applications/platform/{platform} :)

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/AppsListService.java
nandy-andy: I'd leave it just to be explicit :p

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/AppsListService.java
```diff
@@ -9,5 +11,7 @@
...
+
   public Boolean isUp() throws IOException;
```
aniaMu: will this method return a null value? if not then you can change to ```boolean```

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/AppsListService.java
```diff
@@ -9,5 +11,7 @@
 
   public Boolean isValidAppTag(String appTag) throws IOException;
```
aniaMu: will this method return a null value? if not then you can change to ```boolean```

mobile-config-service/src/main/java/com/wikia/mobileconfig/resources/AppListResource.java
jcellary: wow, I think I'll create a pool for the name :) Ok I'll go with applications

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/AppsDeployerList.java
jcellary: You're right I can move it, although in general we want to remove the dependency of the config servcie on the app panel, although we don't have a fixed plan yet how to do it.

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/AppsDeployerList.java
jcellary: Actually this way is more convinient for debugging (you can easly go line by line and have the result temp. variable in the watch list), this is why I sometimes like to declare result variables, but I agree the code looks better if it's one line.

mobile-config-service/src/main/java/com/wikia/mobileconfig/core/AppList.java
jcellary: Ok, so what do you mean here - using Collections.unmodifiableList ? (Because the object already has no setters - the list can only be set through constructor)

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/ConfigurationServiceBase.java
jcellary: I'm not sure how to avoid it in this situation - I guess you mean not to use Object but the actual type? I agree that in most cases it should be done this way, but in this situation we are analyzing nodes of a tree and have to dynamically check if an element is a leaf or node, so I don't think we can avoid type checking.

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/ConfigurationServiceBase.java
jcellary: I assume you mean deleting line 50

mobile-config-service/src/test/java/com/wikia/mobileconfig/integration/ConfigIntegrationTest.java
jcellary: I guess it does, we talked about it on our last update, we need to figure out a better way of handling this globally

jcellary:
I merged changes from master and adapted code to use new lib versions. Also changed things from the PR comments. Let me know if there are any other issues, otherwise I will merge it tomorrow.

mobile-config-service/src/main/java/com/wikia/mobileconfig/core/AppList.java
drsnyder: All I mean is `private final List apps`. It's a minor concern though.

mobile-config-service/src/main/java/com/wikia/mobileconfig/resources/AppListResource.java
drsnyder: Haha. Sorry my mistake, it should be [`apps`](https://github.com/Wikia/guidelines/tree/master/APIDesign#resources-and-naming).

PR #22, opened 2015-02-03
create simple example

sqreek:
Add pandora-example project with simple Resource


pandora-example/example.yaml
```diff
@@ -0,0 +1,12 @@
+logging:
```
pchojnacki: this file shouldn't be committed I believe

pandora-example/src/main/java/com/wikia/exampleservice/configuration/ExampleConfiguration.java
```diff
@@ -0,0 +1,7 @@
+package com.wikia.exampleservice.configuration;
+
+import io.dropwizard.Configuration;
+
+public class ExampleConfiguration extends Configuration {
+
```
pchojnacki: Since this is an example, I think adding 1 dummy option  would greatly help people getting started.
That is add something here in example.sample.yml and some equally dummy use case.

pandora-example/src/main/java/com/wikia/exampleservice/domain/builder/SimplePojoBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+    this.someString = someString;
+    return this;
+  }
+
+  public SimplePojoBuilder withSomeBool(boolean someBool) {
+    this.someBool = someBool;
+    return this;
+  }
+
+  public SimplePojoBuilder but() {
```
pchojnacki: this method seems strangely named, its also unused. Is it needed ?

pandora-example/src/test/java/com/wikia/exampleservice/resources/ExampleResourceTest.java
```diff
@@ -0,0 +1,30 @@
...
+  @Test
+  public void testGetSimplePojo() throws Exception {
+    RepresentationFactory representationFactory = new StandardRepresentationFactory();
+    ExampleResource resource = new ExampleResource(representationFactory);
+    Representation representation = resource.getSimplePojo(7, "John", true);
+    assertEquals(representation.getProperties().get("id"), 7);
+    assertEquals(representation.getProperties().get("someString"), "John");
+    assertEquals(representation.getProperties().get("someBool"), true);
+  }
+}
```
pchojnacki: Ultra small 'issue'/nitpick. No new line at the end of file :D

pchojnacki:
I think small commands overview in README.md might be in order like:
```bash
../gradlew test # run all tests
../gradlew run # launch application
```

Good stuff! After the comments are addressed it LGTM

pandora-example/src/main/java/com/wikia/exampleservice/domain/builder/SimplePojoBuilder.java
sqreek: https://github.com/mjedynak/Builder-Generator/issues/10

pchojnacki: how is it better than simple:
http://en.wikipedia.org/wiki/Clone_%28Java_method%29 ?

pchojnacki: ok just found
http://stackoverflow.com/questions/2427883/clone-vs-copy-constructor-which-is-recommended-in-java
that clone in Java might be broken

in that case no buts for but.

pchojnacki: But I stil don't like the idea of adding code that no one has use case for.
Generated or not.

sqreek: Tradeoff create builder by hand without plugin vs create builder by plugin with but
alternative: create fork/request from plugin repository and delete but/ask for create setup for it

pandora-example/src/main/java/com/wikia/exampleservice/configuration/ExampleConfiguration.java
sqreek: added example

pandora-example/src/test/java/com/wikia/exampleservice/resources/ExampleResourceTest.java
sqreek: added EOF

pchojnacki:
LGTM

PR #23, opened 2015-02-04
Decouple host from MercuryGateway templates

drsnyder:
Decouple host from MercuryGateway templates-- this will allow the service to connect to a proxy or intermediary which will provide 2 benefits:
 * easier integration testing (e.g. using http://wiremock.org/getting-started.html)
 * ability to introduce a caching intermediary e.g. varnish

With this request I also
 * set the default gateway to the `MercuryGateway`. This should be our starting point (or Nirvana).
 * simplified the `MercuryGateway` tests.

Sorry for the lack of atomicity. I went ahead with the above two points so that integration testing can begin.

/cc @Wikia/platform-team


pandora-core/src/main/java/com/wikia/pandora/core/impl/configuration/PandoraConfiguration.java
```diff
@@ -2,6 +2,8 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import org.hibernate.validator.constraints.NotEmpty;
```
pchojnacki: I'm not sure this is correct annotation.
Just a hunch based on package name. It would be good to have test case testing this behavior.

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mercury/MercuryGateway.java
```diff
@@ -29,30 +28,26 @@
...
+  private final String host;
+  private final Integer port;
 
-  public MercuryGateway(HttpClient httpClient, String articleRequestFormat,
-                        String commentsRequestFormat) {
+  public MercuryGateway(HttpClient httpClient, String host, Integer port) {
     this.httpClient = httpClient;
-    this.articleRequestFormat = articleRequestFormat;
-    this.commentsRequestFormat = commentsRequestFormat;
+    this.host = host;
```
pchojnacki: I think using java.net.URI and maybe some kind of URI builder might be more Java way than filling up template.

pandora-core/src/main/java/com/wikia/pandora/core/impl/configuration/PandoraConfiguration.java
pchojnacki: I just looked at the code below, and since this is Dropwizard configuration a unit test might be' too much'.

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mercury/MercuryGateway.java
```diff
@@ -97,47 +90,30 @@ public MercuryGateway(HttpClient httpClient) {
...
-          return EntityUtils.toString(entity.get());
-        } else {
-          throw new ClientProtocolException(
-              String.format("Unexpected response status %d from '%s'.", status, requestUrl));
-        }
-      }
-
-    };
-
+    httpGet.setHeader(HTTP.TARGET_HOST, String.format(DEFAULT_WIKIA_HOST_FORMAT, wikia));
```
pchojnacki: I do wonder if this is correct way to handle this case.
Have you tried just configuring the varnish as a proxy? like e.g.: https://hc.apache.org/httpcomponents-client-ga/httpclient/examples/org/apache/http/examples/client/ClientExecuteProxy.java

And then just normally talking to muppet.wikia.com

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mercury/MercuryGateway.java
```diff
@@ -97,47 +90,30 @@ public MercuryGateway(HttpClient httpClient) {
     return ImmutableMap.copyOf(articleComments);
   }
 
-  public String formatArticleMercuryRequest(String wikia, String title) {
-    return formatMercuryRequest(wikia, title, this.articleRequestFormat);
+  public String formatArticleMercuryRequest(String title) {
```
pchojnacki: Nitpick: I think better naming for these methods would be like getMercuryApiArticleUrl
because when I'm seeing format...Request I expect the method to actualy return formatted Request object

pandora-service/pandora.yaml.sample
```diff
@@ -1,3 +1,5 @@
+mercuryGatewayHost: varnish.service-cache.host
```
pchojnacki: I think actually all gateways will be using this configuration. 
So it might be better and less confusing to have some more general name liike:
internalWikiaProxy etc

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mercury/MercuryGateway.java
```diff
@@ -97,47 +90,30 @@ public MercuryGateway(HttpClient httpClient) {
...
+    return formatMercuryRequest(title, this.DEFAULT_MERCURY_COMMENT_REQUEST_FORMAT);
   }
 
-  public String formatMercuryRequest(String wikia, String title, String format) {
+  public String formatMercuryRequest(String title, String format) {
     try {
-      return String.format(format, wikia, URLEncoder.encode(title, "UTF-8"));
+      return String.format(format, this.host, this.port,
+                           URLEncoder.encode(title, StandardCharsets.UTF_8.toString()));
     } catch (UnsupportedEncodingException e) {
```
pchojnacki: This looks scary, shouldn't it work all the time without causing an exception? If so i think this needs to at least be logged

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mercury/MercuryGateway.java
drsnyder: I think you are right. This "works" but it's not the right way to connect via a proxy.

pandora-gateways/src/main/java/com/wikia/pandora/gateway/mercury/MercuryGateway.java
drsnyder: I have been puzzled by why this throws an exception. How could UTF-8 not be supported? I suppose since it takes arbitrary input then it's possible that the charset supplied doesn't exist.

pandora-core/src/main/java/com/wikia/pandora/core/impl/configuration/PandoraConfiguration.java
drsnyder: Oops. I think Intellij did that for me. I'm pretty sure this should be a Jackson decorator.

pchojnacki:
LGTM :+1:

PR #24, opened 2015-02-04
Register application in Consul on startup

pchojnacki:
As a demo Mobile Config application registers automatically in consul on startup, with TTL heartbeat/deadman's switch.
+ some of the strings used for deployment have been switched to use global gradle variables example

### DIFF
--- a/mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java
+++ b/mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java
@@ -12,6 +12,8 @@ import com.wikia.mobileconfig.gateway.AppsDeployerList;
 import com.wikia.mobileconfig.service.FileConfigurationService;
 import com.wikia.mobileconfig.service.HttpConfigurationService;
 import com.wikia.mobileconfig.service.ImageService;
+import com.wikia.pandora.core.consul.ConsulConfig;
+import com.wikia.pandora.core.consul.ConsulBundle;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -38,6 +40,12 @@ public class MobileConfigApplication extends Application<MobileConfigConfigurati
 
   @Override
   public void initialize(Bootstrap<MobileConfigConfiguration> bootstrap) {
+    bootstrap.addBundle(new ConsulBundle<MobileConfigConfiguration>() {
+      @Override
+      protected ConsulConfig narrowConfig(MobileConfigConfiguration config) {
+        return config.getConsulConfig();
+      }
+    });
   }
 
   @Override
--- a/mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigConfiguration.java
+++ b/mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigConfiguration.java
@@ -4,6 +4,7 @@ import io.dropwizard.Configuration;
 import io.dropwizard.client.HttpClientConfiguration;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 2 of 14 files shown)


mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java
```diff
@@ -37,6 +37,12 @@ public String getName() {
 
   @Override
   public void initialize(Bootstrap<MobileConfigConfiguration> bootstrap) {
+    bootstrap.addBundle(new ConsulBundle<MobileConfigConfiguration>() {
+      @Override
+      protected ConsulConfig narrowConfig(MobileConfigConfiguration config) {
+        return config.getConsulConfig();
```
sqreek: Consider add ConsulConfig to 'default' configuration, and put it to Core.

mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigConfiguration.java
```diff
@@ -28,6 +29,11 @@
   @NotNull
   private String cephPort;
 
+  @Valid
+  @NotNull
+  @JsonProperty
+  private ConsulConfig consulConfig = new ConsulConfig();
```
sqreek: Consider add ConsulConfig to 'default' configuration, and put it to Core.

sqreek: Also move initialization to constructor

pandora-core/src/main/java/com/wikia/pandora/core/consul/ConsulBundle.java
```diff
@@ -0,0 +1,91 @@
...
+                  heartbeat.run();
+                  logger.debug("Heartbeat sent");
+                }
+              } catch (RuntimeException ex) {
+                logger.warn("Heartbeat failed", ex);
+              }
+            },
+            0,
+            config.getHeartbeatInterval().getQuantity(),
+            config.getHeartbeatInterval().getUnit());
```
sqreek: parenthesis are weird.
InteliJ can have trouble with correct adjust them in lambda expressions

pandora-core/src/main/java/com/wikia/pandora/core/consul/ConsulConfig.java
```diff
@@ -0,0 +1,44 @@
...
+  @Valid
+  private Duration heartbeatTTL;
+
+  @NotNull
+  @Valid
+  private List<String> tags;
+
+  @NotNull
+  @Valid
+  private URI consulUri = URI.create("http://localhost:8500");
```
sqreek: should initialization be in constructor?
in this case I think is ok -ish

sqreek:
except questions lgtm

pandora-core/src/main/java/com/wikia/pandora/core/consul/ConsulBundle.java
pchojnacki: yah I'll make it a separate method
I'll try to fix parenthesis but no promises

pchojnacki: I just tried to extract a method out of it and imo closure here makes the most sense since I'd have to either create a method that creates this closure or create the clousure within the same function

mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigConfiguration.java
pchojnacki: I don't want to add this to default Config yet.
Actually WRT to Initialization, I haven't found any guidelines that said that you should always initialize in constructor.

PR #25, opened 2015-02-04
Add Comments endpoint

sqreek:
add HalCommentResource
add Mock in services


pandora-service/src/main/java/com/wikia/pandora/service/mediawiki/MediawikiCommentService.java
```diff
@@ -0,0 +1,51 @@
+package com.wikia.pandora.service.mediawiki;
```
pchojnacki: if we want to keep both mocks and normal implementation then separate package for mocks might be in order, but I believe that As soon as possible this service should be filled with proper implementation

pandora-service/src/main/java/com/wikia/pandora/resources/HalCommentResource.java
```diff
@@ -0,0 +1,103 @@
...
+    this.representationFactory = representationFactory;
+  }
+
+  @GET
+  @Timed
+  @Path("/{commentId}")
+  public Object getComment(@PathParam("wikia") String wikia,
+                           @PathParam("commentId") Long commentId) {
+
+    javax.ws.rs.core.UriBuilder uriBuilder = UriBuilder.getSelfUriBuilder(wikia, commentId);
```
pchojnacki: please rename pandora UriBuilder to something else so that referencing full package is not needed.
Its something of a common pattern that if you create a subclass you name your subclass differently.

pandora-service/src/main/java/com/wikia/pandora/resources/HalCommentResource.java
```diff
@@ -0,0 +1,103 @@
...
+    Representation
+        representation =
+        this.representationFactory
+            .newRepresentation(uriBuilder.build(wikia, commentId));
```
pchojnacki: it seems I dont understand something here. but
Why do you need to specify wikia and commentId twice ? once when constructing builder and now ?
Could it be refactored so that only once is enough ?

pandora-service/src/main/java/com/wikia/pandora/api/service/CommentService.java
```diff
@@ -3,10 +3,12 @@
 
 import com.wikia.pandora.domain.Comment;
 
+import java.util.List;
 import java.util.Optional;
 
 public interface CommentService {
 
-  public Optional<Comment> getCommentsForArticle(String wikia, String title);
+  Comment getComment(String wikia, Long commentId);
```
pchojnacki: I think returning optional is good pattern we shouldn't skip it.

pandora-service/src/main/java/com/wikia/pandora/api/service/CommentService.java
```diff
@@ -3,10 +3,12 @@
 
 import com.wikia.pandora.domain.Comment;
 
+import java.util.List;
 import java.util.Optional;
 
 public interface CommentService {
```
pchojnacki: I'm really interested why we have this as a interface.
There should be one and only one implementation of it?

pandora-service/src/main/java/com/wikia/pandora/resources/HalCommentResource.java
sqreek: Its URIBuilder magic. 
It was becuse helper, get magic stuff(reflection) to put out of what method is calling, return javax UriBuilder, and then javax UriBuilder need parameters to create valid uri.

Actually I found some issue with this aproach, and will remove UriBuilder becuse it's too magic (reflection),  and have issue with simple data (long, int insted of Long, Integer)

pandora-service/src/main/java/com/wikia/pandora/resources/HalCommentResource.java
sqreek: I will remove this helper becuse of some issue. Too much magic(reflection)

pandora-service/src/main/java/com/wikia/pandora/api/service/CommentService.java
sqreek: ok. But in other PR, I will change all at once (in all services)

pandora-service/src/main/java/com/wikia/pandora/api/service/CommentService.java
sqreek: it's valid question and we can talk about it with @drsnyder and @nmonterroso.

Story: I want use pandora stuff but with different aproach in CommentService only.

pchojnacki: hm

pandora-service/src/main/java/com/wikia/pandora/resources/HalCommentResource.java
drsnyder: @sqreek and I talked about this some yesterday. We need a utility that is globally accessible for resolving the link to a given resource. I haven't found any good examples of how to do this with Jersey/dropwizard.

sqreek:
ping @pchojnacki @drsnyder @nmonterroso

pandora-service/src/main/java/com/wikia/pandora/examples/PandoraApplication.java
```diff
@@ -41,10 +43,12 @@ public void run(PandoraConfiguration configuration, Environment environment) {
 
     ArticleService articleService = serviceFactory.createArticleService();
     CategoryService categoryService = serviceFactory.createCategoryService();
+    CommentService commentService = serviceFactory.createCommentService();
```
pchojnacki: please use service instance. Factory is unnecessary in this case.

pandora-service/src/main/java/com/wikia/pandora/service/mercury/MercuryCommentService.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.wikia.pandora.api.service.CommentService;
+import com.wikia.pandora.domain.Comment;
+import com.wikia.pandora.domain.builder.CommentBuilder;
+import com.wikia.pandora.gateway.mercury.MercuryGateway;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class MercuryCommentService extends MercuryService implements CommentService {
```
pchojnacki: This should not implement any interface. There is only 1 implementation that is needed right now so using interfaces and factory is just code smell.

pchojnacki: In other words please remove both the interface and and all usage of it.

pchojnacki:
There is still some unecessary code, once that is addressed then it looks good for merging

sqreek:
I will remove fabric in separate pull request, becuse now, it will provide conflict with 
https://github.com/Wikia/pandora/pull/33
https://github.com/Wikia/pandora/pull/32

PR #29, opened 2015-02-05
Fix not implemented exception version after libraries upgrade

sqreek:


pchojnacki:
We have to add CI build so these kind of problems are detected in PR
But thats different story.
LGTM

PR #35, opened 2015-02-11
Static checks

pchojnacki:
Static checks look for common mistakes in the code. Running them on every pull request seem like a good idea.


gradle/static_checks.gradle
```diff
@@ -11,10 +15,10 @@ dependencies {
 }
 
 tasks.withType(FindBugs) {
-    reportLevel = "low"
+//    reportLevel = "low"
```
nandy-andy: Do we need those commented out lines? :japanese_ogre:

pchojnacki: hmm I want to enable them in the future once we have dealt with current crop of errors. So I left the comments as a reminders

pchojnacki: If thats not ok I can remove them

nandy-andy: I think it's OK after given information :smile_cat:  Thanks for clarification :+1:

gradle/static_checks.gradle
```diff
@@ -1,9 +1,13 @@
 apply plugin: 'findbugs'
+apply plugin: 'pmd'
+
 findbugs {
     ignoreFailures = true
 }
 
-apply plugin: 'pmd'
+pmd {
+    ignoreFailures = true
```
sqreek: Please add comments, about why we ignore this failures

gradle/static_checks.gradle
```diff
@@ -1,9 +1,13 @@
 apply plugin: 'findbugs'
+apply plugin: 'pmd'
+
 findbugs {
     ignoreFailures = true
```
sqreek: Please add comments, about why we ignore this failures

pchojnacki: There is 35 errors from findbugs alone on medium settings. Once we fix that and it stops breaking then I can think of removing this

gradle/static_checks.gradle
pchojnacki: see comment below

sqreek:
lgtm

nandy-andy:
LGMT :+1:

PR #34, opened 2015-02-10
Update example with deployment logging and consul support

pchojnacki:

PR #37, opened 2015-02-11
Mobile Service deployment 

pchojnacki:
first this: https://github.com/Wikia/pandora/pull/34 needs to be merged because of commits overlap

### DIFF
--- a/mobile-config-service/mobile-config.yml.sample
+++ b/mobile-config-service/mobile-config.yml.sample
@@ -38,7 +38,6 @@ server:
 consulConfig:
   tags:
     - auto_test
-    #- production #uncomment on production
   consulUri: http://localhost:8500
   heartbeatInterval: 2s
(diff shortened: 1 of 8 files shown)


mobile-config-service/mobile-config.yml.sample
```diff
@@ -32,7 +37,8 @@ server:
     port: 8081
 consulConfig:
   tags:
-    - auto_pandora
+    - auto_test
+    #- production #uncomment on production
```
sqreek: Is this shouldn't be more verbose? 
Maybe add infomartion about production configuration changings in README.md.

pchojnacki: this is all in flux right now
so adding that to Readme does not make much sense

nandy-andy:
LGTM ;)

PR #40, opened 2015-02-13
Testing sonar-pandora Jenkis job

nandy-andy:
Testing pandora-test if it uses sonar-pandora as well for new pull requests...


nandy-andy:
It works :princess:

PR #42, opened 2015-02-13
fix sonar lower case enums

sqreek:


nandy-andy:
yay! technical debt from 14d to 4d :+1:

PR #44, opened 2015-02-18
SERVICES-12 Separated unit tests and integration tests

nandy-andy:
In order to run full automated test run we need new gradle task for running integration tests. So far, we had a simple unitTest task:
```gradle
task unitTest(type: Test) {
  exclude '**/integration/**'
}
```
and it was good enough. But if we want to group tests in more organized groups it will come difficult to follow the pattern:
```gradle
task unitTest(type: Test) {
  exclude '**/integration/**'
  exclude '**/db/**'
  exclude '**/performance/**'
}

task integrationTest(type: Test) {
  exclude '**/resources/**'
  exclude '**/db/**'
  exclude '**/performance/**'
}

task dbTest(type: Test) {
  exclude '**/resources/**'
  exclude '**/integration/**'
  exclude '**/performance/**'
}

task performanceTest(type: Test) {
  exclude '**/resources/**'
  exclude '**/integration/**'
  exclude '**/db/**'
}
```
I found out a really cool JUnit feature which is `@Category` annotations and below you'll find a different approach - marking your tests classes or separate tests with this annotation:
```gradle
task integrationTest(type: Test) {
  useJUnit {
    includeCategories 'com.wikia.pandora.test.IntegrationTest'
    excludeCategories 'com.wikia.pandora.test.FastTest'
  }
}
```
It seems a way cleaner solution which gives us flexibility to extend our tests suits in future.

The only issue is right now connected to Consul instance on our local machine - if you run integration tests on your local machine it will fail because of no Consul instance.

//cc @drsnyder @gbenson-wikia @jcellary @sqreek @nmonterroso @pchojnacki

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,13 +9,13 @@ subprojects {
 
     task unitTest(type: Test) {
         useJUnit {
-            excludeCategories 'com.wikia.pandora.test.IntegrationTest'
+            excludeCategories 'com.wikia.pandora.core.test.IntegrationTest'
         }
     }
 
     task integrationTest(type: Test) {
         useJUnit {
-            includeCategories 'com.wikia.pandora.test.IntegrationTest'
+            includeCategories 'com.wikia.pandora.core.test.IntegrationTest'
         }
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -7,15 +7,9 @@ subprojects {
     sourceCompatibility = 1.8
     version = "0.1.0"
 
-    test {
-        useJUnit {
-            includeCategories 'com.wikia.pandora.test.FastTest'
-        }
-    }
-
     task unitTest(type: Test) {
         useJUnit {
-            includeCategories 'com.wikia.pandora.test.FastTest'
+            excludeCategories 'com.wikia.pandora.test.IntegrationTest'
         }
     }
 
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,22 @@ subprojects {
     sourceCompatibility = 1.8
     version = "0.1.0"
 
+    test {
+        useJUnit {
+            includeCategories 'com.wikia.pandora.test.FastTest'
+        }
+    }
+
     task unitTest(type: Test) {
-        exclude '**/integration/**'
+        useJUnit {
+            includeCategories 'com.wikia.pandora.test.FastTest'
+        }
+    }
+
+    task integrationTest(type: Test) {
+        useJUnit {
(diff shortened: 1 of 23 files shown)


build.gradle
```diff
@@ -7,7 +7,22 @@ subprojects {
...
-        exclude '**/integration/**'
+        useJUnit {
+            includeCategories 'com.wikia.pandora.test.FastTest'
+        }
+    }
+
+    task integrationTest(type: Test) {
+        useJUnit {
+            includeCategories 'com.wikia.pandora.test.IntegrationTest'
+            excludeCategories 'com.wikia.pandora.test.FastTest'
```
sqreek: Is this line is neccessery?

nandy-andy: I don't think so. I'll remove it.

pchojnacki:
I like this
```
task integrationTest(type: Test) {
  useJUnit {
    includeCategories 'com.wikia.pandora.test.IntegrationTest'
    excludeCategories 'com.wikia.pandora.test.FastTest'
  }
}
```

much better. I think it might be best to go with this approach. A

pchojnacki:
Also since tasks like 'check' depend on 'test' I'm starting to thing if maybe main test task should be cut down to non integration tests
and instead have task called e.g. allTests ?

nandy-andy:
OK, guys (@sqreek @nmonterroso @pchojnacki). Are we good to merge it? If yes, I'll add `integrationTest` task to Jenkins and I think the story is done :smile_cat: 

We were worried because of "experimental" part in the package name `org.junit.experimental.categories.Category` but there are lots of blog post about it, wiki page on GitHub project and Paweł found out this experimental feature is 5 years old already. We're pretty sure it won't hurt us if we use it. @nmonterroso @drsnyder are you fine with this?

build.gradle
```diff
@@ -8,6 +8,14 @@ subprojects {
...
     task unitTest(type: Test) {
-        exclude '**/integration/**'
+        useJUnit {
+            excludeCategories 'com.wikia.pandora.core.test.IntegrationTest'
+        }
+    }
+
+    task integrationTest(type: Test) {
+        useJUnit {
+            includeCategories 'com.wikia.pandora.core.test.IntegrationTest'
```
sqreek: is this run UnitTests also? is this should run UnitTests?

nandy-andy: it shouldn't and it doesn't :)

cool trick how to check it:
```
tasks.withType(Test) {
  testLogging {
    events 'passed'
  }
}
```

sqreek:
lgtm

PR #45, opened 2015-02-19
add cobertura plugin

sqreek:
Add gradle task for cobertura.
I realy don't know why it is work only when coberturaRuntime and dependencies has 2 different version.
when both have 1.9.4, there is error while compile (not working at java 1.8 issue solved in later versions)
when both have 2.0.3 compile error does not appear, but no coverage was done. The cobertura.xml file was created but without any data.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,6 +4,7 @@ subprojects {
         mavenCentral();
     }
     apply plugin: 'java'
+    apply from: "${parent.projectDir.canonicalPath}/gradle/cobertura.gradle"
     sourceCompatibility = 1.8
     version = "0.1.0"
 
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ subprojects {
         mavenCentral();
     }
     apply plugin: 'java'
-    apply from: "${parent.projectDir.canonicalPath}/gradle/cobertura.gradle"
+    apply from: rootProject.file("gradle/cobertura.gradle")
     sourceCompatibility = 1.8
     version = "0.1.0"
 
--- a/gradle/cobertura.gradle
+++ b/gradle/cobertura.gradle
@@ -1,0 +1,53 @@
+logger.info "Configuring Cobertura Plugin"
+
+configurations{
+    coberturaRuntime {extendsFrom testRuntime}
+}
+
+dependencies {
+    coberturaRuntime 'net.sourceforge.cobertura:cobertura:1.9.4'
+}
+
+def serFile="${project.buildDir}/cobertura.ser"
+def classes="${project.buildDir}/classes/main"
+def classesCopy="${classes}-copy"
+
+
+task cobertura(type: Test){
+    dependencies {
+        testRuntime 'net.sourceforge.cobertura:cobertura:2.0.3'
+    }
+
+    systemProperty "net.sourceforge.cobertura.datafile", serFile
+    exclude '**/integration/**'
+
+}
+
+cobertura.doFirst  {
+    logger.quiet "Instrumenting classes for Cobertura"
+    ant {
+        delete(file:serFile, failonerror:false)
+        delete(dir: classesCopy, failonerror:false)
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -4,6 +4,7 @@ subprojects {
         mavenCentral();
     }
     apply plugin: 'java'
+    apply from: "${parent.projectDir.canonicalPath}/gradle/cobertura.gradle"
```
pchojnacki: rootProject.file("gradle/cobertura.gradle") is nicer

gradle/cobertura.gradle
```diff
@@ -0,0 +1,53 @@
...
+        testRuntime 'net.sourceforge.cobertura:cobertura:2.0.3'
+    }
+
+    systemProperty "net.sourceforge.cobertura.datafile", serFile
+    exclude '**/integration/**'
+
+}
+
+cobertura.doFirst  {
+    logger.quiet "Instrumenting classes for Cobertura"
```
pchojnacki: why quiet ?

gradle/cobertura.gradle
```diff
@@ -0,0 +1,53 @@
+logger.info "Configuring Cobertura Plugin"
+
+configurations{
+    coberturaRuntime {extendsFrom testRuntime}
+}
+
+dependencies {
+    coberturaRuntime 'net.sourceforge.cobertura:cobertura:1.9.4'
```
pchojnacki: shouldn't it be 2.0.3 ?

pchojnacki:
also this: https://github.com/cobertura/cobertura/issues/166
suggest that cobertura 2.1.0 will support java 8 properly
you might want to try using version 2.1.0-SNAPSHOT before 2.1.0 is released

gradle/cobertura.gradle
pchojnacki: oh i read the comment

pchojnacki:
also there is alread y a gradle plugin for cobertura coverages
and it seems that at least their latest versions supports cobertura 2.1.0
https://github.com/stevesaliman/gradle-cobertura-plugin/blob/Release_2.2.6/build.gradle

gradle/cobertura.gradle
sqreek: QUIET    Important information messages
https://gradle.org/docs/current/userguide/logging.html

sqreek:
I'm vote for waiting until 2.1.0 will have stable release. After that, I think we can use this plugin then

nandy-andy:
Could you merge it with recent `origin/master`? After resolving conflicts the checks should become green :)

pchojnacki:
You can also try using 2.1.0-SNAPSHOT version Internets say it should work

sqreek:
Could not find net.sourceforge.cobertura:cobertura:2.1.0-SNAPSHOT.
Also I find, sometimes they put unstable version there.
If there is no other issue I will merge it

pchojnacki:
Does the current setup provide working coverage reports ?

sqreek:
yes

gradle/cobertura.gradle
```diff
@@ -0,0 +1,53 @@
...
+def classesCopy="${classes}-copy"
+
+
+task cobertura(type: Test){
+    dependencies {
+        testRuntime 'net.sourceforge.cobertura:cobertura:2.0.3'
+    }
+
+    systemProperty "net.sourceforge.cobertura.datafile", serFile
+    exclude '**/integration/**'
```
pchojnacki: you can use new notation introduced by Nandy, additionally I think it would be best to be able to get repoerts from full test run from cobertura, at least on Jenkins. So maybe two tests configuration would be better

pchojnacki:
One comment otherwise LGTM

gradle/cobertura.gradle
sqreek: I will change it on other PR.

pchojnacki: Why ?

PR #46, opened 2015-02-20
Sonar issues cleanup

nandy-andy:
Clean up of most of the major issues in [sonar](http://sonar-s1/drilldown/issues/3508?&rule=squid%3AS1444&rule_sev=MAJOR).

### DIFF
--- a/mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java
+++ b/mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java
@@ -25,9 +25,10 @@ import io.dropwizard.setup.Environment;
 public class MobileConfigApplication extends Application<MobileConfigConfiguration> {
 
   public static final RepresentationFactory
-      representationFactory =
+      REPRESENTATION_FACTORY =
       new StandardRepresentationFactory();
-  public static final Logger logger = LoggerFactory.getLogger(FileConfigurationService.class);
+
+  public static final Logger LOGGER = LoggerFactory.getLogger(FileConfigurationService.class);
 
   public static void main(String[] args) throws Exception {
     new MobileConfigApplication().run(args);
--- a/mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java
+++ b/mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java
@@ -28,7 +28,7 @@ public class MobileConfigApplication extends Application<MobileConfigConfigurati
       REPRESENTATION_FACTORY =
       new StandardRepresentationFactory();
 
-  public static final Logger LOGGER = LoggerFactory.getLogger(FileConfigurationService.class);
+  public static final Logger LOGGER = LoggerFactory.getLogger(MobileConfigApplication.class);
 
   public static void main(String[] args) throws Exception {
     new MobileConfigApplication().run(args);
@@ -80,6 +80,7 @@ public class MobileConfigApplication extends Application<MobileConfigConfigurati
     final ApplicationsResource
         appList =
         new ApplicationsResource(listService);
+
     environment.jersey().register(mobileConfig);
     environment.jersey().register(image);
     environment.jersey().register(appList);
--- a/mobile-config-service/src/main/java/com/wikia/mobileconfig/gateway/AppsDeployerList.java
+++ b/mobile-config-service/src/main/java/com/wikia/mobileconfig/gateway/AppsDeployerList.java
@@ -4,6 +4,7 @@ import com.google.common.base.Optional;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.wikia.mobileconfig.MobileConfigApplication;
 import com.wikia.mobileconfig.MobileConfigConfiguration;
 
 import org.apache.http.client.ClientProtocolException;
@@ -113,6 +114,9 @@ public class AppsDeployerList implements AppsListService {
 
       return true;
     } catch (ClientProtocolException exception) {
+      MobileConfigApplication.LOGGER.error(
+          "Apps deployer host is unreachable", exception
+      );
       return false;
     }
   }
--- a/mobile-config-service/src/main/java/com/wikia/mobileconfig/gateway/AppsDeployerList.java
+++ b/mobile-config-service/src/main/java/com/wikia/mobileconfig/gateway/AppsDeployerList.java
@@ -105,6 +105,8 @@ public class AppsDeployerList implements AppsListService {
 
   @Override
   public boolean isUp() throws IOException {
+    Boolean result = false;
+
     try {
       String appsDeployerHealthCheckUrl = String.format(
           APPS_DEPLOYER_HEALTH_CHECK_URL_FORMAT,
@@ -112,12 +114,13 @@ public class AppsDeployerList implements AppsListService {
       );
       this.executeHttpRequest(appsDeployerHealthCheckUrl);
 
-      return true;
+      result = true;
     } catch (ClientProtocolException exception) {
       MobileConfigApplication.LOGGER.error(
           "Apps deployer host is unreachable", exception
       );
-      return false;
     }
+
+    return result;
   }
 }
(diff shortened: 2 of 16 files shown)


nandy-andy:
We had 48 (33 major, 15 minor) issues with those changes 25 left (9 major, 16 minor).

mobile-config-service/src/main/java/com/wikia/mobileconfig/gateway/AppsDeployerList.java
```diff
@@ -113,6 +114,9 @@ public boolean isUp() throws IOException {
 
       return true;
```
sqreek: In my opinion this should have variable for return, and only one return point. It's more readable

mobile-config-service/src/main/java/com/wikia/mobileconfig/resources/MobileConfigResource.java
```diff
@@ -39,23 +48,53 @@ public Representation getMobileApplicationConfig(
       @PathParam("app-tag") String appTag,
       @QueryParam("ui-lang") String uiLang,
       @QueryParam("content-lang") String contentLang
-  ) throws java.io.IOException, MobileConfigException {
+  ) throws MobileConfigException {
+    Boolean isValidAppTag = isValidAppTag(platform, appTag);
 
-    if (!this.appsList.isValidAppTag(platform, appTag)) {
+    if (!isValidAppTag) {
       throw new InvalidApplicationTagException(appTag);
```
sqreek: here you are throw new exception, but loose main cause (stacktrace) why exception was thrown.
Yes, it's logged, but its harder to debug this way, instead having one exception with full stacktrace

mobile-config-service/src/main/java/com/wikia/mobileconfig/resources/MobileConfigResource.java
```diff
@@ -39,23 +48,53 @@ public Representation getMobileApplicationConfig(
...
     }
 
-    MobileConfiguration configuration =
-        this.appConfiguration.getConfiguration(platform, appTag, uiLang, contentLang);
+    MobileConfiguration configuration = getConfiguration(platform, appTag, uiLang, contentLang);
 
     if (configuration.getModules().isEmpty()) {
       throw new ConfigurationNotFoundException(platform, appTag);
```
sqreek: similar issue like above

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/FileConfigurationService.java
```diff
@@ -46,8 +49,8 @@ public MobileConfiguration getConfiguration(String platform, String appTag, Stri
 
       return configuration;
     } catch (IOException e) {
-      MobileConfigApplication.logger.info(
-          String.format(CONFIGURATION_NOT_FOUND_DEBUG_MESSAGE_FORMAT, appTag, platform)
+      MobileConfigApplication.LOGGER.info(
+          CONFIGURATION_FOR_APP_TAG_NOT_FOUND_DEBUG_MESSAGE_FORMAT, appTag, platform, e
```
sqreek: As of SLF4J version 1.6, SLF4J will interpret the last parameter as you intended.
Can you check is exception is actual log as exception, and not used in String.format?

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/FileConfigurationService.java
```diff
@@ -46,8 +49,8 @@ public MobileConfiguration getConfiguration(String platform, String appTag, Stri
 
       return configuration;
```
sqreek: Can you exctract this to variable and use only one return point?

mobile-config-service/src/main/java/com/wikia/mobileconfig/service/FileConfigurationService.java
nandy-andy: As we discussed it offline it seems the best solution would be to go with `String.format` as the first parameter and then exception object as the second. I'll do it :+1:

mobile-config-service/src/main/java/com/wikia/mobileconfig/resources/MobileConfigResource.java
nandy-andy: Yeah, that was the hardest Sonar issue I was struggling with ;) Here is the Sonar's rule and examples how to solve it:
> Public methods should throw at most one checked exception
> 
> Using checked exceptions forces method callers to deal with errors, either by propagating them or by handling them. This makes those exceptions fully part of the API of the method.
> To keep the complexity for callers reasonable, methods should not throw more than one kind of checked exception.
> The following code:
```java
public void delete() throws IOException, SQLException {      // Non-Compliant
  /* ... */
}
```
> should be refactored into:
```java
public void delete() throws SomeApplicationLevelException {  // Compliant
  /* ... */
}
```
> Overriding methods are not checked by this rule and are allowed to throw several checked exceptions.

Do you think I should revert my way of solving the issues and it's a rule we should discuss if we want to follow?

mobile-config-service/src/test/java/com/wikia/mobileconfig/resources/ApplicationsResourceTest.java
```diff
@@ -26,13 +26,14 @@
  * Unit tests for {@link ApplicationsResource}.
  */
 public class ApplicationsResourceTest {
-  private static final HttpClient httpClientMock = mock(HttpClient.class);
```
pchojnacki: This looks awkward.
I'd suggest using a method over capitalized 'const'.

mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java
```diff
@@ -25,9 +25,10 @@
 public class MobileConfigApplication extends Application<MobileConfigConfiguration> {
 
   public static final RepresentationFactory
-      representationFactory =
+      REPRESENTATION_FACTORY =
       new StandardRepresentationFactory();
-  public static final Logger logger = LoggerFactory.getLogger(FileConfigurationService.class);
+
+  public static final Logger LOGGER = LoggerFactory.getLogger(MobileConfigApplication.class);
```
pchojnacki: This also seem awkward and out of place.
private Logger is more standard way of handling logging.

mobile-config-service/src/main/java/com/wikia/mobileconfig/resources/MobileConfigResource.java
pchojnacki: See @cmiller-wikia  addition to our coding guidelines:
https://github.com/Wikia/guidelines/tree/master/Java

on checked exceptions should be used sparingly.
Also I see you have like 4 different exceptions declared and thrown.

IMO we shouldn't give Java critics any more ammo (oververbosity, too much code). My proposal is to unify those exceptions to hopefully one exception where you will just modify the message appropriately. You can always inspect the message in the tests using junit rule http://junit.org/javadoc/latest/org/junit/rules/ExpectedException.html

mobile-config-service/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java
pchojnacki: hmm, more like lowercase logger is more standard :) It was just am matter of LOGGER looking baad to me.
I did small research and SO article pointed me here:
http://slf4j.org/faq.html#declared_static

There seems to be some merit to both approaches.

I think if you are not creating thousands of class instances per request then I think having non static version might be better.

nandy-andy: Interesting reading, thanks! :smile_cat: Yeah, I don't think we're creating thousands of class instances... But I'm also not sure if it's worth changing it ATM. The point of this PR is to fix issues found by Sonar. 

There was two SonarQube issues here: 1) naming convention - we should use upper-cases for constants 2) loggers should be private static. I would leave it as it is for now because the Sonar issues are solved and we safe some CPU and memory usage additionally. If the team thinks the logger should be an instance variable then I suggest creating new story to make two changes: in code and in SonarQube rules.

pchojnacki:
LGTM for current Sonar Rules

I think we need to go over sonar Rules Upper case LOGGER just looks strange

PR #47, opened 2015-02-20
change exclude integration tests approach

sqreek:
Use excludeCategories using interface instead exclude using regular expression


pchojnacki:
LGTM

PR #49, opened 2015-02-20
Increase test coverage demo

sqreek:
Increase test coverage, and remove unused URIBuilder


pchojnacki:
More tests and less library code - Awesome
LGTM

PR #50, opened 2015-02-23
refactor RegexFilter, add unit tests to RegexFilter

sqreek:
Simplifier RegexFilter
Add null option instead empty ImmutableSet
Add excludeFromInclude option to explicit behaviour
Use one Pattern combine with | (pipe) instead of pattern list
Add unit tests

When adding tests to regexFilter I found logical [mistake](https://github.com/Wikia/pandora/compare/Add_unit_test_to_regex_filter?expand=1#diff-16d09b474e40661297ae6608ba025619L57), and [inconsistent in logic](https://github.com/Wikia/pandora/compare/Add_unit_test_to_regex_filter?expand=1#diff-16d09b474e40661297ae6608ba025619L53) (no explicit is exclude excludes from includes or vice versa)

//cc @nmonterroso

### DIFF
--- a/pandora-core/src/main/java/com/wikia/pandora/core/metrics/RegexFilter.java
+++ b/pandora-core/src/main/java/com/wikia/pandora/core/metrics/RegexFilter.java
@@ -5,58 +5,67 @@ import com.google.common.collect.ImmutableSet;
 import com.codahale.metrics.Metric;
 import com.codahale.metrics.MetricFilter;
 
-import java.util.ArrayList;
 import java.util.regex.Pattern;
 
 public class RegexFilter implements MetricFilter {
-  private ArrayList<Pattern> includes;
-  private ArrayList<Pattern> excludes;
+
+  private Pattern includes;
+  private Pattern excludes;
+  private boolean excludeFromInclude;
 
   public RegexFilter(ImmutableSet<String> includes, ImmutableSet<String> excludes) {
+    this(includes, excludes, false);
+  }
+
+  /**
+   * @param includes           includes patterns
+   * @param excludes           excludes patterns
+   * @param excludeFromInclude if true, exclude from include, if false include from exclude
+   */
+  public RegexFilter(ImmutableSet<String> includes, ImmutableSet<String> excludes,
+                     boolean excludeFromInclude) {
+    this.excludeFromInclude = excludeFromInclude;
     this.includes = compilePatterns(includes);
     this.excludes = compilePatterns(excludes);
   }
 
-  private ArrayList<Pattern> compilePatterns(ImmutableSet<String> patternStrings) {
-    ArrayList<Pattern> patterns = new ArrayList<>();
-    patternStrings.forEach(s -> patterns.add(Pattern.compile(s)));
-
-    return patterns;
+  private Pattern compilePatterns(ImmutableSet<String> patternStrings) {
+    Pattern pattern = null;
+    if (patternStrings != null && !patternStrings.isEmpty()) {
+      String combinePattern = String.join("|", patternStrings);
+      pattern = Pattern.compile(combinePattern);
+    }
+    return pattern;
   }
 
   private boolean includesMatches(String name) {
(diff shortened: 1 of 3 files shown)


nmonterroso:
Thanks for catching that mistake! I took the logic from `BaseReporterFactory` in `io.dropwizard.metrics`:

```java
    /**
     * Gets a {@link MetricFilter} that specifically includes and excludes configured metrics.
     * <p/>
     * Filtering works in 3 ways:
     * <dl>
     *     <dt><i>excludes</i>-only</dt>
     *     <dd>All metrics are reported, except those with a name listed in <i>excludes</i>.</dd>
     *     <dt><i>includes</i>-only</dt>
     *     <dd>No metrics are reported, except those with a name listed in <i>includes</i>.</dd>
     *     <dt>mixed (both <i>includes</i> and <i>excludes</i></dt>
     *     <dd>All metrics are reported, except those with a name listed in <i>excludes</i>, unless
     *     they're also listed in <i>includes</i> (<i>includes</i> takes precedence).</dd>
     * </dl>
     *
     * @return the filter for selecting metrics based on the configured excludes/includes.
     * @see #getIncludes()
     * @see #getExcludes()
     */
    public MetricFilter getFilter() {
        return new MetricFilter() {
            @Override
            public boolean matches(final String name, final Metric metric) {
                boolean useIncl = !getIncludes().isEmpty();
                boolean useExcl = !getExcludes().isEmpty();

                if (useIncl && useExcl) {
                    return getIncludes().contains(name) || !getExcludes().contains(name);
                }
                else if (useIncl && !useExcl) {
                    return getIncludes().contains(name);
                }
                else if (!useIncl && useExcl) {
                    return !getExcludes().contains(name);
                }
                else {
                    return true;
                }
            }
        };
    }
```

pandora-core/src/main/java/com/wikia/pandora/core/metrics/RegexFilter.java
```diff
@@ -5,58 +5,67 @@
...
-import java.util.ArrayList;
 import java.util.regex.Pattern;
 
 public class RegexFilter implements MetricFilter {
-  private ArrayList<Pattern> includes;
-  private ArrayList<Pattern> excludes;
+
+  private Pattern includes;
+  private Pattern excludes;
+  private boolean excludeFromInclude;
```
nmonterroso: What's the point of this variable? It's never set anywhere, so it's always `false`.

pandora-core/src/main/java/com/wikia/pandora/core/metrics/RegexFilter.java
```diff
@@ -5,58 +5,67 @@
...
 
   public RegexFilter(ImmutableSet<String> includes, ImmutableSet<String> excludes) {
+    this(includes, excludes, false);
+  }
+
+  /**
+   * @param includes           includes patterns
+   * @param excludes           excludes patterns
+   * @param excludeFromInclude if true, exclude from include, if false include from exclude
```
nmonterroso: This description is weird.. what does it mean to "exclude the include" or to "include from exclude"?

pandora-core/src/main/java/com/wikia/pandora/core/metrics/RegexFilter.java
sqreek: We are not setting in our code yet. You can set it (here) (https://github.com/Wikia/pandora/pull/50/files#diff-16d09b474e40661297ae6608ba025619R26)

pandora-core/src/main/java/com/wikia/pandora/core/metrics/RegexFilter.java
sqreek: My mistake. I will provide proper explanation as comment.
Simplest way is:
A - includes
B - excludes
Z - all
1 case) A-B
2 case) (Z-B)+A

nmonterroso:
LGTM

PR #51, opened 2015-02-23
Environment Variables From Consul

nmonterroso:
@pchojnacki 
https://wikia-inc.atlassian.net/browse/SERVICES-61

Introduce project `consul-config` with class `ConsulAppConfig` that reads an app's configs from Consul. This is to be used in the [marathon plugin](https://github.com/Wikia/marathon_gradle_plugin) to set the correct environment variables when the application is launched. I've created configs [in consul](http://consul.service.sjc.consul:8500/ui/#/sjc/kv/config/mobile-config/prod/) for the production deployment of the mobile-config service, and have updated the sample yaml file to source from these environment variables


nmonterroso:
Spoke with @pchojnacki - gonna change this to load configs from consul during boot, from the same consulConfig as the heartbeat stuff. This will require an addition to `consulConfig` to have an `app` field, and specify the `env` from CLI

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -0,0 +1,83 @@
...
+    if (key.length() <= remove.length()) {
+      return null;
+    }
+
+    return key.substring(remove.length());
+  }
+
+  public static ConsulKeyValueConfig fromEnv() {
+    try {
+      URI uri = new URI(System.getenv(ENV_KEY));
```
nmonterroso: The reason this needs to be passed in via env config and doesn't use the same consul as defined in `application.yaml` is because the substitution from consul -> file happens before the file is parsed into an object, so we don't have access to those values.

pandora-core/src/main/java/com/wikia/pandora/core/consul/config/ConsulConfigurationFactory.java
```diff
@@ -0,0 +1,361 @@
...
+
+import io.dropwizard.configuration.ConfigurationException;
+import io.dropwizard.configuration.ConfigurationFactory;
+import io.dropwizard.configuration.ConfigurationSourceProvider;
+import io.dropwizard.configuration.ConfigurationValidationException;
+import io.dropwizard.configuration.FileConfigurationSourceProvider;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class ConsulConfigurationFactory<T> extends ConfigurationFactory<T> {
```
nmonterroso: This class is mostly a copy/paste of ConfigurationFactory, with the only changes being the loading of consul variables. This was done because of private methods/members

pandora-core/src/main/java/com/wikia/pandora/core/consul/config/ConsulConfigurationFactory.java
```diff
@@ -0,0 +1,361 @@
...
+          .setDetail(e.getMessage())
+          .setFieldPath(e.getPath())
+          .setLocation(e.getLocation())
+          .setCause(e)
+          .build(path);
+    }
+  }
+
+  private void replaceCosulVariables(final JsonNode root) {
+    Queue<JsonNode> q = Queues.newArrayDeque();
```
nmonterroso: The `replaceConsul*` and `getReplacementForValue` methods are the real additions here.

pandora-core/src/main/java/com/wikia/pandora/core/consul/config/ConsulConfigurationParsingException.java
```diff
@@ -0,0 +1,379 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+
+import io.dropwizard.configuration.ConfigurationException;
+
+public class ConsulConfigurationParsingException extends ConfigurationException {
```
nmonterroso: Copy/paste of `ConfigurationException`

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -0,0 +1,83 @@
...
+    try {
+      consul.keyValueClient().getValues(configFolder).stream()
+          .filter(v -> getConfigKey(v.getKey()) != null)
+          .forEach(v -> {
+            String key = getConfigKey(v.getKey());
+            String value = ClientUtil.decodeBase64(v.getValue());
+            config.put(key, value);
+          });
+    } catch (Exception e) {
+      return null;
```
sqreek: This can provide null pointer exception, maybe returning empty config will be better?
Also this is second return point.

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -0,0 +1,83 @@
...
+
+  public String get(String key) {
+    return getConfig().get(key);
+  }
+
+  protected String getConfigKey(String key) {
+    String remove = configFolder + "/";
+
+    if (key.length() <= remove.length()) {
+      return null;
```
sqreek: if this will return null more than once, will provide strange behaviour in config

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -0,0 +1,83 @@
...
+      return new ConsulKeyValueConfig(uri.getHost(),
+                                      uri.getPort(),
+                                      uri.getPath().substring(1));
+    } catch (Exception e) {
+      return null;
+    }
+  }
+
+  protected static String getConfigFolder(String app, String env) {
+    return "config/"+app+"/"+env;
```
sqreek: String.format looks better :)

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
nmonterroso: Ah yeah, you're right. I had this return `null` when I was thinking of implementing this a different way. I'll change it to a single return statement with an empty hashmap.

pandora-core/src/main/java/com/wikia/pandora/core/consul/config/ConsulConfigurationFactory.java
sqreek: Big NO.
1) just Copy/paste is bad on start.
2) You are override every public method, and also inherit from ConfigurationFactory.

I recommend to try only override methods with entry point to private method, 
or remove extends (becuse its override everything anyway) and give good arguments why copy/paste is good here.

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
nmonterroso: How so? This should only be true when the key is equal to the config folder, which is true at least once because the queried key is returned as part of the set of `Value`s

pandora-core/src/main/java/com/wikia/pandora/core/consul/config/ConsulConfigurationFactory.java
nmonterroso: 1. Pretty much every method calls into the private `build` method.
2. This has to extend `ConfigurationFactory`, because that's what the interface `ConfigurationFactoryFactory.build` says it will return, and that's what `bootstrap.setConfigurationFactoryFactory` expects as input.

I wouldn't have just copy/pasted without reason.

nmonterroso: Thanks for making me rethink this. I'm gonna try doing `bootstrap.setConfigurationSourceProvider` to do this instead. [Here's](https://github.com/minisu/Dropwizard-EnvVar-Interpolation) a project that does something similar.

nmonterroso:
@sqreek - changed the way the replacements is done. Take another look, please.

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -0,0 +1,87 @@
...
+      return new ConsulKeyValueConfig(uri.getHost(),
+                                      uri.getPort(),
+                                      uri.getPath().substring(1));
+    } catch (Exception e) {
+      return null;
+    }
+  }
+
+  protected static String getConfigFolder(String app, String env) {
+    return String.format("config/%s/%s", app, env);
```
pchojnacki: I don't like that you cannot change the prefix 'config' here Also I actually wonder should we really be this opinionated with regards to config path in consul.

That is I think that one single variable called e.g. prefix could suffice for first iteration of this library. Then once usage patterns emerge we can codify them.

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -0,0 +1,87 @@
...
+      consul.keyValueClient().getValues(configFolder).stream()
+          .filter(v -> getConfigKey(v.getKey()) != null)
+          .forEach(v -> {
+            String key = getConfigKey(v.getKey());
+            String value = ClientUtil.decodeBase64(v.getValue());
+            config.put(key, value);
+          });
+    } catch (Exception e) {
+      Logger logger = LoggerFactory.getLogger(ConsulKeyValueConfig.class);
+      logger.warn("Count not fetch consul key/value configs", e);
```
pchojnacki: Logo

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -0,0 +1,87 @@
...
+  public HashMap<String, String> getConfig() {
+    if (config != null) {
+      return config;
+    }
+
+    config = new HashMap<>();
+
+    try {
+      consul.keyValueClient().getValues(configFolder).stream()
+          .filter(v -> getConfigKey(v.getKey()) != null)
```
pchojnacki: I think we should actually log when its == null. Because as far I can gather it shouldn't happen.

Also I think extractConfigKey might be better name.

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    return config;
+  }
+
+  public String get(String key) {
+    return getConfig().get(key);
+  }
+
+  protected String getConfigKey(String key) {
```
pchojnacki: I think this whole method could me implementend using some single 'standard' library call like apache StringUtils.replaceOnce

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
nmonterroso: It should happen exactly once. When you ask for keys in a folder `a/b/c`, it will return `a/b/c/key1`, `a/b/c/key2`, etc as you'd expect, but will also return `a/b/c`, so this should not be in the result set.

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
nmonterroso: This is meant to provide a default location for configs. You can also use the `ConsulKeyValueConfig(Consul consul, String configFolder)` constructor, in which case this method is never called (so maybe this should be called `getDefaultConfigFolder`).

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
pchojnacki: Ok.  Then since I think its the same thing sqreek mentioned maybe its worth putting clarification comment here.

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
nmonterroso: Hm, actually, I don't know why I didn't just use `String.replace()`

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
nmonterroso: Sure thing

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
pchojnacki: Or that

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
nmonterroso: Actually I'm going to ignore any key that ends with `/` (indicating it's a folder), since it's possible to request configs for a folder that has folders inside of it, ex: you request `a/b/c`, which has `a/b/c/key1` and `a/b/c/d/key2`, you'd expect your hashmap to be `key1=key1val` and `d/key2=key2val`

consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
pchojnacki: "Count not fetch..." should be "Could not fetch..." I think :)

nmonterroso: You're right - but I think the count would disapprove of the change.

pchojnacki: Yeah He really likes Counting

pchojnacki:
Looks great.
It seems that only some small things are left. Once that handled it LGTM

PR #52, opened 2015-02-23
allow consulConfig to be null in application.yaml

nmonterroso:
@pchojnacki @nandy-andy 
Currently, we require a properly configured `ConsulConfig` instance when launching the mobile-config-service. This makes it hard to get up and running (I ran into this myself, Charles did as well, and I imagine others will too). This change allows the consulConfig portion to be missing from the configuration entirely, and only register itself with the Consul server if configured, which should be the case for local development.

### DIFF
--- a/pandora-core/src/main/java/com/wikia/pandora/core/consul/ConsulBundle.java
+++ b/pandora-core/src/main/java/com/wikia/pandora/core/consul/ConsulBundle.java
@@ -64,6 +64,7 @@ public abstract class ConsulBundle<T> implements ConfiguredBundle<T> {
     ConsulConfig configuration = narrowConfig(appConfiguration);
 
     if (configuration == null) {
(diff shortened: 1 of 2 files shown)


pandora-core/src/main/java/com/wikia/pandora/core/consul/ConsulBundle.java
```diff
@@ -62,6 +62,11 @@ protected void scheduleHeartbeat(Environment environment, ConsulConfig config,
   @Override
   public void run(T appConfiguration, Environment environment) throws Exception {
     ConsulConfig configuration = narrowConfig(appConfiguration);
+
+    if (configuration == null) {
+      return;
```
pchojnacki: I think some log message describing that consul initialization has been skipped might be good idea.

pchojnacki:
One minor/optional thing otherwise it LGTM

PR #55, opened 2015-02-24
InteliJ inspection quick fix

sqreek:
add final keywords
shorter list initialization
optimize imports


nmonterroso:
IntelliJ tells you when a field can be declared as final? Didn't know that, but it's pretty cool!

LGTM

PR #56, opened 2015-02-26
global optimize imports

sqreek:
without MobileConfigApplication, becuse couse conflict with Service-61 branch and gradle files.


pchojnacki:
LGTM

PR #61, opened 2015-03-02
SERVICES-65 Deployment update

pchojnacki:
*Edit

This updates marathon gradle plugin to latest version plus introduces changes made to stage description Domain Specific Language.

Also I refactored our publishing configuration and Set current default version to Snapshot, and when we will release new versions we would have to make it so specifically.


jcellary:
Will LGTM after you provide description and ticket number :)

PR #62, opened 2015-03-02
SERVICES-61 Set proper configuration for mobile config service deployments

pchojnacki:
update config file and build scripts so that most configuration is taken from consul include consul config itself
/cc @jcellary


jcellary:
LGTM

PR #66, opened 2015-03-03
Remove the dependency on pandora-core in from the common-deps.gradle file

jogura:
This causes cyclical dependencies in `consul-config` on `pandora-core`.

To fix issues in IntelliJ with `consul-config` you may need to manually remove the `pandora-core` dependency.

@pchojnacki @drsnyder @nmonterroso


jogura:
I believe that PR #75 can and will fix this issue.

PR #69, opened 2015-03-05
add AppDeployerListContainer tests

sqreek:
add tests to AppDeployerListContainer.
Its same changes as previous but after reogranization


pchojnacki:
LGTM

PR #70, opened 2015-03-06
SERVICE-64 Add a 5 second maxage to CacheControl for mobile configs

drsnyder:
Add a 5 second maxage to CacheControl using the @CacheControl annotation.

https://wikia-inc.atlassian.net/browse/SERVICES-64

/cc @nmonterroso @nandy-andy @jcellary


sqreek:
LGTM

PR #71, opened 2015-03-06
Codebase cleanup + Guice introduction

pchojnacki:
- Make Integration Tests work with consul
- Centralize common build definitions
- Make configs more opinionated

Sorry for such all over the place commit. I can split it to 3 separate logical PR's if you guys prefer.
I just addressed bunch of stuff while working on other things.
Fixing things


gradle/marathon_deployment.gradle
```diff
@@ -50,7 +55,7 @@ deployments {
             }
         }
         marathon {
-            url = "http://mesos-s1.wikia-prod:8080"
+            url = marathonUrl
```
sqreek: mesos?

gradle/dropwizard_service.gradle
```diff
@@ -0,0 +1,21 @@
...
+                    "http://${testConsulHost}:8500/config/${project.name}/integration",
+    ]
+}
+
+run {
+    environment = [
+            CONSUL_KEY_VALUE_CONFIG_SOURCE:
+                    "http://${testConsulHost}:8500/config/${project.name}/integration",
+    ]
+    args 'server', "src/main/resources/${project.name}.yaml"
```
sqreek: extract src/main/resources as U did in [here](https://github.com/Wikia/pandora/pull/71/files#diff-9f7d944680459182acda248ba6e53ca0R5)

gradle/marathon_deployment.gradle
pchojnacki: Marathon and mesos just share host. On port 8080 a marathonUrl is running

service/example/src/main/java/com/wikia/exampleservice/ExampleModule.java
```diff
@@ -0,0 +1,23 @@
...
+
+  @Override
+  protected void configure() {
+    binder().install(new ConsulGuiceModule());
+  }
+
+
+  @Provides
+  public ConsulConfig providesConsulConfig(Provider<ExampleConfiguration> configuration) {
+    return configuration.get().getConsulConfig();
```
sqreek: Are we should use it as default in every service? Is so, move it to more generic place

gradle/marathon_deployment.gradle
sqreek: ok

service/example/src/main/java/com/wikia/exampleservice/ExampleModule.java
pchojnacki: I'm still not sure wher this config should be handled. I think it would be best to put this elsewhere. Right now since its actually only two services this boilerplate is not too combersome IMO.

service/example/src/main/java/com/wikia/exampleservice/ExampleModule.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.google.inject.AbstractModule;
+import com.google.inject.Provider;
+import com.google.inject.Provides;
+
+import com.wikia.exampleservice.configuration.ExampleConfiguration;
+import com.wikia.pandora.core.consul.ConsulConfig;
+import com.wikia.pandora.core.consul.ConsulGuiceModule;
+
+public class ExampleModule extends AbstractModule {
```
nmonterroso: When I first saw the name `ExampleModule` I thought it would be an example of how to implement a module, but looks like it provides a `ConsulConfig` instance. Maybe this should be renamed to `ConsulConfigModule`?

pchojnacki: >
> When I first saw the name ExampleModule I thought it would be an example of how to implement a module, but looks like it provides a ConsulConfig instance. Maybe this should be renamed to ConsulConfigModule?
>
> Hehe Good point, I was so caught up in following convention
ServiceName+Module that I haven't noticed that Example+Module actually
might mean something different. :)

So if I get it right it seems that there should be at least one 'Module'
per application using Guice. Right now there is only Consul config stuff
there but I think in the future (SQreek actually works on it) there will be
more stuff there.

Either way I'll try renaming it to something that is more clear.

lib/consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -10,8 +10,6 @@
 import java.util.HashMap;
 
 public class ConsulKeyValueConfig {
-  public static final String ENV_KEY = "CONSUL_KEY_VALUE_CONFIG_SOURCE";
```
drsnyder: How will this work after merge? Previously, when you wanted to dev locally you had to [set this](https://one.wikia-inc.com/wiki/Engineering/Mobile_Configuration_Service).

gradle/dropwizard_service.gradle
```diff
@@ -0,0 +1,22 @@
...
+                    "http://${testConsulHost}:8500/config/${project.name}/integration",
+    ]
+}
+
+def defaultConfigLocation="src/main/resources"
+run {
+    environment = [
+            CONSUL_KEY_VALUE_CONFIG_SOURCE:
+                    "http://${testConsulHost}:8500/config/${project.name}/integration",
+    ]
```
drsnyder: Oh, never mind regarding the comment below. I smell what you are cooking :).

lib/service-core/src/main/java/com/wikia/pandora/core/consul/ConsulGuiceModule.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.inject.Provides;
+
+import com.wikia.gradle.ConsulKeyValueConfig;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import javax.inject.Named;
+
+public class ConsulGuiceModule extends AbstractModule {
```
drsnyder: Is the `Guice` necessary? Is that a convention?

service/example/src/main/java/com/wikia/exampleservice/ExampleServiceApplication.java
```diff
@@ -20,18 +24,28 @@ public static void main(String[] args) throws Exception {
...
-    bootstrap.addBundle(new ConsulBundle<ExampleConfiguration>() {
-      @Override
-      protected ConsulConfig narrowConfig(ExampleConfiguration config) {
-        return config.getConsulConfig();
-      }
-    });
+    GuiceBundle<ExampleConfiguration> guiceBundle =
+        GuiceBundle.<ExampleConfiguration>newBuilder()
+            .addModule(new ExampleModule())
+            .setInjectorFactory(new GovernatorInjectorFactory())
```
drsnyder: Your configs... give them to me, now.

Give me your configs!

service/mobile-config/src/test/java/com/wikia/mobileconfig/resources/ApplicationsResourceTest.java
```diff
@@ -63,7 +50,21 @@ public void getAppList() throws IOException {
     }
 
     verify(HTTP_CLIENT_MOCK, times(1))
-        .execute(argThat(new HttpGetWithURL("http://test-domain/api/app-configuration/")), any(ResponseHandler.class));
+        .execute(argThat(new HttpGetWithURL("http://test-domain/api/app-configuration/")),
+                 any(ResponseHandler.class));
     reset(HTTP_CLIENT_MOCK);
   }
+
+  class HttpGetWithURL extends ArgumentMatcher<HttpGet> {
```
drsnyder: Minor: should this be in another file?

drsnyder:
LGTM

service/mobile-config/src/test/java/com/wikia/mobileconfig/resources/ApplicationsResourceTest.java
pchojnacki: auto reformat of code triggered this change. I left it in since it makes it more corresponding to our coding guidelines
and this PR already contains everything :)

pchojnacki: But yeah I think it should actually

PR #72, opened 2015-03-06
Rename the mobile-config sample YAML file to match service name

drsnyder:
The marathon [gradle plugin looks at the service name](https://github.com/Wikia/marathon_gradle_plugin/blob/c9947e276130b2643add51afac5eeb4d19a25d83/src/main/groovy/com/wikia/gradle/marathon/common/App.groovy#L45) to determine the config file name.

This PR renames the sample YAML to match the service name. If it doesn't the deploy will fail.

/cc @nmonterroso @pchojnacki


drsnyder:
This change is also in #71 so we can close this one when it's merged.

PR #73, opened 2015-03-07
Increasing Mobile-Config Code Coverage

nmonterroso:
@drsnyder @pchojnacki @sqreek 
Adding unit tests for classes the MCS uses that weren't being fully tested, but were partially tested due to requests being executed in other tests. This required some refactoring on a couple classes to add constructors that already had their dependencies initialized (so didn't have to initialize through config variables, etc). Also fixed a logic bug in `AppsDeployerListContainer.isUsingCache()`.


service/mobile-config/src/test/java/com/wikia/mobileconfig/health/AppsDeployerHealthCheckTest.java
```diff
@@ -0,0 +1,36 @@
...
+    when(service.isUp()).thenReturn(true);
+    Result result = healthCheck.check();
+    assertThat(result.isHealthy());
+  }
+
+  @Test
+  public void testUnhealthy() throws Exception {
+    when(service.isUp()).thenReturn(false);
+    Result result = healthCheck.check();
+    assertThat(!result.isHealthy());
```
sqreek: IMO assertTrue is better instead of assertThat

service/mobile-config/src/test/java/com/wikia/mobileconfig/resources/MobileConfigResourceTest.java
```diff
@@ -40,6 +41,46 @@
...
+    } catch (BadRequestException ex) {
+      // safely ignore
+    }
+    verify(APPS_LIST_MOCK, times(1)).isValidAppTag("test-platform", "test-app");
+    verify(HTTP_SERVICE_MOCK, never()).getConfiguration("test-platform", "test-app", "en-us", "en-us");
+    reset(APPS_LIST_MOCK, HTTP_SERVICE_MOCK);
+  }
+
+  @Test
+  public void getConfigurationException() throws IOException {
```
sqreek: Is this test should have a expect exception?

service/mobile-config/src/test/java/com/wikia/mobileconfig/service/HttpConfigurationServiceTest.java
```diff
@@ -0,0 +1,56 @@
...
+  @Before
+  public void initialize() {
+    httpClient = mock(HttpClient.class);
+    objectMapper = mock(ObjectMapper.class);
+    configService = new HttpConfigurationService(httpClient, objectMapper, cephDomain, cephPort);
+  }
+
+  @Test
+  public void getEmptyConfiguration() throws Exception {
+    when(httpClient.execute(any(HttpGet.class), any(BasicResponseHandler.class))).thenThrow(IOException.class);
```
sqreek: You can Use TestHelper to get HttpClient

sqreek:
GJ with rewriting constructors, and don't thought about it :+1:

service/mobile-config/src/test/java/com/wikia/mobileconfig/health/AppsDeployerHealthCheckTest.java
nmonterroso: Oops. You're right, I'll make that change.

service/mobile-config/src/test/java/com/wikia/mobileconfig/resources/MobileConfigResourceTest.java
nmonterroso: I don't think so, the `throws` here is because some of the methods in the test can throw exceptions. When the `HTTP_SERVICE_MOCK` throws an exception the request to MCS results in a 404.

service/mobile-config/src/test/java/com/wikia/mobileconfig/service/HttpConfigurationServiceTest.java
nmonterroso: Here, I didn't want to use `TestHelper.getHttpClientMock`, since that method takes a url and expected string response and creates the `HttpClient` mock for you. However, since the mock is shared across tests, it doesn't seem correct to put the expected request/response pair in the `@Before`. 

I think it's a good idea to extend `TestHelper` to have a `addMockRequestException`. I'll add that and update the PR.

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/HttpConfigurationService.java
```diff
@@ -33,12 +33,20 @@ public HttpConfigurationService(
...
-    this.httpClient = new HttpClientBuilder(environment)
-        .using(configuration.getHttpClientConfiguration())
-        .build("http-configuration-service");
-    this.mapper = new ObjectMapper();
-    this.cephDomain = configuration.getCephDomain();
-    this.cephPort = configuration.getCephPort();
+    this(new HttpClientBuilder(environment)
+             .using(configuration.getHttpClientConfiguration())
+             .build("http-configuration-service"),
+         new ObjectMapper(),
```
sqreek: Object mapper is always new instance, we can remove it from constructor parameter, and put creation in a deepest one

sqreek: Object mapper is always new instance, we can remove it from constructor parameter, and put creation in a deepest one

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/HttpConfigurationService.java
```diff
@@ -72,9 +80,7 @@ public MobileConfiguration getConfiguration(String platform, String appTag, Stri
           e
       );
 
-      if (configuration == null) {
-        configuration = getDefault(platform);
-      }
+      configuration = getDefault(platform);
```
sqreek: this overrite whatever [here](https://github.com/Wikia/pandora/pull/73/files#diff-dec4a884b8066a6a2721380ce4a4d5edR73) was set up

service/mobile-config/src/test/java/com/wikia/mobileconfig/resources/MobileConfigResourceTest.java
sqreek: I meant there is no assert in this function.
I think its should expect exception

service/mobile-config/src/test/java/com/wikia/mobileconfig/service/HttpConfigurationServiceTest.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  @Test
+  public void getDefaultConfiguration() throws Exception {
+    MobileConfiguration cfgMock = new ObjectMapper().readValue(
+        new File("src/test/resources/fixtures/test-platform:test-app.json"),
+        MobileConfiguration.class
+    );
+
+    when(objectMapper.readValue(anyString(), eq(MobileConfiguration.class))).thenReturn(cfgMock);
```
sqreek: We don't need to mock objectMapper.

service/mobile-config/src/test/java/com/wikia/mobileconfig/service/HttpConfigurationServiceTest.java
```diff
@@ -0,0 +1,57 @@
...
+  private HttpClient httpClient;
+  private ObjectMapper objectMapper;
+  private String cephDomain = "ceph-domain";
+  private String cephPort = "80";
+  private HttpConfigurationService configService;
+
+  @Before
+  public void initialize() {
+    httpClient = mock(HttpClient.class);
+    objectMapper = mock(ObjectMapper.class);
```
sqreek: We don't need to mock objectMapper.

service/mobile-config/src/test/java/com/wikia/mobileconfig/resources/MobileConfigResourceTest.java
nmonterroso: Ah OK, I see what you're saying. I modeled this test after others in the test class, and your reason applies to all of them. I'll update to use expected exceptions.

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/HttpConfigurationService.java
nmonterroso: Yah, but if there's an exception, `configuration` will always be null, since it's initialized that way. IntelliJ highlighted this as `Value 'configuration' is always null`, which seems correct to me, so I don't think we need this if check.

sqreek: yes, didn't saw thats actualy in bracket

sqreek:
LGTM

PR #74, opened 2015-03-07
Integrate Jedis into discussion service prototype

armonr:
RFCR @jogura 

This replaces static hash maps with data persisting on Redis.
Also some refactoring is done to make a common abstract data/service layer.


service/discussion/src/main/java/com/wikia/discussionservice/dao/ContentDAO.java
```diff
@@ -0,0 +1,95 @@
...
+import java.lang.Exception;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Optional;
+
+/**
+ * Created by armon on 3/2/15.
+ */
+public class ContentDAO {
+  private final JedisService jedisService = new JedisService();
```
jogura: We should attempt to inject this.

service/discussion/src/main/java/com/wikia/discussionservice/dao/ContentDAO.java
```diff
@@ -0,0 +1,95 @@
...
+        continue;
+      }
+    }
+
+    return contentList;
+  }
+
+  public <T extends Content> T getContent(int siteId, int contentId, Class<T> U) {
+    JsonParser jsonParser;
+    JsonFactory jParser = new JsonFactory();
```
jogura: minor: misleading variable name for the factory

service/discussion/src/main/java/com/wikia/discussionservice/dao/ContentDAO.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    return content;
+  }
+
+  public <T extends Content> void deleteContent(int siteId, int contentId, Class<T> U) {
+    String type;
+    try {
+      Method method = U.getMethod("getType");
+      type = method.invoke(null).toString();
```
jogura: Should consider extracting this try/catch block out. It is used in all of the methods in this class.

service/discussion/src/main/java/com/wikia/discussionservice/dao/ContentDAO.java
```diff
@@ -0,0 +1,95 @@
...
+    } catch (Exception ex) {
+      System.out.printf("Exception in ContentDao::getItems: %s", ex.getMessage());
+      return contentList;
+    }
+
+    ObjectMapper mapper = new ObjectMapper();
+    Optional<ArrayList<String>> stringArrayList = Optional.ofNullable(jedisService.getList(siteId, type));
+    for (String item : stringArrayList.get()) {
```
jogura: If there is a `null` in the `stringArrayList` then this will throw an exception. Consider using the `ifPresent` or wrapping the `for` with an `isPresent`check. I would also say use one of the fancier `map` or `flatmap` functions but I'm not confident in my understanding of them yet.

service/discussion/src/main/java/com/wikia/discussionservice/domain/Content.java
```diff
@@ -0,0 +1,13 @@
+package com.wikia.discussionservice.domain;
+
+/**
+ * Created by armon on 3/2/15.
+ */
+public class Content {
```
jogura: Consider making this a true Abstract class. Then you can remove the need to throw `ExceptionInInitializerError` below.

service/discussion/src/main/java/com/wikia/discussionservice/resources/PostResource.java
```diff
@@ -105,4 +106,27 @@ public Response createPost(@NotNull @PathParam("siteId") IntParam siteId,
             "Unable to create post for site id: %s on thread id: %s", siteId.get(), 
             post.getThreadId()), null, Response.Status.BAD_REQUEST);
   }
+
+  @DELETE
+  @Path("/{siteId}/posts/{postId}")
+  @Timed
+  public Response deleteThread(@NotNull @PathParam("siteId") IntParam siteId,
```
jogura: copy and paste error? method name should be `deletePost`?

service/discussion/src/main/java/com/wikia/discussionservice/services/ContentService.java
```diff
@@ -0,0 +1,31 @@
...
+ * Created by armon on 3/2/15.
+ */
+abstract public class ContentService {
+  @NonNull
+  private JedisService jedisService;
+
+  @Inject
+  public void setJedisService(JedisService jedisService) {
+    this.jedisService = jedisService;
+  }
```
jogura: This one is a little controversial but we should avoid field injection in favor of constructor injection.

[google for: setter injection evil](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=setter%20injection%20evil)

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
```diff
@@ -57,7 +57,7 @@ public ForumService(ThreadService threadService, ForumDAO forumDAO) {
   public Optional<Forum> updateForum(int siteId, Forum forum) {
     Optional<Forum> updatedForum = forumDAO.updateForum(siteId, forum);
     
-    if (updatedForum.isPresent()) {
+    if (updatedForum != null) {
```
jogura: Why are we checking for `null` instead of the `Optional` contains a value?

service/discussion/src/main/java/com/wikia/discussionservice/services/JedisService.java
```diff
@@ -0,0 +1,60 @@
...
+import redis.clients.jedis.ScanParams;
+import redis.clients.jedis.ScanResult;
+
+/**
+ * Created by armon on 2/27/15.
+ */
+public class JedisService implements DataStore {
+  private String host = "localhost";
+
+  private int port = 6379;
```
jogura: Are there plans to make these configurable in the future?

armonr: Yes, I will add a todo.
I had it in config originally, but my yaml file either fails on compile (unrecognized field) or does not get applied. I need to investigate this more.

jogura:
LGTM

PR #75, opened 2015-03-09
Services Core started to become too crowded

pchojnacki:
This is first step to keep the code as modular as possible.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,13 +22,13 @@ subprojects {
 
     task unitTest(type: Test) {
         useJUnit {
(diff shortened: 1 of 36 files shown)


nmonterroso:
The line between `dropwizard` and `service-core` is a bit blurry to me. For example, this PR moves the Consul bundle stuff to `dropwizard`, but leaves the metrics package (which is also dropwizard configuration) in `service-core`.

nmonterroso:
cool. more subprojects! LGTM

PR #76, opened 2015-03-09
Add Swagger as the API documentation tool

jogura:
* Add Swagger annotations to the Forum Resource (others will follow)
* Update build.gradle to include dropwizard-swagger and apache-common-lang 2.x requirement

To access use http://[service]:[port]/swagger

TODO:
* The `try now` functionality does not work, there seems to be an problem with
Swagger understanding JSON+HAL types, investigating what needs to be done to make this work

@jimmordino @armonr


service/discussion/src/main/java/com/wikia/discussionservice/resources/ForumResource.java
```diff
@@ -55,38 +80,55 @@ public Response getForums(@NotNull @PathParam("siteId") IntParam siteId,
...
+          required = false, defaultValue = "10")
+      @QueryParam("limit") @DefaultValue("10") IntParam limit,
+      @ApiParam(name = "offset", value = "The pagination index for the threads of this forum",
+          required = false, defaultValue = "1")
+      @QueryParam("offset") @DefaultValue("1") IntParam offset,
+      @ApiParam(name = "responseGroup",
+          value = "The responseGroup controls the level of details returned with this call",
+          required = false, defaultValue = "small", allowableValues = "small, large")
+      @QueryParam("responseGroup") @DefaultValue("small") String requestedResponseGroup,
+      @Context UriInfo uriInfo) {
```
jimmordino: Really not liking the annotations inline with the method arguments like this.  Makes the method declaration unreadable.  It may be the lesser of 2 evils, but I think this is ugly.

jimmordino: Maybe for the next hackathon we can write out-of-band (above method decl) annotations that will generate these inline annotations. :smile:

jogura: The engineer can use the Swagger UI to reverse engineer the annotations to get a better understanding of the code :stuck_out_tongue_closed_eyes:

jogura: I'll attempt to document the Thread Resource using a yaml file and see if that is easier/more maintainable.

service/discussion/src/main/java/com/wikia/discussionservice/resources/ForumResource.java
```diff
@@ -144,14 +195,17 @@ public Response deleteForum(@NotNull @PathParam("siteId") IntParam siteId,
 
   @PUT
   @Path("/{siteId}/forums")
+  @ApiOperation(value = "Update a forum for a site", notes = "If the forum does not exist it will be created")
+  @ApiResponses(value = {
+      @ApiResponse(code = 201, message = "Successful creation of a forum", response = Forum.class),
+      @ApiResponse(code = 204, message = "Successful update of a forum"),
```
armonr: Is this a common pattern for update to fallback on creation, or should this be a 404?

service/discussion/src/main/java/com/wikia/discussionservice/resources/ForumResource.java
armonr: +1 on @jimmordino's point about the unreadability of the method declaration. 15 lines for arguments alone :open_mouth:
But also understand if that's all we have to work with today.

service/discussion/src/main/java/com/wikia/discussionservice/resources/ForumResource.java
jogura: Straight from our [API Guidelines](https://github.com/Wikia/guidelines/blob/master/APIDesign/README.md)
>PUT: modify a resource at the specified URI with the representation provided. If the resource does not exist it will be created. PUT operations need to be idempotent.

armonr: :+1:

armonr:
LGTM

jimmordino:
I can't say it LG, but given that we're going to user Swagger annotations it seems like this is what we're in for.  Having the description text in the argument list is really awful, though.  Seems like Type Annotations shouldn't be allowed to include elements for the sake of readability.  Oh well.

:shipit:

PR #77, opened 2015-03-09
bump to version 1.0, add option for ?frame query parameter with test

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-984

Java support for `frame` query parameter in https://github.com/Wikia/vignette/pull/79

### DIFF
--- a/lib/vignette/src/main/java/com/wikia/vignette/UrlGenerator.java
+++ b/lib/vignette/src/main/java/com/wikia/vignette/UrlGenerator.java
@@ -224,6 +224,16 @@ public class UrlGenerator {
       return this;
     }
 
+    /**
+     * for animated gifs, capture a specific frame of animation
+     * @param frame
+     * @return UrlGenerator.Builder
+     */
+    public Builder frame(int frame) {
+      queryParams.add(new BasicNameValuePair("frame", Integer.toString(frame)));
+      return this;
+    }
+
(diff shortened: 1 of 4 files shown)


drsnyder:
lgtm

lib/vignette/src/main/java/com/wikia/vignette/UrlGenerator.java
```diff
@@ -224,6 +224,24 @@ public Builder fill(String fill) {
       return this;
     }
 
+    /**
+     * for animated gifs, capture a specific frame of animation
+     * @param frame
+     * @return UrlGenerator.Builder
+     */
+    public Builder frame(int frame) {
```
drsnyder: Minor: any reason why you used `int` and not `Integer`?

nmonterroso: Why would you use `Integer` and not `int`? :P

drsnyder: The convention in the file seems to be the boxed types (`Integer` see `width` below). On the line below the `toString` method is also being used so I imagine the static `Integer.toString(frame)` is probably creating a boxed type underneath anyway. That line could also be simplified to `frame.toString()` if the boxed version was used.

/end nitpick

PR #78, opened 2015-03-10
SERVICES-146 rewrite Translator to be thread-safe

cmiller-wikia:
Fixed the Translator to:

1. Not use double-checked locking on initialization
2. Use a concurrency-safe collection


gradle/common_deps.gradle
```diff
@@ -11,3 +11,9 @@ dependencies {
     testCompile "org.mockito:mockito-core:$mockitoVersion"
     testCompile project(':lib:service-core')
 }
+
+test {
+  testLogging {
+    exceptionFormat = 'full'
```
cmiller-wikia: This isn't relevant to fixing the bug, but it's a nice quality-of-life thing that makes it easier to tell why your tests failed.

pchojnacki: :+1:

service/mobile-config/src/main/java/com/wikia/mobileconfig/utils/Translator.java
```diff
@@ -3,70 +3,99 @@
...
+    );
+  }
+
+  private Future<Map<String, String>> loadAndCacheTranslation(String langCode) {
+      CompletableFuture<Map<String, String>> futureTranslations = new CompletableFuture<>();
+      Future<Map<String, String>> existingFuture = translationCache.putIfAbsent(langCode, futureTranslations);
+
+      if (existingFuture == null) {
+        try {
+          futureTranslations.complete(loadTranslations(langCode));
```
pchojnacki:
~~~I might be missing something. But isn't all the caching and fetching already happening inside the same thread. If so why do we need to wrap in in the Future ?~~~
Ok I think I got it. You are storing the future fist in ConcurentHashMap, that way other users can block earlier.
It seems like a nice pattern.
:+1:

pchojnacki:
LGTM

service/mobile-config/src/main/java/com/wikia/mobileconfig/utils/Translator.java
```diff
@@ -3,70 +3,99 @@
...
-      "/translations/translations.%s.json";
+      "translations/translations.%s.json";
+
+  private final ConcurrentHashMap<String, Future<Map<String, String>>> translationCache = new ConcurrentHashMap<>();
 
-  private Map<String, Map<String, String>> translationCache;
+  private final ObjectMapper mapper = new ObjectMapper();
 
-  private ObjectMapper mapper = new ObjectMapper();
+  public static Translator getInstance() {
```
ArturKlajnerok: Isn't there a risk of unsynchronized access to instance ? I mostly used pattern with 
```java
public static synchronized InstanceClass getInstance() {
  if (sInstance == null) {
    sInstance = new InstanceClass();
  }
  return sInstance;
 }
```

pchojnacki: Right now the synchronization seems to be done through Futures and also thanks to using ConcurrentHashMap its not always needed.
So I think there is no need to Synchronize the access to this singleton.

service/mobile-config/src/main/java/com/wikia/mobileconfig/utils/Translator.java
cmiller-wikia: Yes, the order of operations is:

* Check the map to see if the translations are already in there (technically this is unnecessary, but the check is going to succeed the vast majority of the time and save us the more costly operations)
* If not, cons up a new future to load the translations into
* Try to put that future in the map with an atomic Compare and Set.
* If the CAS succeeds, load the translations and use them to complete the future
* If the CAS fails that means some other thread put their own future in the map since we last checked, so throw the one we created away and wait for that thread to finish loading the data instead

The advantage of doing it this way is that once the translations are loaded, you end up with a lock-free "happy path".

PR #79, opened 2015-03-10
Services 82 create way for qa

sqreek:
@pchojnacki @drsnyder @nmonterroso 
add qaDeploy task
extend guice impact in our code (extend guice module, add httpClient provider)
cleanup code (move appsService from gateway to services)
fix test-platform:test-apps (please dont use ":" someone still using windows)


service/mobile-config/src/main/java/com/wikia/mobileconfig/service/application/CephAppsListService.java
```diff
@@ -0,0 +1,88 @@
...
+      throw new CephAppsListServiceException("Not valid file format",
+                                             "File format is not valid, and could not be parsed");
+    } catch (IOException e) {
+      throw new CephAppsListServiceException("Not expected exception", e.getMessage());
+    }
+  }
+
+  @Override
+  public boolean isUp() throws IOException {
+    return true;
```
sqreek: please advice me what to test here becuse, Im out of idea right now.
Its mostly for qa, so if they change/delete file from ceph, checking ceph is not a good idea.

pchojnacki: Ceph is up is a fine I think. Maybe just don't check file availability. I think.

drsnyder: Performing a `GET` on `String.format("http://%:%/", cephDomain, cephPort)` and checking for a 404/200 would probably be better than defaulting to true. If the service is down I think you'll probably get an `IOException`.

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/application/AppsDeployerListContainer.java
```diff
@@ -1,12 +1,15 @@
-package com.wikia.mobileconfig.gateway;
+package com.wikia.mobileconfig.service.application;
```
drsnyder: Why the ns change from `gateway` to `service.application`?

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/application/AppsDeployerListContainer.java
```diff
@@ -31,26 +31,33 @@
...
     this(httpClient, appsDeployerDomain, DEFAULT_CACHE_TIME_IN_SEC);
   }
 
-  public AppsDeployerListContainer(Environment environment, MobileConfigConfiguration configuration) {
-    this(new HttpClientBuilder(environment)
-             .using(configuration.getHttpClientConfiguration())
-             .build("apps-deployer-list-service"),
-         configuration.getAppsDeployerDomain(),
+  @Inject
+  public AppsDeployerListContainer(Environment environment,
```
drsnyder: Should we remove the dependency on `Environment` since it's not being used? Can we do the same with the `MobileConfigConfiguration` since we only need the domain and cache time?

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/application/AppsDeployerListContainer.java
sqreek: IMO becuse, its have logic about cache, its seems valid to put it here instead gateway.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/bundle/ConsulBundle.java
```diff
@@ -97,6 +98,7 @@ public void run(Configuration configuration, Environment environment) throws Exc
                   logger.info("Scheduling Consul heartbeats");
                   scheduleHeartbeat(environment, this.configurationProvider.get(), heartbeat::send,
                                     true);
+                  logger.info("Heartbeats are scheduled");
```
pchojnacki: looks like debug lvl to me

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/bundle/ConsulBundle.java
```diff
@@ -88,6 +88,7 @@ public void run(Configuration configuration, Environment environment) throws Exc
                 .ifPresent((connector) -> {
                   logger.info("Registering in consul connector");
                   consul.registerConnector(connector);
+                  logger.info("Registered in consul");
```
pchojnacki: Should be debug lvl

service/mobile-config/build.gradle
```diff
@@ -19,12 +20,23 @@ deployments {
...
     testing {
         environment {
-            CONSUL_KEY_VALUE_CONFIG_SOURCE = "http://localhost:8500/config/mobile-config/testing"
+            CONSUL_KEY_VALUE_CONFIG_SOURCE = consulMobileConfigSourceUrl + "testing"
+        }
+    }
+
+    qa {
+        resources {
+            cpus = 1
```
pchojnacki: 0.1 cpu should be enough

service/mobile-config/src/main/java/com/wikia/mobileconfig/health/AppsDeployerHealthCheck.java
```diff
@@ -1,13 +1,17 @@
 package com.wikia.mobileconfig.health;
 
+
+import com.wikia.mobileconfig.service.application.AppsListService;
+
 import com.codahale.metrics.health.HealthCheck;
-import com.wikia.mobileconfig.gateway.AppsListService;
+import com.google.inject.Inject;
```
pchojnacki: outside of guice related code. You should use generice Java injection annotations.
http://docs.oracle.com/javaee/6/api/javax/inject/Inject.html

service/mobile-config/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java
```diff
@@ -41,53 +41,43 @@ public String getName() {
...
-    final AppsDeployerHealthCheck
-        appsDeployerHealthCheck =
-        new AppsDeployerHealthCheck(listService);
-    environment.healthChecks().register("apps-deployer", appsDeployerHealthCheck);
+    registerHealthChecks(environment, injector);
 
-    final MobileConfigResource
-        mobileConfig =
-        new MobileConfigResource(configService, listService);
+    final MobileConfigResource mobileConfig = injector.getInstance(MobileConfigResource.class);
```
pchojnacki: Resources and healthchecks can be autoregistered using .enableAutoConfig(...) if you want. That way there will be no code here what so ever.

service/mobile-config/src/main/java/com/wikia/mobileconfig/MobileConfigModule.java
```diff
@@ -1,18 +1,72 @@
...
+  }
+
+  @Provides
+  public AppsListService provideAppsListServiceFromConfiguration(
+      Injector injector,
+      MobileConfigConfiguration configuration) {
+    AppsListService appsListService = null;
+
+    ConsulConfiguration consulConfiguration = configuration.getConsulConfiguration();
+    if (consulConfiguration.getTags().contains("qa")) {
```
pchojnacki: I don't think this is the best way to handle this. Seems too implicit in my opinion.
I had an idea to explicitly state a 'stage' of the instance preferably using Environment variables.
This way application could only be in one stage at a time. But if you look at the collection of tags then app could be in multiple stages at the same time.
As for the way you inject different instance I don't have any better alternative but somehow I have some reservations. But until I have some alternative I think it can work.

You can make this provider rely on environment variable by also injecting @Named string here and having @Named provider.
@cmiller-wikia any thoughts on this ?

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/application/AppsDeployerListContainer.java
drsnyder: IMHO `service.application` doesn't have much meaning. FWIW I think `gateway` is more descriptive. It's a "gateway" to another service.

Also `service.application` collides with `service`. How do you know where you should look for an app specific service? You now have two directories with the same name.

sqreek: service catalog was created on start of MCS. there is no collides with service folder in root.
before this move there was two folder, gateway and service. behaviour of both of them was similliar so I decide to put it togheter, and I choose service (Its not controller becuse of state, and service is using in this case).

I agree with You about this connect with other service, and I thinking about extract CEPH part to gateway in core (its reusable). 

but also it should be done in another PR, becuse this is a different task.

@drsnyder 
Can You accept this temporary move and I will do this extract in near ocasion?

drsnyder: Sure. Please proceed.

pchojnacki:
Other than the Minor comments about naming it looks good.

PR #82, opened 2015-03-10
connection to redis with authentication

rafalkalinski:
A couple of new methods and properties to allow using password for Auth field when conencting to redis.


service/notifications/src/main/resources/notifications.yaml.sample
```diff
@@ -17,8 +17,12 @@ logging:
       customFields:
         appname: notifications
 redis:
-  # The redis server's address; required.
+  # The redis server's address in format: host:port; required.
   endpoint: localhost:6379
+  # Password - id needed for authorization; default: ""
+  password:
+  # URI scheme for redis connection; default: "redis"
+  scheme: redis
```
kvas-damian: IMO we should decide on one format.
Or we use full URI as a one field in config or we provide password in separate field and start using ```JedisPool``` contstructor that uses password as a parameter.

Current solution mixes those two which can be confusing in future.

@rafalkalinski what do you think?

rafalkalinski: Actually that was my initial thought on that :)
I'm going to change this.

PR #85, opened 2015-03-12
Notification service + dropwizard Jedis bundle

kvas-damian:
**Beginnings of Notifications service**
* working connection to redis
* configuration for deployment to mesos


.gitignore
```diff
@@ -16,3 +16,4 @@ pandora.yaml
 pandora-example/example.yaml
 **/.gradle
 **/build
+service/notifications/src/main/resources/notifications.yaml
```
sqreek: Please put this to module .gitignore

armonr:
@kvas-damian should Jedis be moved to another repo or will this be done in another PR?

pchojnacki:
I think moving it to another repo is not needed. And I hope this PR will not linger too long so then Jedis would be available sooner for use and/or refactoring.

lib/jedis/build.gradle
```diff
@@ -0,0 +1,14 @@
+apply from: rootProject.file('gradle/pandora_dependencies.gradle')
+apply plugin: 'java'
```
pchojnacki: this is provided by subproject template from root pandora project.
so this line and sourceCompatibility and repositories definition is actually not needed.

sqreek:
After offline meeting with damian:
As far as I understand, this is also in PR for Jedis repository.
I suggest to remove this PR, create Fork from Jedis with those changes, and add it to Pandora dependency.
When/If PR in Jedis repository will merge, we will change only dependency.
What do you think about it?
@pchojnacki @kvas-damian @armonr

service/notifications/README.md
```diff
@@ -0,0 +1,46 @@
+# Pandora-example
```
pchojnacki: this is unnecessary

armonr:
@sqreek I thought that was the plan also after I had a meeting with @rafalkalinski about this. I agree forking the Jedis bundle to a separate repo makes pandora happier because it wont have to care about details of redis/jedis lib :)
Or we can go with this and once our custom jedis fork is ready, move over to it so that it wont block notification service's work. (The external bundle, as it stands, is not 100% stable or suitable for us, afaik, hence the need for it to be modified.)
What do you think, @kvas-damian?

pchojnacki:
Just looking at the code a little bit more makes me thing that using Jedis bundle is not needed. We could handle the use case for it pretty generically through Guice alone.

kvas-damian:
As most of you suggested we moved Jedis bundle code into separate repo:
https://github.com/Wikia/droptools

There is also PR to apply our changes in upstream:
https://github.com/benjamin-bader/droptools/pull/8

armonr:
:+1: Thanks @kvas-damian

service/notifications/src/main/java/com/wikia/notificationsservice/health/NotificationsHealthCheck.java
```diff
@@ -0,0 +1,11 @@
+package com.wikia.notificationsservice.health;
+
+import com.codahale.metrics.health.HealthCheck;
+
+public class NotificationsHealthCheck extends HealthCheck {
+
+  @Override
+  protected Result check() throws Exception {
+    return Result.healthy();
```
sqreek: Are you sure this is sufficient for health check?

service/notifications/src/main/java/com/wikia/notificationsservice/NotificationsServiceApplication.java
```diff
@@ -0,0 +1,74 @@
...
+      }
+    });
+  }
+
+  @Override
+  public void run(NotificationsConfiguration configuration, Environment environment) throws Exception {
+
+    //register healthCheck (mandatory)
+    final NotificationsHealthCheck healthCheck = new NotificationsHealthCheck();
+    environment.healthChecks().register("SimpleHealthCheck", healthCheck);
```
sqreek: Healthchecks can be registered by Dropwizard Guice. 
.enableAutoConfig(getClass().getPackage().getName()) should be add in initialize,
and HealhCheck should extends InjectableHealthCheck.

Also in this way Resources are registered into enviorement too.
[example from mobileconfig](https://github.com/Wikia/pandora/blob/master/service/mobile-config/src/main/java/com/wikia/mobileconfig/MobileConfigApplication.java#L63)

sqreek:
LATM

PR #91, opened 2015-03-17
Services 149 - Create gateway for Ceph

sqreek:
Create gateway for CephServer.

In our service we had duplicated code for connecting to Cephserver, this pull request extract it to gateway.

Also refactor ConfigurationServiceBase.
It was helper, not base class for ConfigurationServices.


sqreek:
ping @pchojnacki @drsnyder @nmonterroso

lib/facade-gateway/src/main/java/com/wikia/pandora/gateway/ceph/CephGateway.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public CephGateway(HttpClient httpClient, String cephUrlBucket) {
+    this.httpClient = httpClient;
+    this.cephUrlBucket = cephUrlBucket;
+  }
+
+  public String getText(String Path) throws IOException {
+    HttpGet httpGet = new HttpGet(cephUrlBucket + Path);
+    ResponseHandler<String> responseHandler = new BasicResponseHandler();
+    return this.httpClient.execute(httpGet, responseHandler);
```
drsnyder: Does this depend on the media type of the response object? That is, could you get something other than a `String`?

sqreek: Yes. Thats why its getText.
For now we use only this.

drsnyder:
Should the top level directory be `gateways`?

drsnyder:
Other than the name of the directory lgtm.

pchojnacki:
I don't have much comments about the code itself (it looks awesomely good). But I do wonder About the name/purpose.
To me it looks more like HttpFileGateway or at very least S3FileGateway. Because there is actually no Ceph specific code there. And with regards to S3 protocol there is only url mapping "$bucketname/filepath" and the rest is just http gets to specific urls.

lib/facade-gateway/src/main/java/com/wikia/pandora/gateway/ceph/CephGateway.java
drsnyder: Minor: `getObject` might make more sense depending on how much of a "ceph" interface we want this to be.

drsnyder: Come to think of it, the `bucket` might belong here rather than on the constructor.

drsnyder:
> I don't have much comments about the code itself (it looks awesomely good). But I do wonder About the name/purpose.

+1

> To me it looks more like HttpFileGateway or at very least S3FileGateway. Because there is actually no Ceph specific code there. And with regards to S3 protocol there is only url mapping "$bucketname/filepath" and the rest is just http gets to specific urls.

It might be nice to retain a name that reflects this is an object storage gateway. Either "ceph" or "objectstorage" would probably be fine.

sqreek:
About directory, I stick in previous convention that we keep it in separate folder, becuse if we need additional file for some gateway, then its all stick togather

about bucket, I prefer using different gateway for different bucket, but it can be changed... if its needs to.

PR #92, opened 2015-03-17
don't use an ip address as a port

nmonterroso:
/cc @jcellary @sqreek 

This is why metrics aren't making it into InfluxDB

(cherry picked from commit 74bd0f89c87e41563f16553d8503fc426fcf600d)


sqreek:
LGTM.

jcellary:
Yeah, using ip address as port is usually a bad idea :D 

Thanks for the fix

PR #93, opened 2015-03-17
Add Consul support

armonr:
RFCR @jogura 

* Adds Consul support
* Some refactoring of Jedis
* caveat: Jedis host/port not read from yaml


service/discussion/src/main/java/com/wikia/discussionservice/DiscussionServiceApplication.java
```diff
@@ -32,6 +39,20 @@ public void initialize(Bootstrap<DiscussionServiceConfiguration> bootstrap) {
 
     // Swagger bundle for the SwaggerUI
     bootstrap.addBundle(new SwaggerBundle<DiscussionServiceConfiguration>());
+
+    // View bundle for the document views
+    bootstrap.addBundle(new ViewBundle());
+
+    // AssetBundle for the Hal Browser
+    bootstrap.addBundle(new AssetsBundle());
+
```
jogura: Please remove lines 42-48 this is for the HALBrowser which we aren't going with. Plus there is a bug with `dropwizard-swagger-0.5.3` that prevents adding another AssetBundle. Apparently this is fixed with version `0.5.4` but that hasn't been pushed out as of yet. If we do need to have another AssetBundle then we may need to push to the our internal archiva.

service/discussion/src/main/java/com/wikia/discussionservice/configuration/JedisFactory.java
```diff
@@ -0,0 +1,7 @@
+package com.wikia.discussionservice.configuration;
+
+import com.wikia.discussionservice.services.JedisService;
+
+public interface JedisFactory {
+  JedisService createJedisService(String host, int port);
+}
```
jogura: Where is this interface used? I don't see any classes that implement this or references it.

service/discussion/build.gradle
```diff
@@ -48,3 +48,21 @@ task testAll(type: Test){
...
+        environment {
+            CONSUL_KEY_VALUE_CONFIG_SOURCE = "http://localhost:8500/config/discussion/testing"
+        }
+    }
+    dev {
+        environment {
+            CONSUL_KEY_VALUE_CONFIG_SOURCE = "http://consul.service.sjc.consul:8500/config/discussion/dev"
+        }
+    }
+}
```
jogura: We should remove the URLs to the consul and opt to follow the pattern of using `${testConsulHost}` variables.

service/discussion/src/main/java/com/wikia/discussionservice/DiscussionServiceApplication.java
armonr: Good to know. I don't think we have an immediate need for any `AssetBundle`.

service/discussion/src/main/java/com/wikia/discussionservice/DiscussionServiceModule.java
```diff
@@ -4,25 +4,44 @@
...
 import com.theoryinpractise.halbuilder.standard.StandardRepresentationFactory;
+import com.wikia.discussionservice.configuration.DiscussionServiceConfiguration;
 import com.wikia.discussionservice.mappers.*;
 import com.wikia.discussionservice.services.ForumService;
 import com.wikia.discussionservice.services.PostService;
 import com.wikia.discussionservice.services.ThreadService;
 import com.wikia.discussionservice.services.JedisService;
 import com.wikia.discussionservice.dao.ContentDAO;
-
-import javax.inject.Named;
```
jogura: Why did we switch from the `javax.inject.Named` to the `com.google.inject.name.Named`? From my understanding Guice is fully compatible with the [JSR-330](https://github.com/google/guice/wiki/JSR330) minus some additional options that we aren't using ATM.

service/discussion/src/main/java/com/wikia/discussionservice/configuration/JedisFactory.java
armonr: Good catch, it should be removed. I have been trying out several different ways in Guice to inject the jedis params on object creation to no avail. This interface was part of one of those experiments.

service/discussion/src/main/java/com/wikia/discussionservice/DiscussionServiceModule.java
armonr: ah, no reason, just a result of removing and adding things

service/discussion/build.gradle
```diff
@@ -2,6 +2,11 @@ apply from: rootProject.file('gradle/dropwizard_service.gradle')
 apply plugin: 'application'
 version = "0.1"
 
+def consulDiscussionSourceUrl = "http://consul.service.sjc.consul:8500/config/${project.name}/"
```
jogura: I believe this should be set to `http://localhost:8500/config/{$project.name}` and in our `~/.gradle/gradle.properties` we use the PROD URL.

service/discussion/src/main/java/com/wikia/discussionservice/guicemodule/DiscussionModuleFactory.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class DiscussionModuleFactory {
+
+  public static final String DISCUSSION_SERVICE_ENVIRONMENT = "DISCUSSION_SERVICE_ENVIRONMENT";
+  public static final String STAGE_PROD = "prod";
+  public static final String STAGE_QA = "qa";
+  public static final String STAGE_TESTING = "testing";
+  private static final Logger LOGGER = LoggerFactory.getLogger(DiscussionModuleFactory.class);
+
+  public static DiscussionModuleBase CreateDiscussionModule() {
```
jogura: Coding standards dictate that the static method be camelcase

service/discussion/src/main/java/com/wikia/discussionservice/services/JedisService.java
```diff
@@ -6,15 +6,42 @@
...
+  ) {
+    this.host = host;
+    this.port = port;
+  }
+
+  @NotNull
+  @Valid
+  private String host;
+
+  @NotNull
```
jogura: Unnecessary annotation for an `int` value

service/discussion/src/main/resources/discussion.yaml
```diff
@@ -23,8 +23,11 @@ server:
...
   - type: http
     port: 8081
-#consulConfig:
-#  tags:
-#    - auto_testing
-#  heartbeatInterval: 2s
-#  heartbeatTTL: 5s
+consulConfiguration:
+  tags: ${consul:CONSUL_TAGS:[testing]}
+  consulUri: ${consul:CONSUL_URI:http://consul.service.sjc.consul:8500}
```
jogura: I missed this one but this should also be `localhost:8500` as well.

jogura:
minus the additional comments LGTM.

PR #94, opened 2015-03-18
Fixed perf metric port setting

jcellary:
Applied Nelson's fix of performance metrics port setting to other config files (original PR - https://github.com/Wikia/pandora/pull/92)

@nmonterroso @sqreek


sqreek:
LGTM

PR #95, opened 2015-03-18
SOC-385 | Add jersey bean validation at default to pandora projects

kvas-damian:
This would allow to use input parameters validation through @annotations

ping @sqreek @rafalkalinski


sqreek:
Lgtm

PR #100, opened 2015-03-19
Community Data Service

nmonterroso:
@drsnyder @sqreek @pchojnacki 
https://wikia-inc.atlassian.net/browse/SERVICES-139

Provides a service to get data about a community. Only resource is `/{domain}` ex: `/leagueoflegends.wikia.com`. Spoke with @kenkouot and this is their preferred method for fetching information for a specific wikia. Also piggybacking in some changes that were required to make this work:

* [jooq](jooq.org) code generation task
* udpating to latest swagger-codegen version which allows accessing the Nirvana API asynchronously, though this didn't actually end up being used


lib/jooq/jooq-generation/src/main/groovy/com/wikia/jooq/JooqGenerationTask.groovy
```diff
@@ -0,0 +1,60 @@
...
+    String dbHost = null
+    String username = null
+    String password = null
+    String dbName = null
+    String tables = ".*"
+
+    @TaskAction
+    def generateJooq() {
+        if (new File(getJooqSrcDir()).exists()) {
+            throw new StopExecutionException()
```
sqreek: I had lately problem, that StopExecutionException stop task that thrown exception, but tasks with dependency on it was starting. 
Are you tested unhappy branch here? with dependency on this task? Is it works as expected?

service/community-data/src/main/java/com/wikia/communitydata/core/CommunityDataProvider.java
```diff
@@ -0,0 +1,126 @@
...
+                new ImmutableMap.Builder<String, String>()
+                    .put("domain", domain)
+                    .build()
+            ),
+            "error fetching nirvana data for community",
+            e
+        );
+      }
+
+      communityData = new CommunityData(
```
sqreek: I suggest to use Builder pattern.
Check Builder Generator plugin in IntelliJ

service/community-data/src/main/java/com/wikia/communitydata/exception/CommunityDataException.java
```diff
@@ -0,0 +1,60 @@
...
+  private CommunityDataException(Response response, String domain, Exception exception) {
+    super(response);
+    this.domain = domain;
+    this.exception = exception;
+
+    log();
+  }
+
+  private void log() {
+    LoggerFactory.getLogger(this.getClass()).error(
```
sqreek: I'm not sure is it a good practise to selflog exception in exception.
My insight say no.

service/community-data/src/main/java/com/wikia/communitydata/exception/CommunityDataException.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static net.logstash.logback.marker.Markers.appendEntries;
+
+import com.google.common.collect.ImmutableMap;
+import org.slf4j.LoggerFactory;
+
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Response;
+
+public class CommunityDataException extends WebApplicationException {
```
sqreek: Also this whole class seems overengineering like for exception

service/community-data/src/main/resources/banner.txt
```diff
@@ -0,0 +1,8 @@
+ ________________ 
+< community-data >
+ ---------------- 
+        \   ^__^
+         \  (oo)\_______
+            (__)\       )\/\
+                ||----w |
+                ||     ||
```
sqreek: AWESOME

service/community-data/src/test/java/com/wikia/communitydata/resources/CommunityDataResourceTest.java
```diff
@@ -0,0 +1,80 @@
...
+        null
+    );
+    when(communityDataProvider.getCommunityData(any(DSLContext.class), anyString())).thenReturn(data);
+
+    String response = resource.client()
+        .target("/wowwiki.com")
+        .request()
+        .get(String.class);
+    assertNotNull(response);
+    assertTrue(response.length() > 0);
```
sqreek: IMO it can be more asserts here, to check is response have correct structure and no data is missing in process.

service/community-data/src/main/java/com/wikia/communitydata/CommunityDataApplication.java
```diff
@@ -0,0 +1,69 @@
...
+  private void configureCors(Environment environment) {
+    FilterRegistration.Dynamic
+        filter = environment.servlets().addFilter("CORS", CrossOriginFilter.class);
+    filter.addMappingForUrlPatterns(EnumSet.allOf(DispatcherType.class), true, "/*");
+    filter.setInitParameter(CrossOriginFilter.ALLOWED_METHODS_PARAM, "GET,PUT,POST,DELETE,OPTIONS");
+    filter.setInitParameter(CrossOriginFilter.ALLOWED_ORIGINS_PARAM, "*");
+    filter.setInitParameter(CrossOriginFilter.ACCESS_CONTROL_ALLOW_ORIGIN_HEADER, "*");
+    filter.setInitParameter("allowedHeaders", "Content-Type,Authorization,X-Requested-With,Content-Length,Accept,Origin");
+    filter.setInitParameter("allowCredentials", "true");
+  }
```
jogura: I've found that this is only necessary if you're using Swagger's `Try it out` functionality from your local dev environment (localhost). This shouldn't be required if you're running from your devbox or the server that is going to eventually host the service.

service/community-data/src/main/java/com/wikia/communitydata/core/CommunityData.java
```diff
@@ -0,0 +1,88 @@
...
+
+import com.wikia.nirvana.converters.Namespace;
+import com.wikia.nirvana.generated.model.NavigationItem;
+import com.wikia.nirvana.generated.model.WikiLanguageData;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+public class CommunityData {
```
jogura: In the [discussion service](https://github.com/Wikia/pandora/pull/101/files#diff-c7a4d755496dad82455cc32b7f82ef4aR10) we are using [lombok](http://projectlombok.org) to reduce the boilerplate code. See if that is something you'd find useful. For the constructor you can use the [@AllArgsConstructor](http://projectlombok.org/features/Constructor.html). Just be sure to turn on annotation processing in Intellij.

lib/jooq/jooq-generation/src/main/groovy/com/wikia/jooq/JooqGenerationTask.groovy
nmonterroso: I'm unsure of what problem you're describing, but my main test here was "if the generated code already exists, does it skip this part of the build and continue with the overall target? If it doesn't exist, does it build it as expected and then continue on?", both of which passed. Can you explain a bit more so I can see whether I have the same problem you were seeing?

service/community-data/src/main/java/com/wikia/communitydata/CommunityDataApplication.java
nmonterroso: Yah, I realize it works this way, but I don't think this causes any harm. In general, I'm a fan of [allowing cross origin resource sharing](http://enable-cors.org/).

service/community-data/src/main/java/com/wikia/communitydata/exception/CommunityDataException.java
nmonterroso: My initial thought was that this would serve as a general exception the service could throw during a resource request. But then there was only one resource so it doesn't really make sense to separate out :)

I'll make changes to do the logging/exception thrown inline.

service/community-data/src/main/resources/banner.txt
nmonterroso: We should make a [cowsay](http://en.wikipedia.org/wiki/Cowsay) service

service/community-data/src/main/java/com/wikia/communitydata/core/CommunityDataProvider.java
nmonterroso: Good idea. Added in 24c3be9.

service/community-data/src/main/java/com/wikia/communitydata/exception/CommunityDataException.java
nmonterroso: Done in e6b0cb5

service/community-data/src/main/java/com/wikia/communitydata/exception/CommunityDataException.java
nmonterroso: You're probably right.. I removed these two classes altogether in e6b0cb5

lib/jooq/jooq-generation/src/main/groovy/com/wikia/jooq/JooqGenerationTask.groovy
sqreek: behaviour you described its good for StopExecutionException
Thanks for clarification,

In my case I thought its stop next tasks from execute, but it doesn't. So its good for this case

service/community-data/src/test/java/com/wikia/communitydata/resources/CommunityDataResourceTest.java
nmonterroso: 651bb88

nmonterroso:
Cool. Code looks much better after this round of PR. Thanks @sqreek and @jogura

service/community-data/src/main/java/com/wikia/communitydata/core/CommunityDataProvider.java
```diff
@@ -0,0 +1,126 @@
...
+      CityListRecord cityListRecord = result.into(CITY_LIST);
+      CityVerticalsRecord cityVerticalsRecord = result.into(CITY_VERTICALS);
+
+      NavigationItem navigation = null;
+      List<Namespace> namespaces = null;
+      WikiLanguageData languageData = null;
+
+      try {
+        MercuryApi mercuryApi = NirvanaHelper.getApi(domain, MercuryApi.class);
+        WikiDataContainer wikiDataContainer = mercuryApi.getWikiData();
```
drsnyder: Should this return [`Optional`](https://github.com/Wikia/guidelines/tree/master/Java#do-not-return-null-or-accept-null-as-an-argument)?

service/community-data/src/main/java/com/wikia/communitydata/core/CommunityDataProvider.java
```diff
@@ -0,0 +1,126 @@
...
+          .isAdult(cityListRecord.getCityAdult() == 1)
+          .articlePath(parseArticlePath(result.into(CITY_VARIABLES)))
+          .navigation(navigation)
+          .namespaces(namespaces)
+          .languageData(languageData)
+          .build();
+    }
+
+    return communityData;
+  }
```
drsnyder: This function is pretty long. Should it be split out into data fetch and object creation?

service/community-data/src/main/java/com/wikia/communitydata/resources/CommunityDataResource.java
```diff
@@ -0,0 +1,91 @@
...
+          "error fetching community data",
+          e
+      );
+      Response response = Response
+          .status(Response.Status.INTERNAL_SERVER_ERROR)
+          .entity(new ErrorResponse(String.format("error fetching data for %s", domain),
+                                    Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()))
+          .build();
+      throw new WebApplicationException(response);
+    }
```
drsnyder: I think it would be nice to move the exception handling logic elsewhere so it doesn't clutter the controller. There is a mechanism in [jersey for doing this](https://jersey.java.net/documentation/latest/representations.html#d0e6559).

service/community-data/src/main/java/com/wikia/communitydata/resources/CommunityDataResource.java
```diff
@@ -0,0 +1,91 @@
...
+      Response response = Response
+          .status(Response.Status.NOT_FOUND)
+          .entity(new ErrorResponse(String.format("community %s does not exist", domain),
+                                    Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()))
+          .build();
+      throw new NotFoundException(response);
+    }
+
+    return data;
+  }
```
drsnyder: Same here. I'll take a stab at adding something like [this](http://stackoverflow.com/questions/23628897/dropwizard-exception-handling-giving-custom-error-json-error-messages-to-clie) to core.

PR #106, opened 2015-03-20
SOC-387 | Notifications following

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-387


service/notifications/src/main/java/com/wikia/notificationsservice/domain/FollowableItem.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  public String toRedisKey() {
+    // TODO escaping
+    return StringUtils.join(new String[]{sourceType, sourceId, targetType, targetId}, KEY_DELIMETER);
+  }
+
+  public static FollowableItem fromRedisKey(String redisKey) {
+    String[] keyParts = redisKey.split(KEY_DELIMETER);
+    if (keyParts.length != NUMBER_OF_PROPERTIES) {
```
rafalkalinski: It is probably more than enough for current prototype, but I would put some \@todo here with reminder for preparing something  better for checking that everything is okay. My intuitive approach would be a regex mach, but I'm a perl guy :)

service/notifications/src/main/java/com/wikia/notificationsservice/resources/FollowedResource.java
```diff
@@ -0,0 +1,91 @@
...
+  @Path("/{userId: \\d+}")
+  @DELETE
+  @Timed
+  public Response delete(@PathParam("sourceType") String sourceType,
+                               @PathParam("sourceId") String sourceId,
+                               @PathParam("targetType") String targetType,
+                               @PathParam("targetId") String targetId,
+                               @PathParam("userId") Long userId,
+                               @Context Jedis jedis) {
+    FollowableItem item = new FollowableItem(sourceType, sourceId, targetType, targetId);
```
rafalkalinski: If I would request for that functionality, I would ask about possibility of providing either source or target to remove a subscription to some forum/article/other.

service/notifications/src/main/java/com/wikia/notificationsservice/resources/FollowedResource.java
```diff
@@ -0,0 +1,91 @@
...
+    FollowedModel followedModel = new FollowedModel(jedis);
+    FollowerModel followerModel = new FollowerModel(jedis);
+    followedModel.delete(item, userId);
+    followerModel.delete(userId, item);
```
rafalkalinski: If I understand it correctly - an user can follow many things. And if so, we should'n delete the follower user record along with the followed item. At least we should check if this used doesn't follow anything else.

service/notifications/src/main/java/com/wikia/notificationsservice/resources/FollowedResource.java
```diff
@@ -0,0 +1,91 @@
...
+  @Timed
+  public Response create(@PathParam("sourceType") String sourceType,
+                               @PathParam("sourceId") String sourceId,
+                               @PathParam("targetType") String targetType,
+                               @PathParam("targetId") String targetId,
+                               Long userId,
+                            @Context Jedis jedis) {
+    FollowableItem item = new FollowableItem(sourceType, sourceId, targetType, targetId);
+    FollowedModel followedModel = new FollowedModel(jedis);
+    FollowerModel followerModel = new FollowerModel(jedis);
```
rafalkalinski: We had two different classes here. Both of them separately have a good name, but together - it's a one letter difference. It will be confusing if we leave it like that.
Proposals: 
Follower - Chased
Follower - Lead
FollowerUser - FollowedItem
Fan - Followed
Addict - Followed  :)

service/notifications/src/test/java/com/wikia/notificationsservice/integration/FollowedIntegrationTest.java
```diff
@@ -0,0 +1,164 @@
...
+    Client client = ClientBuilder.newClient();
+    Integer numberOfItems = 6;
+    List<Long> userIds = new ArrayList<>();
+    Long userId;
+    Response response;
+    String requestUrl = String.format(followedResourceUrl, "Wikia", "COD", "thread", "thread-name" + new Date().getTime());
+
+    for (Integer i = 0; i < numberOfItems; i++) {
+      userId = new Date().getTime();
+      userIds.add(userId);
```
rafalkalinski: There is a low possibility of generating the same userId twice (or more). Is it okay for the test? I assume not, because we will have the repeating the same follower, so lower number of followers than items in userId as a result (and exception in line ```assertThat(responseItem.getFollowers().size()).isEqualTo(userIds.size());```

service/notifications/src/test/java/com/wikia/notificationsservice/integration/FollowedIntegrationTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+    Response response = client.target(requestUrl)
+        .request()
+        .post(Entity.entity(userId, MediaType.APPLICATION_JSON));
+    assertThat(response.getStatus()).isEqualTo(Response.Status.CREATED.getStatusCode());
+
+    response = client.target(requestUrl)
+        .request()
+        .post(Entity.entity(userId, MediaType.APPLICATION_JSON));
+    assertThat(response.getStatus()).isEqualTo(Response.Status.CREATED.getStatusCode());
```
rafalkalinski: Should service respond here with CREATED status?

service/notifications/src/test/java/com/wikia/notificationsservice/integration/FollowedIntegrationTest.java
```diff
@@ -0,0 +1,164 @@
...
+    Long userId = new Date().getTime();
+
+    String requestUrl = String.format(followedResourceUrl, "Wikia", "COD", "thread", "thread-name" + new Date().getTime());
+
+    Response response = client.target(requestUrl)
+        .request()
+        .post(Entity.entity(userId, MediaType.APPLICATION_JSON));
+    assertThat(response.getStatus()).isEqualTo(Response.Status.CREATED.getStatusCode());
+
+    String
```
rafalkalinski: Something happen with code format here.

service/notifications/src/test/java/com/wikia/notificationsservice/integration/FollowedIntegrationTest.java
```diff
@@ -0,0 +1,164 @@
...
+        String.format("%s/%s", requestUrl, userId);
+    response = client.target(deleteRequestUrl)
+        .request()
+        .delete();
+    assertThat(response.getStatus()).isEqualTo(Response.Status.NO_CONTENT.getStatusCode());
+
+    response = client.target(requestUrl)
+        .request()
+        .get();
+    assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
```
rafalkalinski: It's good that service sends here some regular HAL response. We should also consider expecting NOT_FOUND or NO_CONTENT status here.

service/notifications/src/main/java/com/wikia/notificationsservice/domain/FollowableItem.java
kvas-damian: I'm going to modify this part of code while solving to from ```toRedisKey``` method, but I'm not sure how regex secure us better than current solution?

service/notifications/src/test/java/com/wikia/notificationsservice/integration/FollowedIntegrationTest.java
kvas-damian: ```new Date().getTime()``` returns miliseconds, so I don't think our code and API call would be too fast.

service/notifications/src/test/java/com/wikia/notificationsservice/integration/FollowedIntegrationTest.java
kvas-damian: Good catch! Will take a look on it again.

service/notifications/src/test/java/com/wikia/notificationsservice/integration/FollowedIntegrationTest.java
kvas-damian: I was considering those statuses during implementation, but none of the sounded correct to me.
NOT_FOUND - suggest that user doesn't exist which is not true in this case - we just return empty set for this user
NO_CONTENT - means that there is no content to be return in HTTP body, which is also not true while we return HAL.

More discussion about it: http://stackoverflow.com/questions/13366730/proper-rest-response-for-empty-table

service/notifications/src/test/java/com/wikia/notificationsservice/integration/FollowedIntegrationTest.java
kvas-damian: This is correct format according to our editor settings.

rafalkalinski: I was referring to this code:
```
String
     deleteRequestUrl =
     String.format("%s/%s", requestUrl, userId);
````
It looks odd :)

File: service/notifications/src/main/java/com/wikia/notificationsservice/models/FollowerModel.java [hunk 9]
```diff
@@ -0,0 +1,37 @@
...
+public class FollowerModel extends AbstractJedisModel {
+
+  private static final String REDIS_KEY_PART = "followed";
+
+  public FollowerModel(Jedis jedis) {
+    super(jedis);
+  }
+
+  public List<FollowableItem> getAll(Long userId) {
+    //List<FollowableItem> followedItems = new ArrayList<FollowableItem>();
```
rafalkalinski: I assume this line is ready to be removed :)

service/notifications/src/test/java/com/wikia/notificationsservice/integration/FollowingIntegrationTest.java
```diff
@@ -0,0 +1,126 @@
...
+  public void verifyFollowersWhenFollowedAdded() {
+
+    Client client = ClientBuilder.newClient();
+    List<FollowableItem> items = new ArrayList<>();
+    FollowableItem item;
+    Long userId = new Date().getTime();
+    String requestUrl;
+
+    for (Integer i = 0; i < 7; i++) {
+      item = new FollowableItem("Wikia", "COD", "board", "63s" + new Date().getTime());
```
rafalkalinski: My proposal:
lets assign result of ```new Date().getTime()``` to some exampleVariable before loop, and in the loop just do:
exampleVariable + 1
instead of new Date object every iteration

PR #107, opened 2015-03-20
Fix not closing request

sqreek:
This fix close response in healthchecks and replace depracated HttpClient to CloseableHttpClient


pchojnacki:
LGTM

PR #110, opened 2015-03-20
Add http problem container class

drsnyder:
Add a container class for [http problem](https://tools.ietf.org/html/draft-nottingham-http-problem-07#page-3). This will allow us to provide predictable [error responses that comply with the guidelines](https://github.com/Wikia/guidelines/tree/master/APIDesign#error-handling).

To use the class you can do the following in your resource:

```java
      Problem response =
          Problem.builder()
              .status(Response.Status.NOT_FOUND.getStatusCode())
              .type(new URI("/user-not-found.html"))
              .title(String.format("No preferences found for user %d", userId))
              .build();

      return
          Response.status(Response.Status.NOT_FOUND)
              .entity(response)
              .type(Problem.MEDIA_TYPE)
              .build();
```

https://wikia-inc.atlassian.net/browse/SERVICES-17

/cc @nmonterroso @jogura


nmonterroso:
LGTM

gradle/pandora_dependencies.gradle
```diff
@@ -11,6 +11,8 @@ dependencies {
 
     compile 'com.wikia:dropwizard-logstash-encoder:1.0.2'
 
+    compile "org.projectlombok:lombok:1.12.6"
```
jogura: Why are we using this version of lombok? And not `1.16.2` from [Maven Central](http://search.maven.org/#search%7Cga%7C1%7Clombok)?

lib/service-core/src/main/java/com/wikia/pandora/http/Problem.java
```diff
@@ -0,0 +1,50 @@
+package com.wikia.pandora.http;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Value;
+import lombok.experimental.Builder;
```
jogura: This reference is deprecated the correct one to use is `lombok.Builder`

gradle/pandora_dependencies.gradle
drsnyder: I think it was a typo. I switched the last 2 numbers in the version :(. This probably explains the comment below as well. I'm guessing it probably didn't exist in this version.

drsnyder: Fixed in 70f5946.

PR #111, opened 2015-03-20
Fix the CONSUL_KEY_VALUE_CONFIG_SOURCE setting

drsnyder:
Fix the `CONSUL_KEY_VALUE_CONFIG_SOURCE` so that non integration environments can set the consul config source through the environment. If you attempt to set `CONSUL_KEY_VALUE_CONFIG_SOURCE` from the environment it is currently getting clobbered by the `environment` settings in `dropwizard_service.gradle`.

This change allows the developer to override that setting from the environment if they so choose. And if not, it will be set as expected.

If we continue with the setting as is working with gradle in dev will be counterintuitive. You'll will be required to configure consul with at child key of `integration`.

/cc @pchojnacki @nmonterroso @armonr


armonr:
Should all the `consulDiscussionSourceUrl` instances defined in services be removed as well? E.g.: https://github.com/Wikia/pandora/blob/fix-consul-source/service/mobile-config/build.gradle#L4-L28

armonr:
That brings up the question, should consul uri be taken directly from Env [here](https://github.com/Wikia/pandora/blob/fix-consul-source/lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/bundle/ConsulModule.java#L25-25) or should it get the env-specific uri defined in each service's gradle build?

drsnyder:
@armonr I want to fix the top level setting first. In practice I think we should standardize this setting  across environments for all services but you should be able to override it without hassle in dev.

I first want to agree on how to do this at the top level. Once we are there we can go and clean up the individual services.

armonr:
I see. As long as this does not break current mobile-config setup, LGTM.

drsnyder:
FWIW, they are all broken in my dev workflow right now-- or more specifically anyone who uses gradle directly to run/test.

PR #112, opened 2015-03-23
SERVICES-135 MobileApps clickstream service

pchojnacki:
This is clickstream service that is meant to stream event data posted in batches to Syslog.
Incoming batches have schema that is present in api subproject

Outgoing messages are 'debatched' and currently also duplicated. One copy is Sent serialized using simple string contactenation without any structure. The second copy reuses schema for JSON data and serialized the message using Apache Avro, then the message is Base64 encoded.

Both messages are sent through syslog.


service/clickstream/src/main/java/com/wikia/service/clickstream/ClickstreamModule.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.inject.Named;
+
+public class ClickstreamModule extends AbstractModule {
+
+  @Override
+  protected void configure() {
+    bind(ProvidesConsulConfiguration.class).to(ClickstreamConfiguration.class);
```
sqreek: Is this should be Configuration.class?

service/clickstream/src/main/java/com/wikia/service/clickstream/ClickstreamModule.java
```diff
@@ -0,0 +1,68 @@
...
+  @Provides
+  @Named("avro")
+  Optional<SyslogMessageSender> avroMessageSenderProvider(ClickstreamConfiguration config) {
+    if (config.getAvroOutput() != null) {
+      UdpSyslogMessageSender sender = new UdpSyslogMessageSender();
+      sender.setSyslogServerHostname(config.getAvroOutput().getHostname());
+      sender.setSyslogServerPort(config.getAvroOutput().getPort());
+      return Optional.of(sender);
+    } else {
+      return Optional.empty();
```
sqreek: I know it's no needed, but can you use Optional.<SyslogMessageSender>empty()?
Also its 2 returning points.

service/clickstream/src/main/java/com/wikia/service/clickstream/healthcheck/SendSyslogMessage.java
```diff
@@ -0,0 +1,34 @@
...
+  private Optional<StreamSender> simpleSender;
+
+  @Override
+  public String getName() {
+    return "syslog";
+  }
+
+  @Override
+  protected Result check() throws Exception {
+    if (avroSender.isPresent() || simpleSender.isPresent()) {
```
sqreek: In my opion, simple ping is better that checking is something was set on startup for simpleSender, or avroSender.
Check it is good, but on startup.

service/clickstream/src/main/java/com/wikia/service/clickstream/resources/MobileAppsResource.java
```diff
@@ -0,0 +1,65 @@
...
+    avroSender.ifPresent(
+        avroSender -> streams.v1()
+            .map(new MobileAppDataToAvroString())
+            .map(new StringToSyslogMessage("mobile_app_avro"))
+            .forEach(avroSender));
+    simpleSender.ifPresent(
+        simpleSender -> streams.v2()
+            .map(new MobileAppDataToSimpleString())
+            .map(new StringToSyslogMessage("mobile_app_simple"))
+            .forEach(simpleSender)
```
sqreek: duplicate lines, it should be refactored

service/clickstream/src/main/java/com/wikia/service/clickstream/ClickstreamModule.java
pchojnacki: Nope

service/clickstream/src/test/java/com/wikia/service/clickstream/resources/MobileAppResourceTest.java
```diff
@@ -0,0 +1,32 @@
+package com.wikia.service.clickstream.resources;
+
+import org.junit.Rule;
+
+import io.dropwizard.testing.junit.ResourceTestRule;
+
+public class MobileAppResourceTest {
+
+//  private final HelloWorldService helloWorldServiceMock = mock(
```
sqreek: commented code

service/clickstream/src/main/java/com/wikia/service/clickstream/resources/MobileAppsResource.java
pchojnacki: Not really,  the code is clear and readable. I don't think there will be third option sooner we will remove one of the senders than add another. If it will show up then we can think of refactoring. Refactoring will only serve to create more boiler plate here IMO.

service/clickstream/src/main/java/com/wikia/service/clickstream/ClickstreamModule.java
pchojnacki: Whats wrong with two return points ? Will adding helper variable really help the readability ?

sqreek: I dont know what happins with my previous comment.
I mean Optional.<SyslogMessageSender>empty()

service/clickstream/src/main/java/com/wikia/service/clickstream/healthcheck/SendSyslogMessage.java
pchojnacki: Well yeah but here this check and returned message helps provide the operator with information why newly started service might be failing.
I really have no better idea how to implement a healthcheck for service that just sends data through UDP

service/clickstream/src/main/java/com/wikia/service/clickstream/resources/MobileAppsResource.java
pchojnacki: Although I'll happily refactor the use of Seq

sqreek:
General question. Do you consider escaped character in this solution.
Are we realy sure that country (for example) cannot contain ``|`` symbol

service/clickstream/src/main/java/com/wikia/service/clickstream/healthcheck/SendSyslogMessage.java
sqreek: I think you cannot check it.
And for point to give information to operator, it should be on application startup

pchojnacki:
Good point. I'm not sure. its essentially user provided data (user being fastly) so who knows. Its legacy format used by analytics every character escape I implement will have to be implemented by them.
That is Why I added Avro serialization which is the correct way to handle these things.

service/clickstream/README.md
```diff
@@ -0,0 +1,40 @@
+# Pandora-example
```
sqreek: if service is almoust ready provide proper readme.md

sqreek:
Lgtm

PR #113, opened 2015-03-23
reduce amount of data returning from appsDeployer

sqreek:
Filter response to return only requested fields


sqreek:
ping @pchojnacki @nmonterroso @drsnyder

service/mobile-config/src/main/java/com/wikia/mobileconfig/resources/ApplicationsResource.java
```diff
@@ -42,7 +64,11 @@ public Representation getMobileApplicationList(
       @PathParam("platform") String platform
   ) throws java.io.IOException, MobileConfigException {
 
-    Applications bean = new Applications(this.appsList.getAppList(platform));
+    List<HashMap<String, Object>>
```
pchojnacki: Is there any reason to keep those two lines separate ? to me it just looks very sloppy.

pchojnacki:
Tests are not passing and there is one whitespace issue. Other than that it LGTM

service/mobile-config/src/main/java/com/wikia/mobileconfig/resources/ApplicationsResource.java
sqreek: Intellij break too long lines in this way. 
It's a little better to read this way.

service/mobile-config/src/main/java/com/wikia/mobileconfig/guicemodule/MobileConfigModuleBase.java
```diff
@@ -66,4 +69,19 @@ public CloseableHttpClient provideTestHtmlClient(MobileConfigConfiguration confi
...
+                                 "app_tag",
+                                 "hub",
+                                 "category_tag",
+                                 "name",
+                                 "android_release",
+                                 "android_image",
+                                 "ios_image",
+                                 "languages",
+                                 "vignetteimage_set");
+  }
```
drsnyder: Minor: Should this be an array in the config?

service/mobile-config/src/main/java/com/wikia/mobileconfig/resources/ApplicationsResource.java
```diff
@@ -21,10 +28,25 @@
...
+                                     "app_tag",
+                                     "hub",
+                                     "category_tag",
+                                     "name",
+                                     "android_release",
+                                     "android_image",
+                                     "ios_image",
+                                     "languages",
+                                     "vignetteimage_set"));
+  }
```
drsnyder: This might be another case for injecting this via the config. Do we need to have this list/set defined twice?

service/mobile-config/src/test/java/com/wikia/mobileconfig/resources/ApplicationsResourceTest.java
```diff
@@ -31,21 +31,24 @@
...
+        "[{ \"app_tag\": \"test data\" }]");
 
     for (int i = 0; i < 10; ++i) {
       String responseData = RESOURCES.client()
           .target("/applications/platform/test-platform")
           .request()
           .get(String.class);
-      assertThat(responseData).contains("test-key");
+      assertThat(responseData).contains("test data");
     }
```
drsnyder: What is this testing? The purpose of the loop is not clear to me.

drsnyder:
I didn't see any test for the new functionality being added.

service/mobile-config/src/test/java/com/wikia/mobileconfig/resources/ApplicationsResourceTest.java
sqreek: This test is testing cache.
I will change testName for less confusing

service/mobile-config/src/main/java/com/wikia/mobileconfig/guicemodule/MobileConfigModuleBase.java
sqreek: In guice way this is a config.
We can also put this into consul/config file.
Thoughts?
Main question is: is this have a business value to change it often? @gbenson-wikia

service/mobile-config/src/main/java/com/wikia/mobileconfig/resources/ApplicationsResource.java
sqreek: Actually, I put it as default (if none Injection/configuration is avaible).
Second thougth, it is not neccessery.

service/mobile-config/src/main/java/com/wikia/mobileconfig/guicemodule/MobileConfigModuleBase.java
pchojnacki: Hmm..
I don't think this is config material i.e. it wont be modified between deployments etc.
I'm somewhat ok that this is in Guice module.
And if we were to change this I think the best place to put it might be simple java properties file where such stuff is usually kept.

service/mobile-config/src/main/java/com/wikia/mobileconfig/resources/ApplicationsResource.java
pchojnacki: I know intellij breaks it like so but that is a fault of algorithm they use to break. Which IMO doesn't produce good results all the time.
And I would argue it this is less readable and I always fix such lines.
But if you really believe that this is more readable and are not just going with the default behavior. Then there is no point in arguing it here.

service/mobile-config/src/main/java/com/wikia/mobileconfig/guicemodule/MobileConfigModuleBase.java
drsnyder: FWIW I did something similar in the [config for the user preference service](https://github.com/Wikia/pandora/blob/master/service/user-preference/src/main/resources/user-preference.yaml#L36).

PR #114, opened 2015-03-23
Refactor to use standard `Problem` for error response.

jogura:
* Refactor ErrorResponseBuilder to use new Problem
* Refactor Resources to use modified ErrorResponseBuilder
* Delete original ErrorResponse class
* Add HalProblem wrapper to produce the correct Swagger response in the UI

I just saw @drsnyder comment regarding using Jersey's built-in exception mappers to handle responses. I can refactor this code again if needed based on the results of that work.

RFCR @jimmordino @armonr


service/discussion/src/main/java/com/wikia/discussionservice/resources/ForumResource.java
```diff
@@ -79,6 +78,7 @@ public Response getForums(
 
     Optional<ForumRoot> forumRoot = forumService.getForums(siteId.get());
 
+    System.out.println("testing5");
```
armonr: Debug statement?

armonr:
minor comment. LGTM

service/discussion/src/main/java/com/wikia/discussionservice/domain/hal/output/HalProblem.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  public String getDetail() {
+    return problem.getDetail();
+  }
+
+  public String getInstance() {
+    return problem.getInstance().toString();
+  }
+}
```
drsnyder: This is required for swagger?

jogura: Yes, in the class's javadoc I referred to the fact the [overriding models technique](`https://github.com/swagger-api/swagger-core/wiki/overriding-models`) doesn't seem to work with our version of `dropwizard-swagger` so in order to produce the correct documentation this was the work-around.

service/discussion/src/main/java/com/wikia/discussionservice/resources/ForumResource.java
```diff
@@ -177,9 +187,13 @@ public Response deleteForum(
...
 
-    return ErrorResponseBuilder.buildErrorResponse(10101, String.format(
-        "No forum found for site id: %s with forum id: %s", siteId.get(),
-        forumId.get()), null, Response.Status.NOT_FOUND);
+    return ErrorResponseBuilder.buildErrorResponse(
+        String.format("No forum found for site id: %s with forum id: %s", siteId.get(),
+                      forumId.get()),
+        null,
+        Response.Status.NOT_FOUND,
+        null,
```
cmiller-wikia: Methods where half the arguments are null half the time make me nervous. I'd much rather have a version of the method that explicitly didn't accept those arguments.

drsnyder:
:eyes: :+1: TM

PR #118, opened 2015-03-24
Add redis database parameter defaulting to db/1

armonr:
This is so that we don't write data to db/0 when we deploy and test

/CC @jogura


service/discussion/src/main/java/com/wikia/discussionservice/configuration/JedisConfiguration.java
```diff
@@ -34,11 +41,19 @@ public int getPort() {
...
   }
 
-  public void getPort(int port) {
+  public void setPort(int port) {
     this.port = port;
   }
+
+  public void setDatabase(int database) {
+    this.database = database;
+  }
```
jogura: Does it make sense to use @lombak in this file?

jogura:
If the answer to my question is no I'm good with it...LGTM

PR #119, opened 2015-03-24
Align DW version dependencies with lib:service-core

jogura:
Also remove the explicit guice reference which is defined in service-core.

/cc @drsnyder @armonr


armonr:
:shipit:

PR #120, opened 2015-03-24
Fix dropwizard-swagger to work with DW 0.8.0 (final)

jogura:
Update the internal version of dropwizard-swagger to include
fix for the 0.8.0 (final) version of DW. The problem goes away
if we use versions 0.8.0-rc2/rc5 or 0.8.1-snapshot.

/cc @drsnyder @armonr @nmonterroso


nmonterroso:
Since we seem to be standardizing on using Swagger annotations to document our resources, I wonder if we should break this config into `lib/swagger/build.gradle` so we can have a central location for this config, since the `discussion`, `community-data`, and `user-preference` services all use these libraries.

jogura:
My concern is if we do what you're suggesting (including the having a dependency on a specific version of DW) is we are potentially introducing problems that weren't there before on unsuspecting services. This particular problem only occurs with DW 0.8.0 (final). By me removing the explicit dw-freemarker-view version I inadvertently broke this service. I understand with better integration tests we may have caught this but the problem still remains.

armonr:
LGTM

PR #121, opened 2015-03-25
Soc 386 As an user I want to get all notifications for me

rafalkalinski:
Notification service supports:
- post method 
  - client can create a notification for an user
- get method 
  - client can read one notification for an user
  - client can read all notifications for an user - supports pagination (offset and limit parameters)

There are integration tests for for all the functionality.


gradle/dropwizard_service.gradle
```diff
@@ -19,6 +20,7 @@ run {
             MOBILE_CONFIG_SERVICE_ENVIRONMENT   : "prod",
             EXAMPLE_SERVICE_ENVIRONMENT   : "prod",
             DISCUSSION_SERVICE_ENVIRONMENT   : "prod",
+            NOTIFICATIONS_SERVICE_ENVIRONMENT   : "prod",
```
pchojnacki: I think this should work by redeclaring 
```
testing {
  environment {
    NOTIFICATIONS_SERVICE_ENVIRONMENT   : "prod",
  }
}
```
that should actually merge the configs in the end.

service/notifications/src/main/java/com/wikia/notificationsservice/domain/Problem.java
```diff
@@ -0,0 +1,20 @@
+package com.wikia.notificationsservice.domain;
+
+public class Problem {
```
pchojnacki: you can reuse this: https://github.com/Wikia/pandora/blob/28ce5d8cffbcee3d24ad69df5a4856fda980f6e8/service/discussion/src/main/java/com/wikia/discussionservice/domain/hal/output/HalProblem.java
I believe

service/notifications/src/main/java/com/wikia/notificationsservice/resources/helpers/InboxUriHelper.java
```diff
@@ -0,0 +1,49 @@
...
+  public URI getURI(Long offset, Long limit) {
+    return UriBuilder.fromResource(InboxResource.class).queryParam("offset", offset)
+        .queryParam("limit", limit).build(userId);
+  }
+
+  public URI getFirstURI(Long offset, Long limit) {
+    return getURI((long) 0, limit);
+  }
+
+  public URI getPrevURI(Long offset, Long limit, Long total) {
```
pchojnacki: Personal nitpick.
You are a fast typer and it will take upi 1-2 additional seconds  to type Previous.
and Prev some one could understand as Preview or Prevent. :dancers:

service/notifications/src/test/java/com/wikia/notificationsservice/integration/ApplicationsIntegrationTest.java
```diff
@@ -0,0 +1,322 @@
...
+          userId
+      ));
+    }
+
+    UriBuilder requestUrl;
+    Response response;
+
+    requestUrl = notificationInboxUrl.clone().path(String.valueOf(userId));
+
+    for (i = 0; i < notificationNumber; ++i) {
```
pchojnacki: Nitpick. I would use java Iteration for this like: for(Notification notif:notificationsToPost)

service/notifications/src/test/java/com/wikia/notificationsservice/integration/ApplicationsIntegrationTest.java
```diff
@@ -0,0 +1,322 @@
...
+  public void notificationGetWithPagination() {
+    Client client = ClientBuilder.newClient();
+
+    int notificationNumber = 21;
+    long userId = new Date().getTime();
+
+    List<Notification> notificationsToPost = new ArrayList<>();
+
+    int i, j;
+    for (i = 0; i < notificationNumber; ++i) {
```
pchojnacki: I see this 2nd time I think it should refactored to common helper function

service/notifications/src/test/java/com/wikia/notificationsservice/integration/ApplicationsIntegrationTest.java
```diff
@@ -0,0 +1,322 @@
...
+                .toString()
+        ).isEqualTo(
+            requestOldCurrentUrl.clone()
+                .replaceQuery("")
+                .path(notificationGetAllResponse.getNotification(i).getId().toString())
+                .toString()
+        );
+      }
+    }
+  }
```
pchojnacki: This test is pretty long and hard to follow. Could you refactor some of the code to helper methods and/or consider splitting this test into two ?

service/notifications/src/test/java/com/wikia/notificationsservice/integration/ApplicationsIntegrationTest.java
```diff
@@ -0,0 +1,322 @@
...
+  public void notificationGetWithOffsetAndLimit() {
+    Client client = ClientBuilder.newClient();
+
+    int notificationNumber = 15;
+    long userId = new Date().getTime();
+
+    List<Notification> notificationsToPost = new ArrayList<>();
+
+    int i;
+    for (i = 0; i < notificationNumber; ++i) {
```
pchojnacki: 3rd time :D

service/notifications/src/main/java/com/wikia/notificationsservice/domain/Notification.java
```diff
@@ -3,10 +3,18 @@
...
 public class Notification {
+
+  private Long id;
   private String text;
+  private String sourceType;
+  private String sourceId;
+  private Long userId;
 
-  public Notification(String text) {
+  public Notification(String text, String sourceType, String sourceId, Long userId) {
```
drsnyder: Our current pattern for domain objects is to use [lombok](http://projectlombok.org/) and the `@Value` (for immutable objects) and `@Builder` for the builder.

service/notifications/src/main/java/com/wikia/notificationsservice/domain/Problem.java
drsnyder: This is now in [core](https://github.com/Wikia/pandora/blob/master/lib/service-core/src/main/java/com/wikia/pandora/http/Problem.java).

service/notifications/src/main/java/com/wikia/notificationsservice/models/NotificationModel.java
```diff
@@ -0,0 +1,75 @@
...
+    return nextId;
+  }
+
+  public Notification get(Long notificationId) throws IOException {
+    String notificationBlob = getJedis().get(getKey(notificationId));
+
+    if (notificationBlob != null) {
+      return unserializeNotification(notificationBlob);
+    } else {
+      return null;
```
drsnyder: See the [guidelines about not accepting or returning null](https://github.com/Wikia/guidelines/tree/master/Java#do-not-return-null-or-accept-null-as-an-argument).

service/notifications/src/test/java/com/wikia/notificationsservice/integration/ApplicationsIntegrationTest.java
```diff
@@ -0,0 +1,322 @@
...
+    UriBuilder requestOldCurrentUrl = requestUrl.clone();
+    UriBuilder requestFirstUrl;
+    UriBuilder requestLastUrl;
+    UriBuilder requestPrevUrl;
+    UriBuilder requestNextUrl;
+    UriBuilder requestSelfUrl;
+
+    int loopMax = (int) Math.ceil(notificationNumber / notificationDefaultCount);
+
+    for (j = 0; j <= loopMax; ++j) {
```
garthwebb: I've read this a few times and I still don't know what ```i``` and ```j``` are counting; notifications and pages?  Even though this is just a test it would be nice to have meaningful names here.

service/notifications/src/main/java/com/wikia/notificationsservice/models/UserInboxModel.java
```diff
@@ -0,0 +1,80 @@
...
+
+import redis.clients.jedis.Jedis;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class UserInboxModel extends AbstractJedisModel {
```
drsnyder: These aren't really models but data access objects. I think a better place for these would be `com.wikia.notifications.dao` with names like `UserInbox[DAO]`.

service/notifications/src/main/java/com/wikia/notificationsservice/resources/InboxResource.java
```diff
@@ -0,0 +1,156 @@
...
+  public InboxResource(RepresentationFactory representationFactory) {
+    this.representationFactory = representationFactory;
+  }
+
+  @GET
+  @Timed
+  public Representation multiGet(@PathParam("userId") Long userId,
+                                 @QueryParam("offset") @DefaultValue("0") @Min(0) Long offset,
+                                 @QueryParam("limit") @DefaultValue("10") @Min(1) @Max(1000) Long limit,
+                                 @Context Jedis jedis) {
```
drsnyder: Personally, I'm not a fan of injecting data access objects via resource method params. There is a "one of these does not look like the other" moment when you see query and path params along side a data access param.

service/notifications/src/main/java/com/wikia/notificationsservice/domain/Notification.java
pchojnacki: Just to make sure. It is up to a service developers to choose to use lombok or when to i immutable objects and/or builders, right?

service/notifications/src/main/java/com/wikia/notificationsservice/resources/InboxResource.java
pchojnacki: There is a reason for though.
Afaik Instance of the resource class can be shared/reused for different requests.
adding @Context Jedis jedis automatically provides one instance of this 'DAO' from Jedis pool per request through this hk2 factory https://github.com/benjamin-bader/droptools/blob/master/dropwizard-redis/src/main/java/com/bendb/dropwizard/redis/jersey/JedisFactory.java

what I understand also is that when the context is finished then the instance is automatically closed / returned to connection pool.

This naturally provides per request resources and is I think quite common way of jersey to provide various Context objects to the request like e.g. Headers.

So while it might look that one of this things doesn't belong is a part of the design of Jersey.

service/notifications/src/main/java/com/wikia/notificationsservice/domain/Notification.java
drsnyder: Yeah, it's not a requirement, but this seems like a good candidate for immutability and lombok. I imagine most domain objects will be immutable.

service/notifications/src/main/java/com/wikia/notificationsservice/exceptionhandlers/WebApplicationExceptionHandler.java
```diff
@@ -0,0 +1,31 @@
...
+  private HttpHeaders headers;
+
+  @Override
+  public Response toResponse(WebApplicationException exception) {
+    return Response.status(Response.Status.INTERNAL_SERVER_ERROR)
+        .entity(
+            Problem.builder()
+                .title("Exception")
+                .detail(exception.getMessage())
+                .status(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode())
```
kvas-damian: @drsnyder it's strange that ```Problem``` requires status and in the same time it's not a ```Response``` type.
This way I need to provide status code two times.

service/notifications/src/main/java/com/wikia/notificationsservice/domain/Notification.java
kvas-damian: It's nice, thanks for tip.

service/notifications/src/main/java/com/wikia/notificationsservice/exceptionhandlers/WebApplicationExceptionHandler.java
drsnyder: Yes, it is duplicated. See the [RFC](http://www.ietf.org/archive/id/draft-nottingham-http-problem-07.txt) for more information about the rationale.

kvas-damian: Thanks for source.

service/notifications/src/test/java/com/wikia/notificationsservice/integration/ApplicationsIntegrationTest.java
rafalkalinski: The method was refactored. It is not perfect, but I have no idea how to split it more on smaller pieces and keeping readability on some decent level.

garthwebb: I think @pchojnacki has a fair point.  I think there are a number of chunks of code that could still be refactored.  For example L141-L153:
```java
    Response response;

    requestUrl = notificationInboxUrl.clone().path(
        notificationsToPost.get(0).getUserId().toString()
    );

    for (Notification notificationToPost : notificationsToPost) {
      response = client.target(requestUrl)
          .request()
          .post(Entity.entity(notificationToPost, MediaType.APPLICATION_JSON));

      assertThat(response.getStatus()).isEqualTo(Response.Status.CREATED.getStatusCode());
    }
```
This could be replaced with something like:
```java
assertCanCreateNotifications(notificationsToPost);
```
There are two other loops that could probably been converted like this as well.  I think this reads more clearly and allows the reader to get a summary of the action without having to understand all the details all at once.

service/notifications/src/test/java/com/wikia/notificationsservice/integration/ApplicationsIntegrationTest.java
```diff
@@ -0,0 +1,326 @@
...
+    List<Notification> notificationsToPost = notificationsToPostCreate(notificationNumber);
+
+    UriBuilder requestUrl;
+    Response response;
+
+    requestUrl = notificationInboxUrl.clone().path(
+        notificationsToPost.get(0).getUserId().toString()
+    );
+
+    for (Notification notificationToPost : notificationsToPost) {
```
garthwebb: The difference between ```notificationToPost``` and ```notifiactionsToPost``` is only one character out of a fairly long string.  Consider making the loop variable a bit different so they are easier to tell apart.  In this case, its probably ok to shorten the loop variable since it has a smaller scope.  Maybe just ```notification```?  The 'ToPost' action is clear enough by it being the loop variable of ```notifiactionsToPost```

service/notifications/src/test/java/com/wikia/notificationsservice/integration/ApplicationsIntegrationTest.java
rafalkalinski: I moved this part to separate method, but not assert.

service/notifications/src/test/java/com/wikia/notificationsservice/integration/ApplicationsIntegrationTest.java
rafalkalinski: changed in the new branch: SOC-387-after-CR

PR #125, opened 2015-03-30
dropwizard-guice update

nmonterroso:
@pchojnacki @jogura @drsnyder @sqreek @armonr 
This changes our dropwizard-guice integration from [com.hubspot.dropwizard:dropwizard-guice:0.8.0](https://github.com/HubSpot/dropwizard-guice) to [ru.vyarus:dropwizard-guicey:2.1.2](https://github.com/xvik/dropwizard-guicey). The main differences are that the injector is now created in the `run` phase instead of the `initialize` phase, so we can't use dependency injection at that point, but introduces the concept of [installers](https://github.com/xvik/dropwizard-guicey#installers).

I've migrated all the existing services to use the new integration, and ran manual tests where I'm familiar with the service output to verify everything still works as expected.


service/discussion/src/main/java/com/wikia/discussionservice/guicemodule/DiscussionModuleBase.java
```diff
@@ -25,79 +25,30 @@
...
     bind(ForumRepresentationMapper.class).to(HALForumRepresentationMapper.class);
     bind(ThreadRepresentationMapper.class).to(HALThreadRepresentationMapper.class);
     bind(PostRepresentationMapper.class).to(HALPostRepresentationMapper.class);
     bind(RepresentationFactory.class).to(StandardRepresentationFactory.class);
-    bind(ProvidesConsulConfiguration.class).to(DiscussionServiceConfiguration.class);
-  }
-
-  @Provides
-  @Named("forumService")
-  public ForumService provideForumService() {
```
nmonterroso: @jogura @armonr 
I don't think these `@Provides` are needed, so I removed them. From what I can see, `ForumService`, `PostService`, and `ThreadService` should be provided by the injector properly without these bindings. Let me know if I'm wrong.

lib/consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -72,4 +74,18 @@ protected String getConfigKey(String key) {
...
+
+  public static ConsulKeyValueConfig fromEnv() {
+    String consulUri = System.getenv(ENV_CONFIG_SRC);
+    if (consulUri != null) {
+      try {
+        URI uri = new URI(consulUri);
+        return new ConsulKeyValueConfig(uri.getHost(),
+                                        uri.getPort(),
+                                        uri.getPath().substring(1));
+      } catch (Exception e) {}
```
sqreek: Please at least add log.trace.
Maybe this can be handle without catching exception?

service/discussion/src/main/java/com/wikia/discussionservice/guicemodule/DiscussionModuleBase.java
armonr: :+1:

sqreek:
Except empty catch, rest looks good.

lib/consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
pchojnacki: I don't think you have the logger available at this point :P
Maybe just throwing ```new RuntimeException(e)``` would be sufficient ?

pchojnacki:
LGTM,
FWIW I can address the empty catch in my PR

lib/consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
nmonterroso: Added call to `warn` in a2d3ba4, but like @pchojnacki said, we won't see this log in Kibana since this happens before the configuration is parsed.

lib/consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
```diff
@@ -72,4 +74,21 @@ protected String getConfigKey(String key) {
...
+        return new ConsulKeyValueConfig(uri.getHost(),
+                                        uri.getPort(),
+                                        uri.getPath().substring(1));
+      } catch (Exception e) {
+        LoggerFactory.getLogger(ConsulKeyValueConfig.class)
+            .warn(String.format("unable to parse consulUri %s", consulUri), e);
+      }
+    }
+
+    return null;
```
drsnyder: Minor: Should we be using [Optional here](https://github.com/Wikia/guidelines/tree/master/Java#do-not-return-null-or-accept-null-as-an-argument)?

lib/dropwizard/base/build.gradle
```diff
@@ -2,7 +2,9 @@ apply from: rootProject.file('gradle/common_deps.gradle')
 def dropwizardVersion = "0.8.0"
 
 dependencies {
-    compile 'com.hubspot.dropwizard:dropwizard-guice:0.8.0'
+    compile('ru.vyarus:dropwizard-guicey:2.1.2') {
+        exclude group: 'io.dropwizard'
+    }
```
drsnyder: Is this a common [anti-]pattern? I'm curious what the tradeoffs are for exclusions. Does it mean that we might break the world if we update dropwizard to a version that is API incompatible with `dropwizard-guicey`?

lib/consul-config/src/main/java/com/wikia/gradle/ConsulKeyValueConfig.java
nmonterroso: changed to use optionals in a3ba410

PR #126, opened 2015-03-30
Update the user-preference config

drsnyder:
Update the user-preference service config to include: 
- the app name to the logger
- the metrics section

/cc @nmonterroso


nmonterroso:
TRADESIES (also LGTM)

PR #127, opened 2015-03-30
makes more sense for getOrDefault to return a non-optional

nmonterroso:
@drsnyder @cmiller-wikia 

@cmiller-wikia mentioned that it doesn't make much sense for `getOrDefault` to return an optional, since you're providing a default return value anyway. This change reflects that.


drsnyder:
Good point. LGTM

PR #128, opened 2015-03-30
Make the providesWikiCitiesDataSource a singleton

drsnyder:
Make the `providesWikiCitiesDataSource` a singleton. 

The `@LazySingleton was removed in #125 correctly. However there should only be one of these in the system.

/cc @nmonterroso


nmonterroso:
LGTM

PR #131, opened 2015-03-31
Fix tests of infobox service

pchojnacki:
tests were started failing after merging to master. This fixes that issue


sqreek:
LG but, tests are not passes:)

PR #132, opened 2015-03-31
SOC-387 | Notifications following

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-387


kvas-damian:
Was CRed in: https://github.com/Wikia/pandora/pull/106

PR #133, opened 2015-04-01
Services 156 gradle tool to create new service

sqreek:
Provide gradle task (createNewService) for create new service folder structure with application, configuration, healthcheck and one resource.

Java code here is a template.
In template you can use @templateLower@, @templatePascal@, @templateCamel@ to get proper case


sqreek:
ping @nmonterroso @drsnyder @pchojnacki

new-service-template/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+
+deployments {
+    production {
+        resources {
+            cpus = 1
+            mem = 400
+            instances = 2
+        }
+        environment {
+            EXAMPLE_SERVICE_ENVIRONMENT = prodStage
```
nmonterroso: Do we really need this to be included as part of the base service? AFAIK, only MCS uses this.

new-service-template/src/main/java/com/wikia/services/template/TemplateApplication.java
```diff
@@ -0,0 +1,62 @@
...
+    return "@templateLower@"; // no uppercase letters
+  }
+
+
+  @Override
+  public void initialize(Bootstrap<@templatePascal@Configuration> bootstrap) {
+
+    @templateCamel@Module = @templatePascal@ModuleFactory.create@templatePascal@Module();
+
+    guiceBundle = GuiceBundle.<@templatePascal@Configuration>builder()
```
nmonterroso: @drsnyder - i think you'll have to update this template with the GuiceBundle stuff you're currently working on.

new-service-template/src/main/java/com/wikia/services/template/health/TemplateHealthCheck.java
```diff
@@ -0,0 +1,16 @@
...
+public class @templatePascal@HealthCheck extends NamedHealthCheck {
+
+  @Override
+  protected Result check() throws Exception {
+    return Result.healthy();
+  }
+
+  @Override
+  public String getName() {
+    return "example";
```
nmonterroso: should this be `@templateLower@-simple`?

new-service-template/src/main/resources/template.yaml
```diff
@@ -0,0 +1,44 @@
...
+        appname: pandora-example
+metrics:
+  reporters:
+    - type: influxdb
+      host: ${consul:INFLUXDB_HOST:127.0.0.1}
+      port: ${consul:INFLUXDB_PORT:5551}
+      prefix: example
+      frequency: 5 seconds
+      includes:
+        - com\.wikia\.example\.resources\..*
```
nmonterroso: this should probably be changed to `com\.wikia\.service\.@templateLower@\..*`

new-service-template/src/main/resources/template.yaml
```diff
@@ -0,0 +1,44 @@
...
+        tags: rawTags
+        context: rawContext
+      customFields:
+        appname: pandora-example
+metrics:
+  reporters:
+    - type: influxdb
+      host: ${consul:INFLUXDB_HOST:127.0.0.1}
+      port: ${consul:INFLUXDB_PORT:5551}
+      prefix: example
```
nmonterroso: this should probably be the raw project name from user input with spaces replaced by -

new-service-template/src/main/resources/template.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - type: logstash-socket
+      host: ${consul:LOG_HOST:127.0.0.1}
+      port: ${consul:LOG_PORT:1410}
+      fieldNames:
+        message: rawMessage
+        level: rawLevel
+        tags: rawTags
+        context: rawContext
+      customFields:
+        appname: pandora-example
```
nmonterroso: this should probably be the raw project name from user input with spaces replaced by -

nmonterroso:
Do we still need `service/example`? We can probably `git rm -r` that directory now.

nmonterroso:
I really think we should remove a service's knowledge about its current environment (testing, prod, etc) outside of the base service. Only MCS uses this functionality, but about half of the services declare it in their `build.gradle`, probably because of copy-paste from `mobile-config` or `example`. 

Further (but can probably be a different PR), can we change the MCS environment variable from `MOBILE_CONFIG_SERVICE_ENVIRONMENT` to `SERVICE_ENVIRONMENT`? That way **if we absolutely need to** we can all use the same variable that should already be set for us, since this environment variable would be declared in `dropwizard_service.gradle`.

sqreek:
Thanks Nelson, I totally forgot about example-yaml and build.gradle.

About removing sercice/example, there is a valid points to keep it.
* After download our repo you can run example service, and its working without additional work. Service created by this task, doesn't have setup in consul, so it need's additional work. Example show how it's work (at least a little)
* In new service, u haven't template of domain, and how to get value to config from Consul (greetingsWord).

So I will leave example as it is, for now. 
Also, if we want add any nice optional future, it will be a good place to show it how it's work..

sqreek:
@idradm has some input about what template should have,

drsnyder:
:+1:

nmonterroso:
@sqreek good points about the example service - let's keep it

idradm:
All of this feedback, is my own opinion, therefore some of it is just a hunch, personal offence (if any) not deliberate.

I've started work on service before create script was done, without any knowledge on dropwizard or jersey, with little on guice. First thing I did after copying the example was to ask @pchojnacki, what is the purpose of all of those files included, without any hesitation we removed more than half of the files, and this small changes in configuration files to make it work locally.

Questions and remarks:
1. Why there are so many files included, even though they are not needed for the simplest service to run? I'd suggest having the simplest possible service created, and then documentation with additional stuff described.
2. Not sure if this is still applicable, however it would be very useful, if I could run service locally straightaway after creating it, without any work.
3. By simplest application, I mean really simple, without any configuration or params used.
4. I would also suggest changing example structure. Moving all service configuration, and main application files with health checks to one folder. Leaving resources and guice (as it is needed for including your own code) at top directory without folders. The rest is just not useful at the start.

sqreek:
For now this PR is adding new functionality, we will take your advice and create new PR.
Thanks for input Adam

PR #134, opened 2015-04-01
improve coverage for mobile config health package

sqreek:


pchojnacki:
What exactly does this test checks ?

sqreek:
This only increase coverage, nothing more. (100% on health with this)

pchojnacki:
If the test doesn't give any value then it shouldn't be included just for the sake of 100% coverage.

sqreek:
This test is checking healthcheck name, to prevent accidental change healthcheck name and make mess in our log system :wink:

sqreek:
ping @gbenson-wikia.
This test improves our coverage, and sets it to 100%, but does it make any sense?

xaroth8088:
I actually believe that it does make sense to test this explicitly.

Pretend for a moment that someone does an allegedly benign search-and-replace rename in the system, and in amongst the other 100's of touched files was an accidental change to this string.  While we'd like to imagine that the reviewer would read through each of the hundreds of changed files and understand exactly the impact of the rename, it's a lot more likely that they'll look at a handful of them, do some happy-path testing on whatever they were trying to rename, and then they'll 'LGTM' the PR.

In part then, the question of whether it's worth unit testing something so trivial then comes down to the following question: would you rather learn that your health checks have changed their identifiers _before_ or _after_ it hits production?

If you'd like to avoid the future hours-to-days of puzzling out where your logs went, and the hours-to-days of log reconciliation, and the potential for lost data, then yes - it's probably worth spending the extra 15 minutes now to write this test, silly as it may seem at the time of writing.

The time spent on writing this test also has more concrete measurable effects, too.

Pretend for a moment that - instead of automated testing - we wanted to have a human doing our regression testing for us.  Every time that we wanted a comprehensive regression test, that human would have a step in their plan to ensure that the health check naming hadn't changed, so that the logs don't get mangled due to an accidental string replace refactor.

Let's assume that they're reasonably fast at this, and it only takes them 30 seconds to verify this step.  The new test run time is close enough to 0, so to keep the math simpler let's say that we've saved 30 seconds of QA time every time the unit tests are run.

4 developers a day, running a full regression suite 20 times a day (because it's automatic in the system as you make changes) = 80 full regression tests per day.  80 runs * 30 seconds is 40 minutes of QA labor saved per day, at a fixed cost of 15 minutes of developer time.  Inside of the first week, you'll have made back your investment several times over.  Over a working year, that'll be something like 4 work weeks of saved QA labor costs for a 15 minute developer time investment.

Now, a rational QA regression plan might opt to omit this test due to it being so trivial, ostensibly making the whole line of reasoning invalid.  However, to understand better, we'll need to think for a moment about *why* they would omit it, and that's pretty straightforward.  A QA regression plan might omit this test because something so small isn't worth a recurring 4 work weeks per year of a QA engineer's time.  But, if you could pay a one-time cost of 15 minutes to never have to test it explicitly again?  Now it seems cost-effective to put that work in.

So, yes, I believe writing the test makes sense.

PR #137, opened 2015-04-01
Code coverage community data

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/SERVICES-219
Improving quality of tests and increasing code coverage for the community-data service


lib/jooq/jooq-base/src/main/java/com/wikia/test/jooq/MockDataProviderBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+          record.setValue(field, valueList.get(i));
+        }
+
+        result.add(record);
+      }
+
+      return new MockResult[] {new MockResult(valueListSize, result)};
+    };
+  }
+}
```
drsnyder: It would be nice to split this function up. It has three loops wrapped in an anonymous function that's being returned. It's pretty hard to follow what is happening.

lib/service-core/src/main/java/com/wikia/pandora/http/StandardResponses.java
```diff
@@ -4,38 +4,27 @@
...
-                    .status(Response.Status.CONFLICT.getStatusCode())
-                    .title(title)
-                    .build())
+    return Response.status(status)
+        .entity(problem)
         .type(Problem.MEDIA_TYPE)
         .build();
   }
-
 }
```
drsnyder: :+1:

drsnyder:
Minor comments. OTW, LGTM

lib/jooq/jooq-base/src/main/java/com/wikia/test/jooq/MockDataProviderBuilder.java
nmonterroso: 032bb61

lib/nirvana-api/src/main/java/com/wikia/nirvana/NirvanaHelper.java
```diff
@@ -4,13 +4,13 @@
 import com.wikia.nirvana.generated.SwaggerApi;
 
 public class NirvanaHelper {
-  public static <T extends SwaggerApi> T getApi(String domain, Class<T> cls)
+  public <T extends SwaggerApi> T getApi(String domain, Class<T> cls)
```
sqreek: why not static? it looks like it should be helper.

nmonterroso: was originally `static`, but Mockito can't mock `static` functions, which is useful here for testing purposes.

sqreek: PowerMock can do this
https://code.google.com/p/powermock/wiki/MockitoUsage

sqreek: Also if you want go non static way -> change NirvanaHelper for something else (Helper give me assumption that all public method inside are static), or
If you go back to static and use Powermockito, add protected constructor to prevent instance creation of this class

nmonterroso: I don't think it's worth it to introduce a new library for this. I've changed the class name to NirvanaApiProvider in 255ef41

PR #141, opened 2015-04-02
make sure we can deserialize response string to HAL

nmonterroso:
@drsnyder @sqreek @pchojnacki 
https://wikia-inc.atlassian.net/browse/SERVICES-240

make sure we can deserialize response from MCS as valid HAL in addition to a `200` response.


sqreek:
LGTM, but I think we need to add more tests for unhappy scenario in ApplicationsIntegrationTest.

PR #143, opened 2015-04-03
Provide an ExceptionMapping bundle for handling common exceptions

drsnyder:
This PR provides sane default exception to response mappers for top level `Throwable` exceptions and 404s. You can think of this as a [standards compliant](https://github.com/Wikia/guidelines/tree/master/APIDesign#error-handling) replacement for the default dropwizard handlers.

This will also make testing a little easier for developers. The response either from unit or integration tests will be predictable and can be deserialized back into a `Problem` object.

Unfortunately, we can't do this without first disabling the default mappers provided by dropwizard. This can be done by adding the following to the server config in the services YAML file:

```
server:
   registerDefaultExceptionMappers: false
```

Update:  I simplified this significantly. I occurred to me over the weekend that I was clobbering the exception handling mechanism (`WebApplicationException`) provided by jersey. I don't think we want to do that.

https://wikia-inc.atlassian.net/browse/SERVICES-17

/cc @nmonterroso @sqreek @pchojnacki


lib/service-core/src/main/java/com/wikia/pandora/http/StandardResponses.java
```diff
@@ -16,6 +16,34 @@ public static Response Conflict(String title) {
     return buildResponse(Response.Status.CONFLICT, title);
   }
 
+  public static Response NotFound(Problem problem) {
```
nmonterroso: Can we refactor these functions that take `Problem`s to use `buildResponse`? This would require overloading `buildResponse` to take a `Problem` instead of a `String`, something like:

```java
public static Response NotFound(Problem problem) {
  return buildResponse(Response.Status.NOT_FOUND, problem);
}

private static buildResponse(Response.Status status, String title) {
  return buildResponse(status, buildProblem(status, title));
}

private static buildResponse(Response.Status status, Problem problem) {
  return .....
}
```

This would remove a good amount of code duplication in this class.

nmonterroso:
I think we should add documentation on how to use this: what needs to be added to your configuration yaml, and how to add your own custom responders.

nmonterroso:
Some comments, otherwise LGTM

lib/service-core/src/main/java/com/wikia/pandora/http/StandardResponses.java
drsnyder: I reduced the duplication in 4fa69a7.

drsnyder:
Re: docs, I agree. There are several steps required to use this. I'm planning to add some docs once it's merged.

nmonterroso:
BOOM LGTM :ship:

nmonterroso:
LGTM

PR #144, opened 2015-04-07
SERVICES-148 Consul state registration fixes

pchojnacki:
- Unify Consul configuration.
- Make Consul Reregister itself when it is deregistered.
- Make failing heartbeats non fatal
- Cleanup build scripts to not clobber environment settings
- Make Consul deregister itself when the application shuts down.
- All services converted to use new deployment configuration


gradle/dropwizard_service.gradle
```diff
@@ -1,28 +1,26 @@
...
-            MOBILE_CONFIG_SERVICE_ENVIRONMENT   : "prod",
-            EXAMPLE_SERVICE_ENVIRONMENT   : "prod",
-            DISCUSSION_SERVICE_ENVIRONMENT   : "prod",
-            NOTIFICATIONS_SERVICE_ENVIRONMENT   : "prod",
+            CONSUL_HOSTNAME                  : System.getenv("CONSUL_HOSTNAME") ?: testConsulHost,
+            ENVIRONMENT                      : "integration",
+            MOBILE_CONFIG_SERVICE_ENVIRONMENT: "prod", //TODO: remove this
+            EXAMPLE_SERVICE_ENVIRONMENT      : "prod",//TODO: remove this
+            DISCUSSION_SERVICE_ENVIRONMENT   : "prod",//TODO: remove this
+            NOTIFICATIONS_SERVICE_ENVIRONMENT: "prod",//TODO: remove this
```
drsnyder: What is your plan for removing these?

lib/consul-config/src/main/java/com/wikia/gradle/DefaultConsulKeyValueConfig.java
```diff
@@ -30,10 +30,34 @@ public DefaultConsulKeyValueConfig(Consul consul, String app, String env) {
...
+  public static Optional<ConsulKeyValueConfig> fromEnv() {
+    String consulUri = System.getenv(ENV_CONFIG_SRC);
+    DefaultConsulKeyValueConfig config = null;
+
+    if (consulUri != null) {
+      try {
+        URI uri = new URI(consulUri);
+        config = new DefaultConsulKeyValueConfig(uri.getHost(),
+                                                 uri.getPort(),
+                                                 uri.getPath().substring(1));
```
drsnyder: Can both of these statements produce exceptions or just the first one? The log line seems to apply to the `URI` construction.

lib/consul-config/src/main/java/com/wikia/gradle/DefaultConsulKeyValueConfig.java
```diff
@@ -30,10 +30,34 @@ public DefaultConsulKeyValueConfig(Consul consul, String app, String env) {
...
+      } catch (Exception e) {
+        LoggerFactory.getLogger(DefaultConsulKeyValueConfig.class)
```
drsnyder: Should this be `LOG.warn`?

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/ConsulRegistrationInstaller.java
```diff
@@ -0,0 +1,47 @@
...
+public class ConsulRegistrationInstaller implements InstanceInstaller<ConsulManager>,
+                                        FeatureInstaller<ConsulManager>,
+                                        BindingInstaller {
+
+  private final static Logger LOG = LoggerFactory.getLogger(ConsulRegistrationInstaller.class);
+
+  @Override
+  public void install(Environment environment, ConsulManager instance) {
+    LOG.info("Disabling Java DNS caching");
+    Security.setProperty("networkaddress.cache.ttl", "0");
```
drsnyder: I read [here](http://stackoverflow.com/a/17219327/8243) that there is a race condition with this setting. Is there somewhere we can do this to ensure that it's performed before any other network operations.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/ConsulGuiceApplication.java
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Explicit run() method in case of Guice bundle is not necessary because most normal dependencies
+   * are provided through injection.
+   */
+  @Override
+  public void run(T configuration, Environment environment) throws Exception {
+
+  }
+}
```
drsnyder: :+1: @nmonterroso started to dabble with a helper for creating the `GuiceBundle` but this is a better way to go.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/util/ManagedBean.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.wikia.dropwizard.consul.registration.events.DeregistrationRequest;
+import com.wikia.dropwizard.consul.registration.events.ShutdownRequest;
+import com.wikia.dropwizard.consul.registration.guice.ConsulBus;
+
+import com.google.common.eventbus.EventBus;
+import com.google.inject.Inject;
+import io.dropwizard.lifecycle.Managed;
+
+public class ManagedBean implements Managed {
```
drsnyder: Please document what this does.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/util/HostingPortRegistrationListener.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.inject.Inject;
+import io.dropwizard.lifecycle.ServerLifecycleListener;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.ServerConnector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+
+public class HostingPortRegistrationListener implements ServerLifecycleListener {
```
drsnyder: Please document.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/actors/Deregistrator.java
```diff
@@ -0,0 +1,29 @@
...
+
+  private final Consular consular;
+  private final ServiceIdProvider idProvider;
+
+  @Inject
+  public Deregistrator(Consular consular, ServiceIdProvider idProvider,
+                       @ConsulBus EventBus eventBus) {
+    this.consular = consular;
+    this.idProvider = idProvider;
+    eventBus.register(this);
```
drsnyder: Please also document this class. Seeing a call to `eventBus.register(this)` is counterintuitive in a `Deregistrator` class.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/ConsulGuiceApplication.java
```diff
@@ -0,0 +1,111 @@
...
+    return Collections.<Module>emptyList();
+  }
+
+  /**
+   * Application module is the default module that configures the application if additional modules
+   * need to be specified then getAdditionalModules can be overridden
+   *
+   * @return Guice module describing the application
+   */
+  public abstract Module getApplicationModule();
```
nmonterroso: I don't know if it should be required for services to override this. For instance, `community-data` does not have a Guice module.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/ConsulGuiceApplication.java
nmonterroso: Well, technically it was part of the dropwizard-guice library I was thinking of using, but overall yeah, this seems much better since we can tailor it to our needs :+1:.

drsnyder:
@pchojnacki please provide more documentation about these changes. I'm finding it hard to review effectively without a conceptual reference or a state diagram for how this works. It's as though there is a service being created within the service :).

Sorry, I couldn't help myself :)

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/ConsulGuiceApplication.java
pchojnacki: Hmm I think you are right. When I created this class I didn't consider that there will be services without App specific modules. But with auto discovery I think its possible.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/ConsulRegistrationInstaller.java
pchojnacki: I wasn't aware of the race condition, so thanks for pointing it out!
FWIW I just checked the offending code (at least some version of it I was able to find online) https://github.com/zxlooong/cvm/blob/9ce122b6a3dc14281b8600747f67d39488567e9e/cdc/src/share/classes/sun/net/InetAddressCachePolicy.java

And the situation with this setting is such that it can become permanent when some code references any of the static fields in this class.

So I think we have 2 options here:
1. move this code to static { } block inside ConsulGuiceApplication, with (optionally) assert that will check if the setting was respected which will also cause the initialization and prevent further changes to this setting.
2. remove this from the code and entirely rely only on system wide settings

drsnyder: > I wasn't aware of the race condition, so thanks for pointing it out!

I wasn't either until one of the consul nodes in SJC stopped responding to requests and my laptop happened to get it's IP when it queried for the address.

Option one seems tenuous to me. How do we know that it will be set before any other networking event happens on the JVM? Option 2 seems like the only sure way to set this value.

pchojnacki: FWIW ConsulGuiceApplication will be instantiated first since it contains main method which is the root of all static initialization per http://stackoverflow.com/questions/3196016/is-static-code-always-executed-when-we-use-a-class-for-the-first-time

Anyway I'll remove this code from here, and I'll create separate pull request just with this one functionality

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/ConsulGuiceApplication.java
```diff
@@ -0,0 +1,117 @@
...
+   * need to be specified then getAdditionalModules can be overridden
+   *
+   * @return Guice module describing the application
+   */
+  public Module getApplicationModule() {
+    return binder -> {
+    };
+  }
+
+  ;
```
sqreek: unnessesery semicolon

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/guice/ConfigurationBootstrapProvider.java
```diff
@@ -0,0 +1,33 @@
...
+  private final SystemConfigurationProvider systemConfigurationProvider;
+
+  @Inject
+  public ConfigurationBootstrapProvider(
+      PropertiesConfigurationProvider propertiesConfigurationProvider,
+      SystemConfigurationProvider systemConfigurationProvider) {
+    this.propertiesConfigurationProvider = propertiesConfigurationProvider;
+    this.systemConfigurationProvider = systemConfigurationProvider;
+  }
+  //new SystemConfigurationProvider()
```
sqreek: commented code

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/guice/DefaultsFromPropertiesConfigurationProvider.java
```diff
@@ -0,0 +1,46 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+import javax.inject.Provider;
+
+public class DefaultsFromPropertiesConfigurationProvider
+    implements Provider<PropertiesConfigurationProvider> {
+
+  public static final String CONSUL_DEFAULTS_PROPERTIES = "/consul_defaults.properties";
```
sqreek: slash in name?

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/guice/DefaultsFromPropertiesConfigurationProvider.java
```diff
@@ -0,0 +1,46 @@
...
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(DefaultsFromPropertiesConfigurationProvider.class);
+
+  public DefaultsFromPropertiesConfigurationProvider() {
+  }
+
+  @Override
+  public PropertiesConfigurationProvider get() {
```
sqreek: This function log error, but can pass and return null.
Is this behaviour is right?
If yes, consider lower exception level to warn

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/actors/HeartbeatSender.java
```diff
@@ -0,0 +1,110 @@
...
+    consulProvider.get().ifPresent((consul) -> {
+      Optional<String> healthcheckErrors = collectHealthcheckErrors();
+      try {
+        if (!healthcheckErrors.isPresent()) { // no errors
+          passTtlCheck(consul);
+        } else {
+          failTtlCheck(consul, healthcheckErrors.get());
+        }
+      } catch (NotRegisteredException e) {
+        LOG.warn("Service not registered");
```
sqreek: IMO we want now NotRegisterException faster, than reload application, and choose debug level.

sqreek: Also there is a little inconsistence what is log or what isn't

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/actors/Registrator.java
```diff
@@ -0,0 +1,62 @@
...
+    this.eventBus.unregister(this);
+  }
+
+  private void doRegister() {
+    if (portNumber > 0) {
+      try {
+        serviceIdProvider.get()
+            .ifPresent(serviceId -> this.consular.registerInConsul(2, serviceId));
+      } catch (RuntimeException ex) {
+        LOG.error("Registration failure", ex.getCause().getMessage());
```
sqreek: I want stacktrace here if its error, not only cause.
If its only warn, downgrade Log.level

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/guice/ConsulProvider.java
```diff
@@ -0,0 +1,34 @@
...
+  @Configuration("consul.http_port")
+  private Integer port;
+
+  @Override
+  public Optional<Consul> get() {
+    Consul consul = null;
+    try {
+      consul = Consul.newClient(hostname, port);
+    } catch (ConsulException ex) {
+      LOG.error("Couldn't connect to {}:{} Reason: {}", hostname, port, ex.getCause().getMessage());
```
sqreek: Same as above. If its error, I want to have stacktrace, 
if not, this is should be a warn.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/guice/DefaultsFromPropertiesConfigurationProvider.java
pchojnacki: its a path

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/guice/DefaultsFromPropertiesConfigurationProvider.java
pchojnacki: You are right and Injection Provider<?>  get() method should never return null. There is ProvisionException suggested by the Guice docs that I'll use here.

sqreek:
Look AWESOME to me. Good Job :+1: 
one minor: ctrl+r com.google.intject -> javax.inject

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/actors/HeartbeatSender.java
pchojnacki: Please clarify why its inconsistent ? 
NotRegisteredException is a 'normal' state the application can go into and I don't want to pollute the logs with IMO unnecessary stacktraces whenever a expected state transition happens.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/actors/Registrator.java
pchojnacki: I'll change the log to log all stacktrace.
But while this error is 'fatal'  the application is expected to recover from it.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/actors/HeartbeatSender.java
sqreek: inconsistent: 
LOG.debug("",e) and 
LOG.debug("{}", e.getCause().getMessage(), e);
looks odd

If its normal state to get this exception, I will log it as Warn instead error

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/guice/ConsulProvider.java
pchojnacki: This is error. But the application is expected to recover.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/actors/Registrator.java
pchojnacki: Oh. no I actually wont change the logging amount. Since this error will show up constantly when you run the application without VPN or when you wont configure Consul host 
But as soon as you connect to VPN it will recover. So I think Error lvl is valid

pchojnacki:
yeah you right I planned to do that but forgot :D

PR #146, opened 2015-04-07
Add support for nested forums & paginated retrieval of a forum's threads

armonr:
RFCR @jogura @jimmordino 

https://wikia-inc.atlassian.net/browse/DIS-9

"Unread" status is not added here, but will be part of user data.


service/discussion/src/main/java/com/wikia/discussionservice/resources/ForumResource.java
```diff
@@ -63,28 +62,66 @@ public ForumResource(ForumService forumService, ForumRepresentationMapper forumM
...
+    Forum forumRoot;
+    try {
+      forumRoot = forumService.getRootForum(siteId.get());
+    } catch (Exception ex) {
+      // todo log error
+      return ErrorResponseBuilder.buildErrorResponse(
+          ex.getMessage(),
+          Response.Status.NOT_FOUND,
+          uriInfo.getRequestUri());
+    }
```
jogura: personally, I would stick with the original pattern of unwrapping the `Optional` at the resource level. Especially in light of #143. 

The code would be refactored to: 

```
    Optional<ForumRoot> optionalForumRoot = forumService.getForums(siteId.get());

    return optionalForumRoot
        .map(forumRoot -> Response
            .ok(forumMapper.buildRepresentation(siteId.get(), forumRoot, uriInfo, responseGroup, ForumService.THREAD_LIMIT_DEFAULT, 0)
            .build())
        .orElseThrow(() -> new NotFoundException(
            String.format("No forums for site id: %s", siteId.get())));
```

service/discussion/src/main/java/com/wikia/discussionservice/dao/ForumDAO.java
```diff
@@ -24,26 +25,47 @@ public ForumDAO(ContentDAO contentDAO) {
...
+    if (forum != null) {
+      if (!forum.isLeaf()) {
+        List<Forum> children = new ArrayList<>();
+        for (Forum child : forum.getChildren()) {
+          children.add(contentDAO.getContent(siteId, child.getId(), Forum.class));
+        }
+        forum.setChildren(children);
+      }
+
+      List<ForumThread> threads = new ArrayList<>();
```
sqreek: it seems threads is not used.

service/discussion/src/main/java/com/wikia/discussionservice/dao/ForumDAO.java
```diff
@@ -24,26 +25,47 @@ public ForumDAO(ContentDAO contentDAO) {
...
+  public Optional<Forum> getForum(int siteId, long forumId, int limit, int offset) {
+    Forum forum = contentDAO.getContent(siteId, forumId, Forum.class);
+
+    if (forum != null) {
+      if (!forum.isLeaf()) {
+        List<Forum> children = new ArrayList<>();
+        for (Forum child : forum.getChildren()) {
+          children.add(contentDAO.getContent(siteId, child.getId(), Forum.class));
+        }
+        forum.setChildren(children);
```
sqreek: above can be replaced by 
forum.setChildren(forum.getChildren().stream().map(contentDAO.getContent(siteId, child.getId,Forum.class)).collect(toList()));
but its only my lamda quirk.

Other hand. It's changing whole list. its ok but remember that new list is created and old one should be dispose by GC. it can be replaced by using this same object (Forum) and only replace/update object (instead of create new list). 
Its just suggestion, don't treat this as problem

service/discussion/src/main/java/com/wikia/discussionservice/dao/ThreadDAO.java
```diff
@@ -41,6 +46,16 @@ public ThreadDAO(ContentDAO contentDAO) {
 
     contentDAO.createContent(siteId, forumThread);
 
+    // Update containing forum
+    Optional<Forum> optionalForum = forumDAO.getForum(siteId, forumId, 100, 0);
```
sqreek: 100 should be extract to const

service/discussion/src/main/java/com/wikia/discussionservice/domain/Forum.java
```diff
@@ -1,49 +1,61 @@
...
   public String name;
 
   @JsonProperty
-  @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS,
-      include = JsonTypeInfo.As.PROPERTY,
-      property = "@class")
-  @NonNull
   public List<Forum> children;
 
+//  @JsonProperty
```
sqreek: commented code :grey_question:

service/discussion/src/main/java/com/wikia/discussionservice/domain/Forum.java
```diff
@@ -51,16 +63,69 @@ public Forum() {
...
+  }
+
+  @Override
+  public Iterator<Forum> iterator() {
+    return new ForumIter(this);
+  }
+
+  @Override
+  public String toString() {
+    return "Forum_" + Long.toString(id) + "_" + name;
```
sqreek: I recommend to use String.format();

service/discussion/src/main/java/com/wikia/discussionservice/domain/ForumIter.java
```diff
@@ -0,0 +1,72 @@
...
+    this.childrenCurNodeIter = forum.children.iterator();
+  }
+
+  private ProcessStages doNext;
+  private Forum next;
+  private Iterator<Forum> childrenCurNodeIter;
+  private Iterator<Forum> childrenSubNodeIter;
+
+  @Override
+  public boolean hasNext() {
```
sqreek: This logic has many ifs.
Maybe extract it to smaller methods for make it more readable?

service/discussion/src/main/java/com/wikia/discussionservice/exceptions/DiscussionModuleException.java
```diff
@@ -0,0 +1,12 @@
+package com.wikia.discussionservice.exceptions;
+
+public class DiscussionModuleException extends RuntimeException {
```
sqreek: is RuntimeException is correct here?

service/discussion/src/main/java/com/wikia/discussionservice/guicemodule/DiscussionModuleBase.java
```diff
@@ -67,4 +68,9 @@ public ContentDAO provideContentDAO(
     return new ContentDAO(jedisService);
   }
 
+  @Provides
+  public ForumDAO provideForumDAO(ContentDAO contentDAO) {
```
sqreek: Is this could be replaced by bind(ForumDAO.class) in configure?

service/discussion/src/main/java/com/wikia/discussionservice/mappers/HALForumRepresentationMapper.java
```diff
@@ -78,20 +53,67 @@ public Representation buildRepresentation(int siteId, Forum forum, UriInfo uriIn
...
+  @Override
+  public Representation buildRepresentation(int siteId, List<Forum> forumTree,
+                                            UriInfo uriInfo, ResponseGroup responseGroup,
+                                            int limit, int offset) {
+
+    Forum root;
+    try {
+      root = forumTree.get(0);
+    } catch (Exception e) {
+      throw new RootForumNotFoundException("Error locating the root forum");
```
sqreek: Consider add cause to exception.

service/discussion/src/main/java/com/wikia/discussionservice/mappers/HALForumRepresentationMapper.java
```diff
@@ -78,20 +53,67 @@ public Representation buildRepresentation(int siteId, Forum forum, UriInfo uriIn
...
     }
+    Link linkToSelf = new LinkBuilder().buildLink(uriInfo, "self", ForumResource.class, "getForum",
+                                                  siteId, root.getId());
 
-    if (forum.hasThreads()) {
-      for (ForumThread thread : forum.getThreads()) {
-        representation.withRepresentation("doc:thread", threadMapper.buildRepresentation(
-            siteId, thread, uriInfo));
-      }
+    Representation representation =
```
sqreek: It looks like code duplicate with [this](https://github.com/Wikia/pandora/pull/146/files#diff-fb23b960f557920e2a950142de4a9f09R58)

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
```diff
@@ -29,20 +35,61 @@ public ForumService(ThreadService threadService, ForumDAO forumDAO) {
...
+
+  /**
+   * Get a flat list of forums starting with given root
+   *
+   * @param siteId int
+   * @param rootNode Forum
+   * @param flatTree List<Forum>
+   * @return List<Forum>
+   */
+  protected List<Forum> getFlatTree(int siteId, Forum rootNode, List<Forum> flatTree) {
```
sqreek: Is order here have matter?

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
```diff
@@ -63,4 +110,20 @@ public ForumService(ThreadService threadService, ForumDAO forumDAO) {
...
+  public List<ForumThread> getThreads(Forum forum) {
+    return getThreads(forum, THREAD_LIMIT_DEFAULT, 0);
+  }
+
+  public List<ForumThread> getThreads(Forum forum, int limit, int offset) {
+    List<ForumThread> threads;
+    try {
+      threads = forum.getThreads(limit, offset);
+    } catch (IndexOutOfBoundsException e) {
+      throw new DiscussionModuleException(
```
sqreek: Instead of rewriting message, maybe add cause to custom exception?

service/discussion/src/test/java/com/wikia/discussionservice/resources/ForumResourceTest.java
```diff
@@ -42,17 +41,19 @@
...
     // reset the mock because of the static nature of the ResourceTestRule
     reset(forumService);
-    
-    firstOffsetForumRoot = new ForumRoot();
-    firstOffsetForumRoot.setForums(createForumList(1, 10));
-    
-    when(forumService.getForums(anyInt())).thenReturn(Optional.of(firstOffsetForumRoot));
+
+    rootForum = mock(Forum.class);
+    rootForum.setChildren(createForumList(1, 10));
```
sqreek: This line is unnecessary

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
armonr: If you mean the order of elements in the list, yes, it matters

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
armonr: I rewrote `getThreads()` method so that it would not throw an exception; I'll remove try/catch from here accordingly

service/discussion/src/main/java/com/wikia/discussionservice/dao/ThreadDAO.java
armonr: Our DAO files are temporary throw-away files, hence the sloppiness ;)
They soon will be replaced with the implementation of DB-based DAOs so lots of such magic numbers are not even needed.

service/discussion/src/main/java/com/wikia/discussionservice/exceptions/DiscussionModuleException.java
armonr: Yes, intended. Any reason against it?

service/discussion/src/main/java/com/wikia/discussionservice/guicemodule/DiscussionModuleBase.java
jimmordino: This is temp code.  ForumDao will be an interface with a MyBatis implementation and there is a bind call for that.  Coming soon...

armonr: :+1: 
I already fixed this anyways.

service/discussion/src/main/java/com/wikia/discussionservice/domain/ForumIter.java
armonr: Done 9c8a22559d7edf40bf02367cd856f31876836b9d

jimmordino: Does this code process more than 3 levels?  I would implement this using a Stack because it would give you arbitrary depth. For ```next()`` implementation, grab ```next()``` from the top Iterator in the Stack and if that forum has children then push the Iterator for those children onto the Stack before returning it.

service/discussion/src/main/java/com/wikia/discussionservice/exceptions/DiscussionModuleException.java
sqreek: RuntimeExcpetion doesn't need to be handled by java. So if developer use this kind of exception, it can crash whole application. Others exceptions need to be covered by try catch, and java doesn't allow you to leave it without catching.

jimmordino: This is true, but the vast majority of checked exceptions can't be handled anyway.  In this case there's a higher level try/catch block that will convert a RuntimeException into a 500 level response, so we're covered.

jogura:
LGTM :+1:

PR #147, opened 2015-04-07
exclude org.mortbay.jetty

nmonterroso:
@sqreek 
This was causing integration tests to fail due to other implementations of `Request`/`Response` interfaces, similar to other issues we've had in the past.


nmonterroso:
Closing, since fixed as part of https://github.com/Wikia/pandora/pull/148. Thanks, @sqreek !

PR #148, opened 2015-04-07
move wiremock dependency to gateway module

sqreek:
Wiremock depends on jetty6 (org.mortbay.jetty), and it's not working in other modules.
Gateways module is not using jetty, so it should be fine.


nmonterroso:
LGTM

PR #152, opened 2015-04-09
Initial checkin of Shiro framework for Authn/Authz for Discussion Service

jogura:
This PR introduces the [Apache Shiro](http://shiro.apache.org) framework to handle Authn/Authz for the discussion service. The idea is to introduce an abstract `ServiceRealm` that will understand how to authenticate against Helios (using the Bearer tokens) and return the user's information as principals for the user. The principals will be used to retrieve the user's group information which can be used by the concrete class to fetch Service specific permissions.  
**NOTE: Neither of those pieces have been implement as of yet but maybe coming shortly.** 

The resources have been protected declaratively using the Shiro provided [JAX-RS annotations](http://shiro.apache.org/java-annotations-list.html).

**NOTE: In the near future there maybe an Authentication Proxy that may replace this**

RFCR @cmiller-wikia @drsnyder @pchojnacki @nmonterroso @jimmordino @armonr et al

More commit details:
* Create an abstract `ServiceRealm` that *will* in the near future interact with Helios for AuthN and some yet-to-be-created service for user group
subclasses of this realm must implement the `getAdditionalPermissions()` to add Service specific permissions
* Update the "Resources" with jax-rs annotations to declare the required Roles and Permissions access to the Resource's methods
* Configure shiro to have a security filter that checks for Bearer token and creates the appropriate token for the `ServiceRealm` implementation
* Refactored to use `ServiceExceptionMapper`
* Add unit and integration tests


lib/service-core/src/main/java/com/wikia/pandora/core/auth/Role.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Security Roles that represent groups
+ */
+public enum Role {
+  ADMIN,
+  ANON,
+  MOD,
+  STAFF,
+  USER,
+  VSTF;
```
armonr: This acronym begs a comment :) Volunteer Spam Task Force?

lib/service-core/src/main/java/com/wikia/pandora/core/auth/Role.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public String toString() {
+    return this.name().toLowerCase();
+  }
+
+  public static Role getRole(String roleName) {
+    roleName = roleName.toLowerCase();
+    if (roleName.equals(ADMIN.toString())) {
```
armonr: Consider using a switch statement

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceAuthenticatingFilter.java
```diff
@@ -0,0 +1,84 @@
...
+    String authorizationHeader = getAuthzHeader(request);
+    if (authorizationHeader == null || authorizationHeader.length() == 0) {
+      // Create an empty authentication token since there is no
+      // Authorization header.
+      return createToken("", request, response);
+    }
+
+    if (log.isDebugEnabled()) {
+      log.debug("Attempting to execute login with headers [" + authorizationHeader + "]");
+    }
```
drsnyder: Minor: Is the surrounding `if` required? Should `log` be `LOG`?

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceAuthenticatingFilter.java
```diff
@@ -0,0 +1,84 @@
...
+    return optionalAuthenticationToken
+        .map(authorizationToken -> createToken(authorizationToken, request, response))
+        .orElseGet(() -> createToken("", request, response));
+  }
+
+  @Override
+  protected boolean onAccessDenied(ServletRequest request, ServletResponse response)
+      throws Exception {
+    boolean isAuthenticated = false; //false by default or we wouldn't be in this method
+    isAuthenticated = executeLogin(request, response);
```
drsnyder: Does `onAccessDenied` belong here? In my mind, authentication is either success or failure. Denying access sounds more like an authorization concern.

service/discussion/src/main/java/com/wikia/discussionservice/auth/User.java
```diff
@@ -0,0 +1,43 @@
...
+  @NonNull
+  private final Subject subject;
+
+  public <T> T unwrap(Class<T> type) {
+    if (Subject.class.equals(type)) {
+      return type.cast(subject);
+    }
+
+    throw new IllegalArgumentException("User " + this + " cannot be unwrapped to " + type);
+  }
```
drsnyder: What is this for?

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceRealm.java
```diff
@@ -0,0 +1,119 @@
...
+   * Otherwise returns the user id representing an anonymous user.
+   *
+   * @return String userId
+   */
+  private String authenticateCredentials(String credentials) {
+    // TODO: Call out to Helios to authenticate the token and get the user info
+    String userId;
+    if (credentials.equalsIgnoreCase("1234567890")) {
+      // represents an staff user
+      userId = "1";
```
armonr: Why not use `Role`?

lib/service-core/src/main/java/com/wikia/pandora/core/auth/Role.java
drsnyder: IMO a switch isn't any better. You need to be able to dispatch based on the value of the parameter and not the type so as far as I know, the only way to improve this would be fake a multimethod with a lookup table that maps the `roleName` to the enum.

drsnyder:
Hopefully you won't have to worry about the [authentication part](https://wikia-inc.atlassian.net/browse/SERVICES-256).

service/discussion/src/main/java/com/wikia/discussionservice/auth/DiscussionServiceRealm.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Specific Authorization Realm for the Discussion Service
+ */
+public class DiscussionServiceRealm extends ServiceRealm {
+
+  @Override
+  protected Set<String> getAdditionalPermissions(String userId,
+                                                 Set<String> roles) {
+    Set<String> permissions = new HashSet<>();
```
armonr: Minor: initialization can be moved to the default case.

service/discussion/src/main/java/com/wikia/discussionservice/resources/ForumResource.java
```diff
@@ -151,25 +162,26 @@ public Response createForum(
 
   @DELETE
   @Path("/{siteId}/forums/{forumId}")
-  @ApiOperation(value = "Delete a forum for a site")
+  @ApiOperation(value = "Delete a forum for a site",
+      response = HalResponse.class, produces = RepresentationFactory.HAL_JSON)
   @ApiResponses(value = {
       @ApiResponse(code = 204, message = "Successful deletion of a forum"),
       @ApiResponse(code = 404, message = "No forum for the site", response = HalProblem.class)})
+  @RequiresRoles(value = {"mod", "staff", "admin"}, logical = Logical.OR)
```
armonr: Can `Role`s be used here?

service/discussion/src/main/resources/discussion.yaml
```diff
@@ -23,7 +23,10 @@ server:
   adminConnectors:
   - type: http
     port: 8081
+  registerDefaultExceptionMappers: false
```
armonr: Is this because of conflicts in responses?

lib/service-core/src/main/java/com/wikia/pandora/core/auth/Role.java
jimmordino: You don't want to have a condition per instance in any case.  Instead, spin through the roles by calling Role.values()

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceAuthenticatingFilter.java
jogura: This is part of the Shiro framework, not much we can do about it.

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceAuthenticatingFilter.java
jimmordino: It is if you don't want to build Strings for no reason at all.

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceRealm.java
jogura: Helios is currently not designed to return roles only user ids and additional information. If they can return a user's role then that would obviate the need for the other service that will be needed to return the roles.

service/discussion/src/main/java/com/wikia/discussionservice/resources/ForumResource.java
jogura: No these are annotations so unfortunately it is expecting strings and doing a `Role.MOD.toString()` doesn't work in this context.

service/discussion/src/main/resources/discussion.yaml
jogura: No, this is because of the changes made in #143

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceAuthenticatingFilter.java
drsnyder: I know this is nitpicking but unless the string building is a problem I would prefer to not see us wrapping logging statements like this. It adds clutter.

jogura: I think this question [even-with-slf4j-should-you-guard-your-logging](http://stackoverflow.com/questions/8444266/even-with-slf4j-should-you-guard-your-logging) answers the question nicely. And as usual the answer is it depends. In this situation, getting the parameters is not expensive so I removed the if and change to use a simple string format instead. 
b5842c8

armonr:
LGTM, discussion tests are all passing.

PR #156, opened 2015-04-13
add new image to mediawiki integration test

sqreek:
We need to change integration test from production to devboxes. 
This is only hotfix


ArturKlajnerok:
LGTM

ArturKlajnerok:
LGTM :+1:

PR #160, opened 2015-04-16
SOC-595 | Notifications rabbitMQ workers

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-595
From now rabbit workers are not bind to JSONObject. This allows to implement different message serialization methods in different services.

New workers:
* TaskDistributorQueueWorker - fetches all users following notified item
* DelivererQueueWorker - delivers notification to user inboxes


service/notifications/src/main/java/com/wikia/notificationsservice/queue/DelivererQueueWorker.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @Override
+  public String getBindingKey() {
+    return "notifications.toUser";
+  }
+
+  @Override
+  public Boolean handle(String jsonData, QueueingConsumer.Delivery message) throws IOException {
+    Gson gson = new Gson();
+    PublishForUserMessage publishMessage = gson.fromJson(jsonData, PublishForUserMessage.class);
```
rafalkalinski: shouldn't we try to catch an exception here?

service/notifications/src/main/java/com/wikia/notificationsservice/queue/TaskDistributorQueueWorker.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public String getBindingKey() {
+    return "notifications.toDistribute";
+  }
+
+  public Boolean handle(String jsonData, QueueingConsumer.Delivery message) throws IOException {
+    QueuePublisher publisher = new QueuePublisher(getRabbitConnection());
+    Gson gson = new Gson();
+
+    PublishForItemMessage publishMessage = gson.fromJson(jsonData, PublishForItemMessage.class);
```
rafalkalinski: try - catch?

rafalkalinski:
LGTM

lib/rabbit/src/main/java/com/wikia/rabbit/RabbitConnection.java
```diff
@@ -81,12 +80,12 @@ public void disconnectConsumerChannel() {
...
   }
 
-  public boolean publish(String routingKey, JSONObject data) {
+  public boolean publish(String routingKey, String jsonData) {
     try {
       Channel pChannel = getPublishChannel();
       pChannel.basicPublish(getExchange(), routingKey,
                             MessageProperties.PERSISTENT_TEXT_PLAIN,
-                            data.toString().getBytes("UTF-8"));
+                            jsonData.getBytes("UTF-8"));
```
pchojnacki: I think this could safely be named messageText 
since it seems that you:
declare the message as MessageProperties.PERSISTENT_TEXT_PLAIN
and you actually send String data.

pchojnacki:
One small Nitpick
Otherwise it LGTM

service/notifications/src/main/java/com/wikia/notificationsservice/queue/DelivererQueueWorker.java
kvas-damian: We catch it here: https://github.com/Wikia/pandora/pull/160/files#diff-1fc42d23cc4917088b3c178f6ce59219R44
Do you think we should creates special way of handling parsing json errors?

rafalkalinski:
LGTM

PR #161, opened 2015-04-16
Allows Swagger UI to work with our current Services configuration

jogura:
Uses a modified version of the dropwizard-swagger bundle that allows for specifying the a `staticAssetPrefix` that is used to find the static resources for the CSS & JS. This bundle is in our internal archiva repository.

In Consul for the specific environment a new key must be added. For example, if the service is hosted @ `http://api.wikia-services.com/discussion-qa` then the value of `SWAGGER_STATIC_ASSET_PREFIX` would be `/discussion-qa`.


armonr:
LGTM

PR #163, opened 2015-04-17
[DIS] DIS-13 - Initial version of content exporter and import scripts

jimmordino:
This is the first revision of the MVP export tool and import scripts used to export forum/thread content from wikias DBs and import that data into the DIS schema.

- [x] Command line export tool that emits a SQL script to insert forum/thread data into DIS schema
- [x] DIS schema creation script
- [x] DIS import script that takes data from import tables and inserts it into forum/thread/post schema tables

@jogura, @armonr, @frankfarmer


jimmordino:
Closing until I can stabilize the build - trying to break up a large amount of changes into 2 PRs.

PR #164, opened 2015-04-17
[DIS] DIS-13 - Initial version of content exporter and import scripts

jimmordino:
This is the first revision of the MVP export tool and import scripts used to export forum/thread content from wikias DBs and import that data into the DIS schema.
 
- [x] Command line export tool that emits a SQL script to insert forum/thread data into DIS schema
- [x] DIS schema creation script
- [x] DIS import script that takes data from import tables and inserts it into forum/thread/post schema tables

@jogura, @armonr, @frankfarmer


service/discussion/sql/create_discussion_db.sql
```diff
@@ -0,0 +1,520 @@
...
+
+-- -----------------------------------------------------
+-- Schema discussion
+-- -----------------------------------------------------
+DROP SCHEMA IF EXISTS `discussion` ;
+
+-- -----------------------------------------------------
+-- Schema discussion
+-- -----------------------------------------------------
+CREATE SCHEMA IF NOT EXISTS `discussion` DEFAULT CHARACTER SET utf8mb4 ;
```
armonr: Why `utf8mb4`?

jimmordino: We're storing uncompressed post data in our DB and UTF-8 didn't cut it - there were import errors because of "invalid" characters.  utf8mb4 fixed all that.  https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb4.html

service/discussion/sql/create_discussion_db.sql
```diff
@@ -0,0 +1,520 @@
...
+
+-- -----------------------------------------------------
+-- Table `user`
+-- -----------------------------------------------------
+DROP TABLE IF EXISTS `user` ;
+
+CREATE TABLE IF NOT EXISTS `user` (
+  `id` INT NOT NULL,
+  `site_id` INT NOT NULL,
+  `name` VARCHAR(256) NULL DEFAULT NULL,
```
armonr: Is the first `NULL` needed?

service/discussion/sql/create_discussion_db.sql
jimmordino: Another note on this - please be sure your version of MySQLWorkbench (for Mac OS X) is at least version 6.2 - the C++ driver in prior versions explodes when you use utf8mb4.

service/discussion/sql/create_discussion_db.sql
```diff
@@ -0,0 +1,520 @@
...
+USE `discussion` ;
+
+-- -----------------------------------------------------
+-- Table `user`
+-- -----------------------------------------------------
+DROP TABLE IF EXISTS `user` ;
+
+CREATE TABLE IF NOT EXISTS `user` (
+  `id` INT NOT NULL,
+  `site_id` INT NOT NULL,
```
armonr: Should users be tied to a particular site?

service/discussion/sql/create_discussion_db.sql
```diff
@@ -0,0 +1,520 @@
...
+CREATE SCHEMA IF NOT EXISTS `discussion` DEFAULT CHARACTER SET utf8mb4 ;
+USE `discussion` ;
+
+-- -----------------------------------------------------
+-- Table `user`
+-- -----------------------------------------------------
+DROP TABLE IF EXISTS `user` ;
+
+CREATE TABLE IF NOT EXISTS `user` (
+  `id` INT NOT NULL,
```
armonr: Is INT large enough? :)

service/discussion/sql/create_discussion_db.sql
jimmordino: No, this is auto-generated from MySQLWorkbench.  I can pare it down and remove NULL attributes and NULL defaults if you like.

service/discussion/sql/create_discussion_db.sql
jimmordino: Should be INT UNSIGNED to match wikia DBs.  I will fix

service/discussion/sql/create_discussion_db.sql
armonr: In that case, I'm fine leaving it as is.

service/discussion/sql/create_discussion_db.sql
jimmordino: No, they aren't in the wikia world.  To the extend that we have users in the DIS DB the user table will be global as it is in MediaWiki and anything related to a user in the context of a site (i.e. forum, thread read) will be in a separate table with a site ID.

service/discussion/sql/create_discussion_db.sql
armonr: Not surprised about encoding issues :)
:+1:

service/discussion/sql/create_discussion_db.sql
armonr: Does that mean, this table can get as large as `count(wikia_users) * count(wikia_sites)`?

jimmordino: No, it means that the site_id column needs to go away - fixing...

service/discussion/sql/create_discussion_db.sql
```diff
@@ -0,0 +1,520 @@
...
+  `post_id` BIGINT NOT NULL,
+  `raw_content` TEXT NOT NULL,
+  `rendered_content` TEXT NULL DEFAULT NULL,
+  PRIMARY KEY (`id`))
+ENGINE = InnoDB
+DEFAULT CHARACTER SET = utf8mb4;
+
+CREATE INDEX `created_by_idxfk_1` ON `post_revision` (`created_by` ASC);
+
+CREATE INDEX `post_id_idxfk` ON `post_revision` (`post_id` ASC);
```
armonr: Are `post_id` sequence consecutiveness guaranteed to match that of their corresponding timestamps'? In other words, will the post_id for post A at T1 be always lower than post B at T2?

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/ExportConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+  private WikiaEnvironment environment;
+  private String centralDbName;
+  private String centralDbHostIp;
+  @Getter
+  private String dbName;
+  @Getter
+  private String dbHostIp;
+  @Getter
+  private String dbUser;
+  @Getter
```
jogura: You can simplify by putting one `@Getter` above the class declaration and remove the individual.

service/discussion/sql/create_discussion_db.sql
jimmordino: This is the intent.  In reality, the time in the ID matching the time in the created_at timestamp depends on whether or not the stored function gets the same value when it calls for time during the insert trigger as the insert statement that triggered it.  Also, the time portion of the ID is in milliseconds while the post timestamp is stored with 1 second granularity, so it depends on whether values are rounded or truncated, etc.  

The design calls for globally unique IDs to be temporally sortable, so assuming T2 > T1, ID  B will be greater than ID A.

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/ExportConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+    InputStream configInput = null;
+    try {
+      configInput = new FileInputStream(new File(configFilename));
+    } catch (FileNotFoundException fnfe) {
+      // This tool must be run on a box where the config repo has been deployed
+      // in /usr/wikia/source/config
+      String error = "Unable to locate wikia configuration in /usr/wikia";
+      System.err.println(error);
+      throw new WikiaExportException(error);
+    }
```
jogura: Would using a [try-with-resource](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) simplify this code?

service/discussion/sql/create_discussion_db.sql
armonr: :+1: Thanks!

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/WikiaBlob.java
```diff
@@ -0,0 +1,60 @@
...
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class WikiaBlob {
+  @SuppressFBWarnings("EI_EXPOSE_REP")
+  private byte[] contentBytes;
+
+  /**
+   * Extreact content from a page blob.
```
jogura: typo

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/ExportConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  /**
+   * Extract the values we need from the Wikia config yaml.
+   */
+  protected void extractConfig() {
+    // Extract names and IPs for the site and archive (blob) DBs
+    String centralClusterName = this.getClusterName(CONFIG_DB_NAME);
+    this.centralDbName = CONFIG_DB_NAME;
```
jogura: Is this needed?

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/WikiaBlob.java
```diff
@@ -0,0 +1,60 @@
...
+   * @return String
+   */
+  public String getInflatedContent(Charset charset) {
+    InputStreamReader reader = null;
+    StringWriter writer = new StringWriter();
+    InflaterInputStream inputStream = null;
+
+    try {
+      inputStream =
+          new InflaterInputStream(new ByteArrayInputStream(contentBytes), new Inflater(true));
```
jogura: Please add a little bit of documentation explaining where the `contentBytes` comes from?

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/WikiaConfigVariable.java
```diff
@@ -0,0 +1,31 @@
...
+   * @return Object
+   */
+  @SneakyThrows(SerializedPhpParserException.class)
+  public <T> T getValue() {
+    SerializedPhpParser parser = new SerializedPhpParser(this.serializedValue);
+    Optional<T> value = Optional.ofNullable((T) parser.parse());
+    if (!value.isPresent()) {
+      throw new IllegalStateException("Unable to parse value: " + this.serializedValue);
+    }
+    return value.get();
```
jogura: Consider using `.map().orElseThrows()` here

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/ExportConfiguration.java
jimmordino: I think so - it's now exposed via getter (based on your suggestion above).  The fact that it's set to a private constant is an implementation detail that need not be exposed to a consumer.  So it's either set it here like this or create a custom getter that just returns the constant - this seems simpler.

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/WikiaDbUtil.java
```diff
@@ -0,0 +1,261 @@
...
+   * @param dbName instance name of the DB
+   * @return a WikiaBlobMapper with a connection to the specified DB
+   */
+  public static WikiaBlobMapper getArchiveMapper(
+      @NonNull ExportConfiguration exportConfig,
+      @NonNull String dbHostIp,
+      @NonNull String dbName) {
+
+    Optional<SessionEntry> existingEntry =
+        Optional.ofNullable(sessionMap.get(dbName));
```
jogura: I don't believe the use of `Optional` in this context is required. You are not returning the value from sessionMap outside of this method.

service/discussion/sql/create_discussion_db.sql
```diff
@@ -0,0 +1,519 @@
...
+  `site_id` INT NOT NULL,
+  `title` VARCHAR(512) NULL DEFAULT NULL,
+  `created_at` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `modified_at` DATETIME NULL,
+  `mod_modified_at` DATETIME NULL,
+  `created_by` INT NOT NULL,
+  `last_modified_by` INT NULL DEFAULT NULL,
+  `thread_id` BIGINT NOT NULL,
+  `position` INT NULL DEFAULT NULL,
+  `deleted_position` INT NULL DEFAULT NULL,
```
armonr: What is the use case for `deleted_position`?

service/discussion/sql/create_discussion_db.sql
```diff
@@ -0,0 +1,519 @@
...
+  `responder_id` INT NULL DEFAULT NULL,
+  `responded_at` DATETIME NULL,
+  `validated_ind` TINYINT(1) NOT NULL DEFAULT '1',
+  PRIMARY KEY (`id`))
+ENGINE = InnoDB
+DEFAULT CHARACTER SET = utf8mb4;
+
+CREATE INDEX `post_id_idxfk_2` ON `flag` (`post_id` ASC);
+
+CREATE INDEX `responder_id_idxfk` ON `flag` (`responder_id` ASC);
```
armonr: Another index for `reported_by`?

service/discussion/sql/create_discussion_db.sql
jimmordino: This is from Huddler.  My assumption is that when a post is soft deleted is last position is kept in deleted_position and the positions of all subsequent posts are decremented.  If the post is then undeleted we know where to position it (and we'd have to increment all subsequent positions).  If position is always dictated by timestamp (and now therefore by ID) then this wouldn't be needed.

service/discussion/sql/create_discussion_db.sql
jimmordino: Possibly - lots of tuning to do.  I'll add that when I add UNSIGNED.

service/discussion/sql/create_discussion_db.sql
```diff
@@ -0,0 +1,519 @@
...
+
+CREATE INDEX `post_id_idxfk_1` ON `post_draft` (`post_id` ASC);
+
+
+-- -----------------------------------------------------
+-- Table `thread_move`
+-- -----------------------------------------------------
+DROP TABLE IF EXISTS `thread_move` ;
+
+CREATE TABLE IF NOT EXISTS `thread_move` (
```
armonr: As discussed offline, this table can be generic to hold move data for all different content types.

service/discussion/sql/create_discussion_db.sql
```diff
@@ -0,0 +1,519 @@
...
+
+CREATE TABLE IF NOT EXISTS `id_sequence` (
+  `id` INT NOT NULL AUTO_INCREMENT,
+  `stub` TINYINT(1) NOT NULL,
+  PRIMARY KEY (`id`))
+ENGINE = InnoDB
+DEFAULT CHARACTER SET = utf8mb4;
+
+CREATE UNIQUE INDEX `stub_UNIQUE` ON `id_sequence` (`stub` ASC);
+
```
armonr: :crystal_ball:  : )

service/discussion/sql/create_discussion_db.sql
```diff
@@ -0,0 +1,519 @@
...
+VALUES (1, 0);
+
+-- -----------------------------------------------------
+-- Table `config_options`
+-- -----------------------------------------------------
+DROP TABLE IF EXISTS `config_options` ;
+
+CREATE TABLE IF NOT EXISTS `config_options` (
+`name` VARCHAR(128) NOT NULL,
+`value` TEXT NULL,
```
armonr: Is site_id not needed? Do we want to have wiki-level configurations?

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/ExportConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+
+    // Load the environment and pick the appropriate hosts for config, db, and archive
+    String configFilename = String.format(ExportConfiguration.CONFIG_FILE_TEMPLATE,
+                                          this.environment.getFileTag());
+    InputStream configInput = null;
+    try {
+      configInput = new FileInputStream(new File(configFilename));
+    } catch (FileNotFoundException fnfe) {
+      // This tool must be run on a box where the config repo has been deployed
+      // in /usr/wikia/source/config
```
armonr: This comment might be more helpful as part of the error message.

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/ExportConfiguration.java
```diff
@@ -0,0 +1,260 @@
...
+   * Initializes the configuration.  We parsing YAML and much can go wrong, so we don't do this
+   * in a constructor.
+   *
+   * @param siteId the site id
+   * @param exportType the export type
+   * @param environment the environment
+   */
+  public void initialize(@NonNull Integer siteId,
+                         @NonNull ExportType exportType,
+                         @NonNull WikiaEnvironment environment) {
```
armonr: Should this method indicate or document throwing `WikiaExportException`?

service/discussion/src/main/resources/logback.xml
```diff
@@ -0,0 +1,14 @@
+<configuration>
```
armonr: Where is this file used?

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/WikiaDbUtil.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+    return object.toString();
+  }
+
+  /**
+   * Hack to convert an object into an escaped SQL string.
+   * @param object the object to convert
+   * @return String "null" or an escaped, quoted SQL string
+   */
+  public static String nullOrQuotedString(Object object) {
```
armonr: Can something like [StringEscapeUtils](http://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringEscapeUtils.html) be used instead?

service/discussion/build.gradle
```diff
@@ -18,7 +18,14 @@ dependencies {
     compile "com.theoryinpractise:halbuilder-core:4.0.4-wikia"
     compile project(":lib:service-core")
     compile "io.federecio:dropwizard-swagger:0.6.1-wikia"
-
+    compile "io.dropwizard:dropwizard-db:0.8.0"
+    compile "org.mybatis:mybatis:3.2.5"
+    compile "org.mybatis:mybatis-guice:3.6"
+    compile "org.yaml:snakeyaml:1.12"
```
jogura: Could we use the higher level `jackson-dataformat-yaml` to parse the YAML file? I believe that snakeyaml is used at the lower level anyways using the jackson version may avoid potential class conflicts.

service/discussion/src/main/java/com/wikia/discussionservice/tools/export/WikiaDbUtil.java
jogura: Based on the javadocs there are limitations: 

>At present, this method only turns single-quotes into doubled single-quotes ("McHale's Navy" => "McHale''s Navy"). It does not handle the cases of percent (%) or underscore (_) for use in LIKE clauses.

service/discussion/sql/create_discussion_db.sql
jimmordino: I think we need to be really sure it's a good idea to store wikia-specific config in our DB as opposed to WikiFactory.

service/discussion/src/main/resources/logback.xml
jimmordino: This file configures the default log level for the exporter process during Intellij debugging.  As far as I can tell, It has absolutely no effect on the service because dropwizard doesn't user logback.  I suppose it's a reasonable candidate for an ignore file.

jogura:
LGTM :+1:

PR #165, opened 2015-04-19
Add an API gateway AuthFactory provider for handling authenticated requests behind the gateway

drsnyder:
https://wikia-inc.atlassian.net/browse/SERVICES-248

The context of this is that the API gateway will authenticate with Helios and then add an `X-Wikia-UserId` header to the service request. The module provided in this PR will intercept that header and provide a user object to the resource so it can decide how to handle the request based on the user id.

Add a dropwizard 0.8.0 compatible AuthFactor provider. This includes a POJO for the user credentials & authenticated user plus the provider tests.

The dropwizard [AuthFactory](https://github.com/dropwizard/dropwizard/blob/v0.8.0/dropwizard-auth/src/main/java/io/dropwizard/auth/AuthFactory.java) is a little clunky. It looks like they refactored this in the current release candidate. 

To add this authenticator to an application, you do the following in the run phase:

```java
   environment.jersey().register(AuthFactory.binder(new APIGatewayAuthFactory<>(
        new APIGatewayAuthenticator(), AuthenticatedUser.class
    )));
```

You can then add `@Auth AuthenticatedUser user` to the params of your resource. See the [dropwizard docs](http://dropwizard.github.io/dropwizard/manual/auth.html) for more details.

Todo:
 * [ ] Clean up the tests and use a mock object for the authenticator
 * [ ] Test the exception path

/cc @pchojnacki @sqreek @ArturKlajnerok


lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
```diff
@@ -0,0 +1,77 @@
...
+    if (request != null) {
+      final String gatewayProvidedUserId = request.getHeader(GATEWAY_USER_ID_HEADER);
+      if (gatewayProvidedUserId == null) {
+        return (T) ANONYMOUS_USER;
+      }
+
+      final Integer userId = Integer.valueOf(gatewayProvidedUserId);
+      final APIGatewayCredentials apiGatewayCredentials = APIGatewayCredentials.builder()
+          .userId(userId)
+          .build();
```
jogura: We could simplify this logic by using [Apache Commons NumberUtils.toInt(string, default)](http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html). Attempt to get the `userId` otherwise use a default of `0` which will return the anonymous user.

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
```diff
@@ -0,0 +1,77 @@
...
+
+      try {
+        final Optional<T> result = authenticator().authenticate(apiGatewayCredentials);
+        if (result.isPresent()) {
+          return result.get();
+        }
+      } catch (AuthenticationException e) {
+        LOGGER.warn("Error authenticating credentials", e);
+        throw new InternalServerErrorException();
+      }
```
jogura: Could this block be refactored to use the `map().orElseGet()` syntax? 
Not sure if catching/throwing exceptions in the lambda is acceptable and or desired. 
So this is more of a question.

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
drsnyder: Thanks for the tip! I love `NumberUtils`! That cut out 5-6 lines of string->int handling 6ac5fdd.

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
drsnyder: I used `.or` in 6ac5fdd. It's a shame that the authenticator interface forces the throws on you. In this case I'm not sure how much sense it makes but we have handle it regardless.

drsnyder:
Last call for comments.

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
```diff
@@ -0,0 +1,89 @@
...
+ * @param <T>
+ */
+public class APIGatewayAuthFactory<T> extends AuthFactory<APIGatewayCredentials, T> {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(APIGatewayAuthFactory.class);
+
+  private static final AuthenticatedUser ANONYMOUS_USER =
+      AuthenticatedUser.builder().userId(0).build();
+
+  public static final String GATEWAY_USER_ID_HEADER = "X-Wikia-UserId";
```
armonr: Does this need to be exposed?

PR #168, opened 2015-04-20
Append to environment instead of overriding it

pchojnacki:
bad environment settings in gradle caused it to ignore environment overrides


sqreek:
LGTM :shipit:

PR #170, opened 2015-04-20
DIS-15 WIP Remove Jedis and add MyBatis to DIS service

jimmordino:
Synchronizing to MyBatis to allow parallel development without Jedis.

- [x] Remove Jedis references and libraries
- [x] Add MyBatis libraries
- [x] Create managed MyBatis datasource with connection pool based on config yaml
- [x] DAOs become interfaces with MyBatis implementations
- [x] Name changes: Thread => ForumThread (there will be other thread types), camel case acronyms
- [x] Implement content type with an enum
- [x] Add fields to domain objects based on preliminary schema
- [x] Fix Checkstyle and Findbugs issues
- [x] Port all functionality implemented with Jedis to Mybatis (Forum resources)
- [x] Create DAOs and Mybatis mappers for all entities
- [ ] Health checks for DB
- [ ] Unit tests

@jogura, @armonr


service/discussion/src/test/java/com/wikia/discussionservice/domain/ForumTest.java
```diff
@@ -33,15 +38,25 @@ public void serializesAndDeserializesFromJSON(String fixtureFile, int id, Forum
...
+    Forum forum2 = new Forum();
+    forum2.setId(4L);
+    forum1.setSiteId(1);
+    forum2.setParentId(parentForum.getId());
+    forum2.setName("child 2");
+
+    List<Forum> children = Arrays.asList(forum1, forum2);
 
     return new Object[][]{
         {"fixtures/forumWithoutChildren.json", 2, parentForum, "Top Level", new ArrayList<>(),
```
armonr: This is causing test failure

service/discussion/src/test/java/com/wikia/discussionservice/domain/ForumIterTest.java
```diff
@@ -1,34 +1,67 @@
...
+
+    Forum nodeD = new Forum();
+    nodeD.setId(103L);
+    nodeD.setParentId(nodeC.getId());
+    nodeD.setName("child of C");
+
+    Forum nodeD1 = new Forum();
+    nodeD1.setId(113L);
+    nodeD1.setParentId(nodeD.getId());
+    nodeD1.setName("child of C");
```
armonr: Not that it makes any difference in the test result, but for better readability, the names of each node should correspond to its immediate parent's.

service/discussion/src/main/resources/discussion.yaml
```diff
@@ -24,11 +24,24 @@ server:
...
   iniConfigs: ["classpath:discussion.shiro.ini"]
 swaggerConfiguration:
   staticAssetPrefix: ${consul:SWAGGER_STATIC_ASSET_PREFIX:/}
+databaseConfiguration:
+  dataSourceFactory:
+    driverClass: com.mysql.jdbc.Driver
+    user: ${consul:DB_USER:"discussion"}
+    password: ${consul:DB_PASSWORD:""}
+    url: ${consul:DB_URL:"jdbc:mysql://localhost:3306/discussion"}
+    autoCommitByDefault: false
```
armonr: :+1:  I wish turning this off for media wiki was this easy!

service/discussion/src/main/java/com/wikia/discussionservice/domain/ForumRoot.java
```diff
@@ -0,0 +1,42 @@
+package com.wikia.discussionservice.domain;
```
armonr: Did you intentionally re-introduce ForumRoot? It was previously removed [here](https://github.com/Wikia/pandora/pull/146/files#diff-1da241aeb3dabcb3fd18f87acd2f9e18L1)

jimmordino: No, it's product of merges over time.  Probably merged in ForumRoot and had to change it in my branch, which caused a conflict on the subsequent merge and I resolved the wrong way.  Or something like that.  In any case, I'll take it back out.

service/discussion/src/main/java/com/wikia/discussionservice/dao/ContentDao.java
```diff
@@ -0,0 +1,18 @@
...
+import com.wikia.discussionservice.domain.Content;
+
+import java.lang.Class;
+import java.util.List;
+
+public interface ContentDao {
+  public <T extends Content> List<T> getItems(Integer siteId,
+                                              Class<? extends Content> contentClass);
+
+  public <T extends Content> T getContent(Integer siteId, Long contentId, Class<T> contentClass);
```
armonr: Maybe siteId can be Long too, since it is `UNSIGNED` in mysql, just for consistency.

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/ForumThreadDaoImpl.java
```diff
@@ -0,0 +1,65 @@
...
+  @NonNull
+  ContentDao contentDao;
+
+  @Inject
+  public ForumThreadDaoImpl(ContentDao contentDao) {
+    this.contentDao = contentDao;
+  }
+
+  @SuppressFBWarnings("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
+  public Optional<ForumThread> createThread(Integer siteId, ContentType threadType,
```
jogura: What's `ContentType threadType` used for?

service/discussion/src/main/java/com/wikia/discussionservice/domain/Content.java
```diff
@@ -1,8 +1,18 @@
 package com.wikia.discussionservice.domain;
 
+
+import com.wikia.discussionservice.enums.ContentType;
+
+/**
+ * Created by armon on 3/2/15.
+ */
```
armonr: Flash back!

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/ForumThreadDaoImpl.java
```diff
@@ -0,0 +1,65 @@
...
+                                            Long forumId, Post post) {
+
+    ForumThread forumThread = new ForumThread();
+    forumThread.setForumId(forumId);
+    forumThread.setPosts(new ArrayList<>(Arrays.asList(post)));
```
jogura: I'm digging this to build/initialize lists `com.google.common.collect.Lists.newArrayList()`. It seems more compact than `new ArrayList<>(Arrays.asList()`

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/ContentDaoImpl.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.wikia.discussionservice.dao.ContentDao;
+import com.wikia.discussionservice.domain.Content;
+
+import java.util.ArrayList;
+
+
+/**
+ * Created by armon on 3/2/15.
+ */
```
armonr: No it was not! :) #please-remove

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/MyBatisFactory.java
```diff
@@ -0,0 +1,90 @@
...
+      Configuration mybatisConfiguration = new Configuration(myBatisEnvironment);
+      sessionFactory = new SqlSessionFactoryBuilder().build(mybatisConfiguration);
+    }
+
+    environment.lifecycle().manage(dataSource);
+//    environment.healthChecks().
+//        register(name,
+//                 new MyBatisHealthCheck(sessionFactory,
+//                                        name,
+//                                        dataSourceFactory.getValidationQuery());
```
armonr: Should this be added?

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/MyBatisFactory.java
```diff
@@ -0,0 +1,90 @@
...
+   * @param environment The dropwizard environment
+   * @param config      A Mybatis config object
+   * @param name        The name of this mybatis factory used for metrics
+   * @return An instance of MyBatis.
+   */
+  public final SqlSessionFactory build(Environment environment,
+                                       DiscussionDatabaseConfiguration config,
+                                       String name) {
+
+    SqlSessionFactory sessionFactory = null;
```
armonr: Trivial: initialization not needed.

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
```diff
@@ -81,51 +83,41 @@ public ForumService(ThreadService threadService, ForumDAO forumDAO) {
...
+    ForumRoot forumRoot = new ForumRoot();
+    forumRoot.setForums(forumList);
+    forumRoot.setSiteId(siteId);
+    return Optional.of(forumRoot);
   }
 
-  public Optional<Forum> getForum(int siteId, long forumId, int limit, int offset) {
-    return forumDAO.getForum(siteId, forumId, limit, offset);
+  public Optional<Forum> getForum(Integer siteId, Long forumId) {
+    return getForum(siteId, forumId, 0, 10);
```
jogura: We should move these into configurable options or constants at the very minimum.

armonr: :+1: for constants. Although, it wont alleviate having to repeat the same values in the Resource methods.

service/discussion/src/main/java/com/wikia/discussionservice/dao/ForumDao.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.List;
+import java.util.Optional;
+
+public interface ForumDao {
+
+  List<Forum> getForumList(Integer siteId);
+
+  Optional<Forum> getForum(Integer siteId, Long forumId);
+
+  Optional<Forum> createForum(Forum forum);
```
jogura: I think we should add `siteId` to this API.

service/discussion/src/main/java/com/wikia/discussionservice/dao/ForumDao.java
```diff
@@ -0,0 +1,20 @@
...
+
+  Boolean deleteForum(Integer siteId, Long forumId);
+
+  Optional<Forum> updateForum(Forum forum);
```
jogura: Add `siteId` to this one as well.

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
```diff
@@ -81,51 +83,41 @@ public ForumService(ThreadService threadService, ForumDAO forumDAO) {
...
-            return updateForum(siteId, forum);
-          })
-          .orElseGet(() -> Optional.empty());
-    } catch (DiscussionModuleException e) {
-      return Optional.empty();
-    }
-  }
+  // FIXME - rename to "save" which can be create or update
+  public Optional<Forum> updateForum(Integer siteId, Forum forum) {
+    Optional<Forum> updatedForum = forumDao.updateForum(forum);
```
jogura: If we don't want to add `siteId`to the `ForumDao` interface then we should at the minimum set the `siteId` of the `forum` instance. We have no guarantee that `Forum.siteId` is set.

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
```diff
@@ -81,51 +83,41 @@ public ForumService(ThreadService threadService, ForumDAO forumDAO) {
...
 
-  private Optional<Forum> updateForum(int siteId, Forum forum) {
-    Optional<Forum> updatedForum = forumDAO.updateForum(siteId, forum);
-    
     if (updatedForum.isPresent()) {
       return updatedForum;
     }
```
jogura: I don't believe we'll ever execute this path. If the the code [here](https://github.com/Wikia/pandora/pull/170/files#diff-2e9d0f3a4c432b51829baa7a3192ed4aR72) cannot update the Forum (due to it not existing) an exception is thrown. Which will bypass this guard.

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/ForumThreadDaoImpl.java
jimmordino: WIP, working out the interface.  I'd like to have a DiscussionThread superclass with subclasses like ForumThread, BlogThread, CommentThread, WallThread, etc.   This is probably represents my thinking about whether or not a common create method can be used for all of them.  Obviously not needed for a ForumThread specific method.

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/MyBatisFactory.java
jimmordino: WIP (you wanted a WIP PR, you got one)

service/discussion/src/main/java/com/wikia/discussionservice/dao/ForumDao.java
jimmordino: Not at the DAO level - there's no point.  site ID is already embedded in the Forum object.

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
jimmordino: The only reason siteId is a separate field at the API level is for routing purposes.  Everything below that should not have separated out - this is WIP so it's not gone yet.  You can say the same about any required field not being there.  If the ID isn't correct, the update will fail.  If the parent ID isn't set the update will fail.  And so on.  It's up to the API caller to ensure that the data in the Forum object is correct, including site ID and to additionally pass site ID when the API requires it.

service/discussion/src/main/java/com/wikia/discussionservice/dao/ContentDao.java
jimmordino: That would be a waste, I think.  We have over 2 billion positive site ids to work with, so there's a lot of headroom right now.  If we ever get over 2.15 billion sites,  the numbers will start to be negative in Java, but should still function correctly when loading from/writing to the DB.  Let's worry about that when the time comes.

service/discussion/src/main/java/com/wikia/discussionservice/dao/ForumDao.java
jimmordino: As discussed, we'll remove siteId as an argument from the service layer on down.  We'll also add object/arg validation at the resource layer to be sure the API siteId arg matches the Forum siteId field and that the Forum field is set.

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceAuthenticatingFilter.java
```diff
@@ -7,45 +7,37 @@
...
 public class ServiceAuthenticatingFilter extends AuthenticatingFilter {
 
   private static final Logger log = LoggerFactory.getLogger(ServiceAuthenticatingFilter.class);
 
-  /**
-   * HTTP Authorization header, equal to <code>Authorization</code>
-   */
-  protected static final String AUTHORIZATION_HEADER = "Authorization";
-
+  public static final String GATEWAY_USER_ID_HEADER = "X-Wikia-UserId";
```
drsnyder: Does it make sense to have some global constants since this is the [second usage of this string](https://github.com/Wikia/pandora/blob/45987cee8e6d68389301de854b45a51062352d68/lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java#L26)?

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceAuthenticatingFilter.java
```diff
@@ -7,45 +7,37 @@
...
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 
 /**
- * ServiceAuthenticatingFilter will create an AuthenticationToken with the Bearer token so it can be
- * authenticated against Helios.
+ * ServiceAuthenticatingFilter will create an AuthenticationToken based on the existence of
+ * a user id. The user id will only be present if the call came from the API Gateway.
  */
 public class ServiceAuthenticatingFilter extends AuthenticatingFilter {
```
drsnyder: I wonder if this should be the default auth filter for services. The [dropwizard provided one is clunky](https://github.com/Wikia/pandora/blob/45987cee8e6d68389301de854b45a51062352d68/lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java#L24).

lib/service-core/src/main/java/com/wikia/pandora/core/dropwizard/exceptions/ValidationExceptionMapper.java
```diff
@@ -0,0 +1,44 @@
...
+              .instance(URI.create(request.getRequestURI()))
+              .build());
+    }
+
+    return StandardResponses.InternalServerError(
+        Problem.builder()
+            .title(
+                "The request could not be processed.")
+            .instance(URI.create(request.getRequestURI()))
+            .build());
```
drsnyder: Is this still a 400 of some kind and not a 500 since it's some form of `ValidationException`?

service/discussion/sql/import_forum_data.sql
```diff
@@ -125,11 +136,37 @@ AND import_revision.revision_id = (
   SELECT revision_id
   FROM import_revision
   WHERE import_revision.page_id = import_page.page_id
-  AND import_revision.title IS NOT NULL
+  AND import_revision.title > ''
```
drsnyder: That looks weird. Does this mean not empty?

jimmordino: In MySql, this means not blank and not null.  "Empty" is an overloaded term in our world :smile:

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/MyBatisFactory.java
```diff
@@ -0,0 +1,90 @@
...
+    final ManagedDataSource dataSource = dataSourceFactory.build(environment.metrics(), name);
+
+    // Try to use the mybatis configuration file if it is specified and exists.
+    try (InputStream inputStream = Resources.getResourceAsStream(config.getConfigFile())) {
+      sessionFactory = new SqlSessionFactoryBuilder().build(inputStream);
+    } catch (IOException ioe) {
+      // Build session factory from configuration values given in the dropwizard config.
+      TransactionFactory transactionFactory = new JdbcTransactionFactory();
+      org.apache.ibatis.mapping.Environment myBatisEnvironment =
+          new org.apache.ibatis.mapping.Environment("FIXME", transactionFactory, dataSource);
```
drsnyder: `"FIXME"` Is that a reminder :)

jimmordino: Yes, this code was transitional, is not currently used, and will likely be deleted entirely when we move to Nelson's common DB configuration classes.

PR #171, opened 2015-04-21
add update user preference functionality

sqreek:
Added Update user preference functionality


sqreek:
Refactor authentication flow
Bump dropwizard Version to 8.1
add update my preferences functionality
add no other user can edit or read my preferences
and that handle [SERVICE-161](https://wikia-inc.atlassian.net/browse/SERVICES-161) and [SERVICE-190](https://wikia-inc.atlassian.net/browse/SERVICES-190)
add tests after mini test course with @ludwikkazmierczak 

to answer any question that may come up:
word test at every testMethod in Test class is meaningless.
If you agree just write noTestWord:+1:  or noTestWord :-1: and we can discuss this

service/user-preference/src/test/java/com/wikia/userpreference/resources/UserPreferenceResourceIntegrationTest.java
```diff
@@ -3,6 +3,7 @@
 import static org.junit.Assert.assertEquals;
```
sqreek: I'm thinking about delete those tests, becuse DAO integration tests and Resource Unit tests are solid

sqreek:
ping @drsnyder @pchojnacki @ArturKlajnerok @gbenson-wikia 

and btw. There is still failing tests from community data. Can we do something with this?

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
```diff
@@ -57,33 +79,27 @@ public void setRequest(HttpServletRequest request) {
...
-   * @return <T, U>Binder
-   */
-  public static <T, U>Binder create() {
-    return AuthFactory.binder(new APIGatewayAuthFactory<>(
-        new APIGatewayAuthenticator(), AuthenticatedUser.class
-    ));
+    if (required) {
+      throw new WebApplicationException(Response.status(Response.Status.UNAUTHORIZED).build());
+    }
+    return null;
```
drsnyder: According to the guidelines we should [avoid returning null](https://github.com/Wikia/guidelines/tree/master/Java#do-not-return-null-or-accept-null-as-an-argument). That was the original motivation behind the [`ANONYMOUS_USER` object](http://en.wikipedia.org/wiki/Null_Object_pattern).

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthenticator.java
```diff
@@ -10,7 +10,7 @@
   @Override
   public Optional<AuthenticatedUser> authenticate(APIGatewayCredentials credentials)
       throws AuthenticationException {
-    if (credentials.getUserId() > 0) {
+    if (credentials != null) {
```
drsnyder: Personally, I think it's better to *always* have an object here instead of having to check for null.

service/user-preference/src/main/java/com/wikia/userpreference/resources/UserPreferenceResource.java
```diff
@@ -11,21 +12,26 @@
...
 import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
-@Api(basePath = "/preferences", value = "User preferences",
+@Api(basePath = "/", value = "User preferences",
     produces = MediaType.APPLICATION_JSON, description = "User preference resources.")
-@Path("/preferences/{userId}")
+@Path("/{userId}")
```
drsnyder: :+1:

service/user-preference/src/test/java/com/wikia/userpreference/resources/UserPreferenceResourceIntegrationTest.java
drsnyder: They are a bit of a pain to maintain, but I would leave them. I would rather have the tests and know I can make changes to the resource with confidence without breaking anything.

service/user-preference/src/test/java/com/wikia/userpreference/resources/UserPreferenceResourceTest.java
```diff
@@ -9,104 +16,315 @@
...
+  public void getUserPreferencesForInvalidUserWithOtherUserAuthenticateHeader()
+      throws SQLException, IOException {
+    addHeaderToMockRequest(VALID_USER_ID_2);
+
+    Response response = getResponseForGetPreferences(INVALID_USER_ID);
+
+    assertEquals(HttpStatus.UNAUTHORIZED_401, response.getStatus());
+
+//    String expectedMessage = String.format("No preferences found for user %s", INVALID_USER_ID);
+//    verifyResponseProblemTitle(response, expectedMessage);
```
drsnyder: Can these comments be removed?

drsnyder:
Next time, please separate the functionality into 2 separate pull requests. I think the `AuthFactory` changes should be separate from the preference update addition.

drsnyder:
One part that is missing is the distinction between primary and secondary database connections (or master/slave). When we enable this in production  we want reads isolated on the secondaries and of course writes will have to go to the primary.

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
```diff
@@ -57,33 +79,27 @@ public void setRequest(HttpServletRequest request) {
...
+              .build();
+
+          final Optional<T> result = authenticator().authenticate(apiGatewayCredentials);
+          if (result.isPresent()) {
+            return result.get();
+          }
+        }
       } catch (AuthenticationException e) {
         LOGGER.warn("Error authenticating credentials", e);
+        throw new InternalServerErrorException();
```
drsnyder: From the dropwizard example, this part was puzzling to me. Why do we need to switch exception types? In practice, with our authentication flow, this should never happen. Maybe I shouldn't say never, but all we are currently doing is converting strings to integers.

gradle/pandora_dependencies.gradle
```diff
@@ -1,6 +1,6 @@
 apply from: rootProject.file('gradle/common_deps.gradle')
-def dropwizardVersion = "0.8.0"
-def jerseyVersion = "2.15"
+def dropwizardVersion = "0.8.1"
```
jogura: This is a general statement by bumping dropwizard up to `0.8.1` all services have will now have to ingest the change. What if this introduces incompatibilities with 3rd party modules like dropwizard-swagger, dropwizard-guice, etc. There should be a mechanism to allow each service to adopt the version of the core service at their own pace.

drsnyder: Good point @jogura. It seems like we are in lacking a convention or pattern with regard to our dropwizard version. I'm starting to wonder if at the module level this should be defined in one place (e.g. here) so that we don't have manually set it [here](https://github.com/Wikia/pandora/pull/171/files#diff-664f44886469406ca3c955eb73788dc2R3).

That conflicts with your suggestion though. If this is a per-service setting then the global libraries would probably have to be compatible with the lowest version dropwizard used by the services unless the versions were backwards compatible which they may not always be at the patch level.

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
sqreek: Yes, we can return here something, but something what?

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthenticator.java
sqreek: Heh, That was a tricky one. becuse there in our flow there is never a null in credentials (except tests). So I was thinking about removing this completly.
What's your opinion?

service/user-preference/src/test/java/com/wikia/userpreference/resources/UserPreferenceResourceTest.java
sqreek: Yes, or uncomment, forget to mentioned in PR. Do we want to check message response, or 401 is good enough. 
Becuse there is 2 exception throwing place. One in Lib/Auth, second in UserPreferences Resource, when userId from Path doesn't match userId from Auth Header.

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
sqreek: It's mostly for hide all possible exceptions information (stacktrace etc.) from exception that can be showed to user.

gradle/pandora_dependencies.gradle
sqreek: From my point of view, tests should show us if there is some backwards compatibilty problem.

Other thing, is when you want to use something new that was not supported in new version of dropwizard. 
In my opinion, dropwizard should be updated every new release, except when it have backwards compatible issue showed by tests

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
drsnyder: I think `ANONYMOUS_USER`(user id 0) is ok and a better option than `null`.

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthenticator.java
drsnyder: The majority of our traffic is missing credentials. That is, most of the requests that we receive now and will receive to our services in the future are from "anonymous" or logged out users.

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthFactory.java
pchojnacki: Dropwizard convention makes User null when its not set (if @Auth(required=false)

+ I think the best way to avoid null is by using Optional<> and not by adding null like state to the object itself

drsnyder: In my opinion, `AuthenticatedUser(0)` is the best option, `Optional` is passible but less desirable, and `null` is going to introduce a lot of conditional logic. Regarding conditional logic, `null` and `Optional` aren't that much different but the tooling is better in the case of `Optional`.

In the MediaWiki and Huddler worlds a user missing credentials is "anonymous" and has the user id 0. In the MediaWiki universe this translates into rows in the database associated with user id 0 (see the `archive` table in the muppet database as an example).

As a developer, what would you rather work with?

```java
if (authenticatedUser == null) {
   User loggedInUser = new User(Constants.ANONYMOUS);
   serviceLayer.createContribution(loggedInUser, data);
} else {
   serviceLayer.createContribution(loggedInUser, data);
}
```

Or `serviceLayer.createContribution(loggedInUser, data);`?

For my part, I would much rather work with the latter regardless of whether or not it's the dropwizard convention. In this layer, I think you should either get a valid object back or an exception but not null.

gradle/pandora_dependencies.gradle
pchojnacki: I agree with @sqreek if all tests pass after bumping the version then its good to go.
This is why we have one big repository so such changes are atomic. Allowing other services to keep different versions of some library is just asking for trouble and bunch of unmaintainable services in couple years time.

We can organize the meeting around it and/or setup a process to provide final decision but I strongly advise against allowing 'adopting modules at own pace'. I see some wiggle room that we could have two version tracks but allowing every service to have its own dropwizard version looks to me like just asking for trouble.

lib/dropwizard/auth/src/main/java/com/wikia/dropwizard/auth/APIGatewayAuthenticator.java
pchojnacki: New upcoming version of Dropwizard seems to fix that by allowign the use of Optional.
Anyway adding Anonymous user is against the spirit of how dropwizard handles users.

If you annotate your Resource method parameter with \@Auth then by default its code will only be run when \@Auth passes and object annotated with it is not null. This provides strong signal that all the code in service can run safely.

now when type in \@Auth(required=false) then the resource will be run even if auth fails.
But here I would strongly suggest not using AnonymousableUser instead something along the lines of Optional<User> would be much more prefered. Because by the very nature you can't do anything useful with AnonymizableUser apart from checking (obj->ifAnon() == true). Anonymous user doesn't have any id, you can't query for anything related to it there is no tracking etc.
In this case Optional.ofNullable(user) can be used before Dropwizard supports injecting optional user as it does with any other parameter.

pchojnacki: I just noticed that Optional.ofNullable(user) is even suggested by Dropwizard docs. http://www.dropwizard.io/manual/auth.html#protecting-resources

sqreek:
Offline meeting result:

distinction between primary and secondary database connections will be added in new PR.
We don't introduce Anonymous user, until we will have case for this scenario.
New Dropwizard version should be applied whenever possible if all tests are passing. Dev who wants to increase version have the responsibility to fix all version conflict.

PR #173, opened 2015-04-21
User's can only see their preferences

drsnyder:
https://wikia-inc.atlassian.net/browse/SERVICES-190

This PR integrates the dropwizard `AuthFactory` with the user preference service. If the authentication header is missing or if the user id does not match the owner of the preferences then the request is rejected.

There has been some discussion around using [shiro](http://shiro.apache.org/) for authentication. I think we should consider going that route if our requirements begin to extend beyond integer comparison.

/cc @jogura @sqreek


drsnyder:
It looks like the build failures are due to the dev consul crash.

jogura:
LGTM

sqreek:
Sorry for late comment.

I hijact code, and change it to more dropwizzard style.
I'm still working on it, and will create new PR for this stuff.

notes about work:
APIGatewayAuthFactory should handle required parameter from clone(boolean required), and throw new WebApplicationException when required is true.

Becuse of that, developer don't need 'double' check authorization in Resource method.

PR #174, opened 2015-04-22
Restores first level read of posts and threads

armonr:
WIP
ResponseGroup small (default) should work on reading post and thread objects.

/CC @jimmordino @jogura


armonr:
Test failures are unrelated to the changeset.

PR #176, opened 2015-04-27
Use standard response for unauthorized errors

drsnyder:
Use the `StandardResponses` for unauthorized errors.

/cc @sqreek


service/user-preference/src/main/java/com/wikia/userpreference/resources/UserPreferenceResource.java
```diff
@@ -93,7 +93,8 @@ public Response updateUserPreference(
 
   private void checkIsAuthUserIdIsSameAsUserId(Integer userId, AuthenticatedUser user) {
     if (!Objects.equals(userId, user.getUserId())) {
-      throw new WebApplicationException(Response.status(Response.Status.UNAUTHORIZED).build());
+      throw new WebApplicationException(StandardResponses.Unauthorized(
+          String.format("Unable to access user preferences belonging to %d.", userId)));
```
sqreek: very very minor.
Consider this:
Some random client -> call Service XXX -> call UserPreferences Service and try update some data, that client of XXX is not aware of.
random client get information about he tried to update some userId. 
Personally I prefer to give lack as possible information about authorization process.
In this case It seems legit

sqreek:
LGTM

service/user-preference/src/main/java/com/wikia/userpreference/resources/UserPreferenceResource.java
drsnyder: Yeah, this is something we have to be careful about. The only information they will get back in this case is the user id they requested which I think is ok.

drsnyder: Just to be safe, I removed it in c2853c4.

sqreek:
LE(ven)B(etter)TM

PR #177, opened 2015-04-28
Adds new fields for mobile app tracking

aniaMu:
Adds new fields for mobile app tracking, removes some fields that the mobile apps will not be sending to the warehouse


aniaMu:
@pchojnacki as agreed - please check the tests

PR #180, opened 2015-04-29
SERVICES-266 Filter applications by hub name

pchojnacki:
application list can now be filtered by appending ?hub=HubName


service/mobile-config/src/main/java/com/wikia/mobileconfig/service/application/AppsListService.java
```diff
@@ -1,17 +1,64 @@
...
-  public boolean isValidAppTag(String platform, String appTag) throws IOException;
+  boolean isValidAppTag(String platform, String appTag) throws IOException;
 
-  public List<HashMap<String, Object>> getAppList(String platform) throws IOException;
+  List<Map<String, Object>> getAppList(String platform) throws IOException;
 
-  public boolean isUp() throws IOException;
+  boolean isUp() throws IOException;
+
+  default Stream<Map<String, Object>> streamAppsList(String platform) throws IOException {
```
sqreek: Is this neccesserry?

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/application/AppsListService.java
```diff
@@ -1,17 +1,64 @@
...
+    return getAppList(platform).stream();
+  }
+
+  class FilterAppsByFieldPredicate implements Predicate<Map<String, Object>> {
```
sqreek: I'm disagree with keeping class with logic in interface file's, but it's only my opinion, and it can be ok.

service/mobile-config/src/test/java/com/wikia/mobileconfig/resources/ApplicationsResourceTest.java
```diff
@@ -95,6 +98,43 @@ public void testFilterTags() throws IOException {
...
+          .target("/applications/platform/test-platform?hub=cantbefound")
+          .request()
+          .get(String.class);
+    } finally {
+      reset(HTTP_CLIENT_MOCK);
+    }
+  }
+
+  @Test
+  public void testHubFilteringReturnsApp() throws IOException {
```
sqreek: Test should check something.
I saw that if apps.size() == 0 throw exception, but if someone change this behaviour, this test is meaningless.

sqreek:
:+1: for changing HashMap -> Map
except minor comment lgtm

service/mobile-config/src/test/java/com/wikia/mobileconfig/resources/ApplicationsResourceTest.java
pchojnacki: This test checks if Exception is thrown so it does check something :dancer:

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/application/AppsListService.java
pchojnacki: I don't have opinions about wheater its good or bad to add some generic code to interface. I just wanted to test out the concept. So far I like it

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/application/AppsListService.java
pchojnacki: It isn't but Defender methods are new thing in java 1.8 so i decided to start using them to see their up and downsides

service/mobile-config/src/main/java/com/wikia/mobileconfig/service/application/CephAppsListService.java
```diff
@@ -41,11 +42,11 @@ public boolean isValidAppTag(String platform, String appTag) throws IOException
...
-  public List<HashMap<String, Object>> getAppList(String platform) throws IOException {
+  public List<Map<String, Object>> getAppList(String platform) throws IOException {
     return getAppListFromCeph(platform);
   }
 
-  private List<HashMap<String, Object>> getAppListFromCeph(String platform) {
+  private List<Map<String, Object>> getAppListFromCeph(String platform) {
     try {
       String listResponse = this.cephGateway.getText(platform);
       return mapper.readValue(listResponse, new TypeReference<List<HashMap<String, Object>>>() {
```
ArturKlajnerok: @pchojnacki Is "TypeReference<List<HashMap<String, Object>>>" ok, since you changed  the return to "List<Map<String, Object>>" ?

pchojnacki: err I think yes since mapper.readValue will read out HashMap that then gets cast down to Map. Although for consistency sake it might be better to Readout Map instead of hashmap.

PR #181, opened 2015-04-29
DIS-15 move MyBatis DB mappers up to the DAO layer, remove former DAOs

jimmordino:
Use MyBatis DB Mappers at the DAO level instead of creating DAO implementations that just pass through to the MyBatis mappers.  Move any logic that was in the DAOs up into the services that call them.

@armonr, @jogura


service/discussion/src/main/java/com/wikia/discussionservice/dao/PostDao.java
```diff
@@ -2,17 +2,15 @@
...
 
-import java.util.Optional;
+import org.apache.ibatis.annotations.Param;
 
 public interface PostDao {
 
-  Optional<Post> createPost(Integer siteId, Long threadId, Post post);
+  Post getPost(@Param("siteId") Integer siteId, @Param("postId") Long postId);
 
-  Optional<Post> getPost(Integer siteId, Long postId);
```
armonr: Why not keep `Optional`?

jimmordino: The implementation for DAOs is supplied by MyBatis - my assumption is that MyBatis mappers can't return Optional.

jimmordino: It looks like we might be able to create a TypeHandler for Optional - I'll give this a look.

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
```diff
@@ -21,15 +21,15 @@ public PostService(PostDao postDao) {
...
 
-  public Optional<Post> createPost(int siteId, long threadId, Post post) {
-    return postDao.createPost(siteId, threadId, post);
+  public Optional<Post> createPost(Post post) {
+    return Optional.ofNullable(postDao.createPost(post));
   }
 
-  public Optional<Post> deletePost(int siteId, long postId) {
+  public Boolean deletePost(Integer siteId, Long postId) {
     return postDao.deletePost(siteId, postId);
```
armonr: ` == 1`?

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
```diff
@@ -100,8 +59,18 @@ public ForumService(ThreadService threadService, ForumDao forumDao) {
       rethrowExceptionsAs = DiscussionServiceException.class,
       exceptionMessage = "Unable to create forum with site ID %s"
       )
-  public Optional<Forum> createForum(Integer siteId, Forum forum) {
-    return forumDao.createForum(forum);
+  public Optional<Forum> createForum(Forum forum) {
+    Optional<Forum> parentForum = getForum(forum.getSiteId(), forum.getParentId(), 10, 0);
+
+    if (!parentForum.isPresent()) {
```
armonr: Consider using `.map()`

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
```diff
@@ -13,29 +14,52 @@
...
-    return threadDao.createThread(siteId, ContentType.FORUM_THREAD, forumId, post);
+  public <T extends DiscussionThread> Optional<T> createThread(Long forumId, Post post) {
+    
+    // Create and save the thread with partial information
+    ForumThread forumThread = new ForumThread();
+    forumThread.setSiteId(post.getSiteId());
+    forumThread.setForumId(forumId);
+    forumThread.setTitle(post.getTitle());
+    forumThread.setThreadStarterId(post.getCreatorId());
+    threadDao.createThread(forumThread);
```
armonr: Should there be a status check?

service/discussion/src/main/java/com/wikia/discussionservice/dao/ForumDao.java
```diff
@@ -2,19 +2,20 @@
...
 public interface ForumDao {
 
-  List<Forum> getForumList(Integer siteId);
-
-  Optional<Forum> getForum(Integer siteId, Long forumId, Integer limit, Integer offset);
+  public Forum getForum(@Param("siteId") Integer siteId, @Param("forumId") Long forumId,
+                  @Param("limit")Integer limit, @Param("offset")Integer offset);
 
-  Optional<Forum> createForum(Forum forum);
+  public List<Forum> getForumList(Integer siteId);
```
armonr: `public` not required.

jogura:
Other than the minor things that have been pointed out - LGTM

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
jimmordino: e51b5fc73b60834e123bf8cb45f91976556b0c50

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
jimmordino: Fixed in e51b5fc73b60834e123bf8cb45f91976556b0c50

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
jimmordino: Yes, this code should be refined by whomever is implementing the thread resource behavior.

armonr:
:ship: it

service/discussion/src/main/java/com/wikia/discussionservice/dao/PostDao.java
jimmordino: MyBatis has a couple of mechanisms for custom type handling, TypeHandler and ObjectFactory.  Neither of these really allows us to return Optional because they only come into play when there are results to process.  We'd need to override higher level framework behavior to create and return ```Optional.empty()``` when there are no results for a given query.  I don't think this is worth the effort.  We can write the services to wrap the results of DAO calls in an Optional if that's what we need.

armonr: Totally reasonable. Thanks for looking into it.

PR #184, opened 2015-04-30
Cleanup dependencies and buildscripts + random port support

pchojnacki:
Tests were often failing because Integration tests were run on port 8080
This change adds helpers that set random ports for integration tests


nmonterroso:
LGTM. looks like tests failed because of a failure to connect to archiva. re-running.

PR #185, opened 2015-04-30
Multiple database support

nmonterroso:
@Wikia/services-team @jogura @jimmordino 
https://wikia-inc.atlassian.net/browse/SERVICES-218

Adds support for multiple database connections in Dropwizard using jOOQ. Integration should be pretty simple - a class, most likely a service's `Configuration`, needs to implement the `DatabaseProvider` interface. ex:

```java
Class MyConfiguration extends Configuration implements DatabaseProvider {
```

`DatabaseProvider` implements one method:

```java
DatabaseConfigs getDatabases();
```

`DatabaseConfigs` can be defined in your `yaml` config:

```yaml
databases:
  configurations:
    wikicities:
      master:
        driverClass: com.mysql.jdbc.Driver
        url: ${consul:DATABASE_MASTER_URL}
        user: ${consul:DATABASE_USER:root}
        password: ${consul:DATABASE_PASSWORD}
        validationQuery: select * from user_properties limit 1
      slave:
        driverClass: com.mysql.jdbc.Driver
        url: ${consul:DATABASE_SLAVE_URL}
        user: ${consul:DATABASE_USER:root}
        password: ${consul:DATABASE_PASSWORD}
        validationQuery: select * from user_properties limit 1
```

`master` and `slave` are hard-coded for read/write and read-only access, but `wikicities` in the example above can be any string, and `configurations` supports any number of database names. `master` and `slave` are [`DataSourceFactory`](http://www.dropwizard.io/0.8.1/dropwizard-db/apidocs/io/dropwizard/db/DataSourceFactory.html) configurations. Once this is set up, `DatabaseConfigs` can be injected, ex: 
```java
@Inject
public MyClass(DatabaseConfigs dbs) {
```
and [`DSLContext`](http://www.jooq.org/doc/3.6/manual/sql-building/dsl-context/) for a database can be acquired using `dslFor`, ex:

```java
DSLContext context = dbs.dslFor("wikicities"); // defaults to use SQLDialect.MYSQL, can be overwritten using DatabaseConfigs.setDefaultDialect(SQLDialect dialect)
// or, for postgres sql dialect
DSLContext context = dbs.dslFor("wikicities", SQLDialect.POSTGRES);
```

Alternatively, if you wish to modify the configuration before instantiating the `DSLContext`, you can use `configFor(String name)` or `configFor(String name, SQLDialect dialect)`:

```java
Configuration config = dbs.configFor("wikicities");
// do other stuff to config
DSLContext context = DSL.using(config);
```

Queries will automatically switch to the `master` configuration when a `write` operation is attempted. An exception is thrown if there is no master configured. This is done by `QueryExecuteListener`.

Automatic health checks are also installed into the application.


lib/dropwizard/consul/build.gradle
```diff
@@ -1,6 +1,8 @@
 apply from: rootProject.file('gradle/common_deps.gradle')
 dependencies {
     compile project(':lib:dropwizard:base')
+    // TODO: de-couple guice and consul, it's weird to have consul depend on the database
```
nmonterroso: This `TODO` is tracked in https://wikia-inc.atlassian.net/browse/SERVICES-300

pchojnacki: I haven't yet looked at the code but FWIW lib:dropwizard:base should be just bare minimum dropwizard dependencies and serve as a base for all other projects adding to dropwizard.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/ConsulGuiceApplication.java
```diff
@@ -75,12 +81,21 @@ public void initialize(Bootstrap<T> bootstrap) {
         .addAll(getAdditionalModules())
         .build();
 
+    @SuppressWarnings("unchecked")
+    Class<? extends FeatureInstaller>[] featureInstallers = ImmutableList.<Class<? extends FeatureInstaller>> builder()
+        .add(ConsulRegistrationInstaller.class)
+        .add(DatabaseInstaller.class)
```
pchojnacki: shouldn't DatabaseInstaller be hadndled through getAdditionalFeatureInstaller? That is what if we have an application that does not require any DB access ?

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseInstaller.java
```diff
@@ -0,0 +1,50 @@
...
+  public void install(Environment environment, DatabaseProvider instance) {
+    instance.getDatabases().getConfigurations().forEach((name, dbSource) -> {
+      dbSource.ensureDataSources(environment.metrics(), name);
+
+      try {
+        DatabaseHealthCheck masterCheck = new DatabaseHealthCheck(
+            dbSource.getMasterSource(),
+            dbSource.getMasterFactory().getValidationQuery());
+        environment.healthChecks().register(name+"-db-master", masterCheck);
+      } catch (NullPointerException npe) {}
```
pchojnacki: I think you should log the npe instead of silently ignoring it.

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseInstaller.java
```diff
@@ -0,0 +1,50 @@
...
+
+      try {
+        DatabaseHealthCheck slaveCheck = new DatabaseHealthCheck(
+            dbSource.getSlaveSource(),
+            dbSource.getSlaveFactory().getValidationQuery());
+        environment.healthChecks().register(name+"-db-slave", slaveCheck);
+      } catch (NullPointerException npe) {}
```
pchojnacki: +

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseModule.java
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  @Override
+  protected void configure() {
+    bind(DatabaseProvider.class).to(databaseProvider);
+  }
+
+  @Provides
+  @Singleton
+  public DatabaseConfigs databaseSourceConfigurations(DatabaseProvider provider) {
```
pchojnacki: Why this is a Singleton?

lib/dropwizard/databases/build.gradle
```diff
@@ -0,0 +1,6 @@
+apply from: rootProject.file('gradle/common_deps.gradle')
+dependencies {
+    compile project(':lib:dropwizard:base')
+    compile project(':lib:jooq:jooq-base')
+    compile "io.dropwizard:dropwizard-db:0.8.1"
```
sqreek: We need to figure out where we want to handle current version of Dropwizard.
This add another point of future version bump.

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseConfigs.java
```diff
@@ -0,0 +1,70 @@
...
+  public DatabaseConfigs() {
+    configurations = new HashMap<>();
+  }
+
+  public HashMap<String, DatabaseSource> getConfigurations() {
+    return configurations;
+  }
+
+  public void setConfigurations(
+      HashMap<String, DatabaseSource> configurations) {
```
sqreek: minor (formatting)

sqreek: Also Its should be Map instead of HashMap

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseConfigs.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jooq.impl.DefaultConfiguration;
+import org.jooq.impl.DefaultExecuteListenerProvider;
+
+import java.util.HashMap;
+
+public class DatabaseConfigs {
+  private static SQLDialect defaultDialect = SQLDialect.MYSQL;
+
+  @JsonProperty
+  private HashMap<String, DatabaseSource> configurations;
```
sqreek: use Map instead of HashMap

lib/dropwizard/databases/src/test/java/com/wikia/dropwizard/databases/DatabaseConfigsTest.java
```diff
@@ -0,0 +1,54 @@
...
+  private DatabaseSource dbSource;
+  private SQLDialect defaultDialect;
+
+  @Before
+  public void setup() {
+    defaultDialect = DatabaseConfigs.getDefaultDialect();
+    configs = new DatabaseConfigs();
+    dbSource = Mockito.mock(DatabaseSource.class);
+
+    HashMap<String, DatabaseSource> sources = new HashMap<>();
```
sqreek: should be map instead of HashMap

lib/dropwizard/databases/build.gradle
pchojnacki: if you include from common_deps then globlal variable $dropwizardVersion is set so you can use ```compile "io.dropwizard:dropwizard-db:$dropwizardVersion"```

pchojnacki: But this is only now available in the master

pchojnacki:
Few minor comments otherwise looks good.

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseHealthCheck.java
```diff
@@ -0,0 +1,33 @@
...
+  private String validationQuery;
+
+  public DatabaseHealthCheck(DataSource dataSource, String validationQuery) {
+    this.dataSource = dataSource;
+    this.validationQuery = validationQuery;
+  }
+
+  @Override
+  protected Result check() throws Exception {
+    try (Connection conn = dataSource.getConnection()) {
```
sqreek: It will be good to ensure 2 things about this code:
1) is it reuse connections? (it should be, but if not, and always new connection will be created it will be pain)
2) if 1) works as expected (using connection pool) we need to check is nothing hold this connection, and it is returned into connection pool. Otherwise we get connections leak.

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseConfigs.java
```diff
@@ -0,0 +1,70 @@
+package com.wikia.dropwizard.databases;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.jooq.Configuration;
+import org.jooq.DSLContext;
+import org.jooq.SQLDialect;
+import org.jooq.impl.DSL;
+import org.jooq.impl.DefaultConfiguration;
+import org.jooq.impl.DefaultExecuteListenerProvider;
+
```
jimmordino: This implementation includes several objects that have Jooq dependencies.  If we're going to commit to Jooq as our access layer then let's make that decision and move forward.  Otherwise, if we want to allow for other options then would you consider adding generic interfaces for the Jooq-specific implementations to support that flexibility?

pchojnacki: Hey John. I don't think any decision is necessary and if you want to use other framework with this functionality then you are more than welcome to submit patches to our common code. AFAIK this functionality is using some things that are specific to Jooq and for one I'm not sure if its adaptable to mybatis.

But your comment makes it clear that this package should be named com.wikia.dropwizard.jooq instead of (...)database.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/ConsulGuiceApplication.java
nmonterroso: Yeah, that's how I had it originally, though in that case I think there would be a lot of duplication going on for applications adding the database installer. The way it's currently implemented, applications that don't require DB access don't pass the `matches` function in `DatabaseInstaller`, so nothing further down gets used/installed (guice module, health checks, etc).

Making it be required by `getAdditionalFeatureInstaller` is a valid option, though. It's just a matter of how magical we want this to be. I'm OK either way. Though I still think SERVICES-300 is valid, because I imagine there will be something in the future we'll want to add to the list of "always-on" installers.

lib/dropwizard/databases/build.gradle
nmonterroso: Yah, I saw that in the dependency cleanup PR. I'll merge in master and update this. Thanks!

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseConfigs.java
nmonterroso: Good point about package names @pchojnacki, I'll update. 

@jimmordino I was thinking about how to make this using generic interfaces, but since the things we're interfacing (Jooq and Mybatis) are out of our control, I wonder how we could do that? I'm definitely open to it, but it would require some thought.

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseModule.java
nmonterroso: Hm, I previously had code in here that was instantiating the connection pools, so I didn't want to have multiple of those flying around. I've moved that to the Installer, so maybe I can probably remove the `@Singleton`. Good catch.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/ConsulGuiceApplication.java
nmonterroso: changed to use `getAdditionalFeatureInstaller` in cd2c80f

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseInstaller.java
nmonterroso: logging npes in c5b3ce7

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseModule.java
nmonterroso: `@Singleton` removed in 68ef686

lib/dropwizard/databases/build.gradle
nmonterroso: 566dddc

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseHealthCheck.java
nmonterroso: Good points, thanks for calling these out. I checked the code and observed through the debugger and #1 is true - the `DataSource` is a `ManagedPooledDataSource` which manages a connection pool. #2 also works as expected (from observation) - I see the pool with busy connections when a query is running, and they return to idle when completed.

lib/dropwizard/databases/src/main/java/com/wikia/dropwizard/databases/DatabaseConfigs.java
nmonterroso: I'll open a separate PR for making the DB stuff a bit more generic

PR #189, opened 2015-05-04
First attempt of adding Authz into Discussion Service

jogura:
This PR attempts to encapsulate Authz into functional interfaces. The idea is to avoid sprinkling logic that checks user's roles all over the code base. Instead, the activity (`SecureActivity`) is used to authorize the current user. This subtle, change will allow for business changes to only occur in place in the code.

The PostResource/PostService were also refactored slightly to use the `com.wikia.discussionservice.auth.User` instead of the domain user. @armonr please review those changes.

Unit tests were added for the Authz code and integration tests were added for ThreadResource.

RFCR @jimmordino @armonr and anyone else


jogura:
The unittest failure is because DIS-15 branch cannot be cleanly merged to master without manually resolving conflicts.

service/discussion/src/main/java/com/wikia/discussionservice/resources/ThreadResource.java
```diff
@@ -206,4 +204,41 @@ public Response deleteThread(@Auth User user,
...
+                               LongParam threadId,
+                               @Valid @SiteIdValidator DiscussionThreadInput<ForumThread> threadInput,
+                               @Context UriInfo uriInfo) {
+    DiscussionThread toBeUpdatedThread = threadInput.getdiscussionThread();
+    Optional<DiscussionThread> forumThread = threadService.updateThread(user, toBeUpdatedThread);
+    return forumThread.map(updatedThread -> {
+      if (updatedThread.getId() == toBeUpdatedThread.getId()) {
+        return Response.noContent().build();
+      } else {
+        return Response.status(Response.Status.CREATED)
```
jimmordino: DIS cannot create a new post on a PUT if update fails.  We simply don't have the information we need, specifically a post.  I would rather have the code be specific about never creating on PUT instead of adding code to handle a case that can't happen.

jogura: Yes, we did talk about this. I was following the standards outlined in the our guidelines but for discussion service that behavior doesn't make sense. I'll change the behavior to throw an error if an update fails.

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
```diff
@@ -59,7 +74,47 @@ public ThreadService(DiscussionThreadDao threadDao, PostService postService) {
...
   public Boolean deleteThread(Integer siteId, Long threadId) {
-    return Boolean.valueOf(threadDao.deleteThread(siteId, threadId) == 1);
+    // TODO: remove and implement for real
+    throw new NotFoundException();
+//    return Boolean.valueOf(threadDao.deleteThread(siteId, threadId) == 1);
   }
 
+  /**
+   * Attempt to update a DiscussionThread this is a full replacement of the existing thread (ignores
+   * posts). If the thread does not exist an attempt to create a new one is made. Authorization will
```
jimmordino: Trivial: Stale comment

service/discussion/src/main/resources/com/wikia/discussionservice/dao/DiscussionThreadDao.xml
```diff
@@ -48,6 +48,22 @@
...
+    <insert id="createThread" parameterType="com.wikia.discussionservice.domain.ForumThread"
+            keyProperty="thread.id">
+        INSERT INTO thread (
+        id, site_id, title, created_by, container_type, container_id, post_count, first_post_id,
+        last_post_id
+        )
+        VALUES (
+        0, <!-- we specify zero because column is non-null.  sproc will replace with GUID -->
+        #{siteId}, #{title}, #{creatorId}, #{type}, #{containerId}, 0,
+        0, 0) <!-- we specify zero because column is non-null; subsequent update will replace with GUID) -->
```
jimmordino: Trivial: copypasta comment - should state that post IDs are updated after post has been created.

jimmordino: Trivial: might want to state that these are post IDs and will be updated after post has been created.

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceRealm.java
```diff
@@ -86,6 +86,10 @@ public boolean supports(AuthenticationToken token) {
       case("2"):
         userInfo.put("roles", Sets.newHashSet(Role.USER.toString()));
         break;
+      case("3"):
```
armonr: Can these be constant-ized?

service/discussion/src/main/java/com/wikia/discussionservice/resources/PostResource.java
```diff
@@ -182,10 +171,7 @@ public Response updatePost(
     post.setId(postId.get());
 
     // Todo: User auth user instead
```
armonr: Please remove this comment also

service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
```diff
@@ -0,0 +1,36 @@
...
+    if (user.isStaff() || user.isAdmin() || user.isModerator() ||
+        (user.isLoggedInUser() && threadService.isThreadOwner(user.getUserId(), threadId))) { // NOPMD
+      return true;
+    }
+    throw new UnauthorizedException("This user is not authorized to update the thread.");
+  };
+
+  public final static Authorizable authorizeCreateThread = (user, forumId) -> {
+    // TODO account for the possibility that the user doesn't have perms to View/Edit the forum
+    if (user.isStaff() || user.isAdmin() || user.isModerator() || user.isLoggedInUser()) {
```
armonr: It might be worth factoring out `user.isStaff() || user.isAdmin() || user.isModerator()`

service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
```diff
@@ -0,0 +1,36 @@
...
+      return true;
+    }
+    throw new UnauthorizedException("This user is not authorized to update the thread.");
```
armonr: /s/update/create

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
```diff
@@ -59,7 +74,49 @@ public ThreadService(DiscussionThreadDao threadDao, PostService postService) {
   }
 
   public Boolean deleteThread(Integer siteId, Long threadId) {
-    return Boolean.valueOf(threadDao.deleteThread(siteId, threadId) == 1);
+    // TODO: remove and implement for real
+    throw new NotFoundException();
```
armonr: `NotImplementedException`?

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
```diff
@@ -59,7 +74,49 @@ public ThreadService(DiscussionThreadDao threadDao, PostService postService) {
...
+
+    throw new NotFoundException(String.format(
+        "Thread with id: %s does not exist.", thread.getId()));
+  }
+
+  public boolean isExistingThread(Long threadId) {
+    return threadDao.isExistingThread(threadId);
+  }
+
+  public boolean isThreadOwner(Integer userId, Long contentId) {
```
armonr: What's the purpose of wrapping this?

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
```diff
@@ -59,7 +74,49 @@ public ThreadService(DiscussionThreadDao threadDao, PostService postService) {
...
+        return this.getForumThread(user, thread.getSiteId(), thread.getId());
+      }
+      return Optional.empty();
+    }
+
+    throw new NotFoundException(String.format(
+        "Thread with id: %s does not exist.", thread.getId()));
+  }
+
+  public boolean isExistingThread(Long threadId) {
```
armonr: Same here, why not call DAO method directly?

lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceRealm.java
jogura: It can but I don't think it is worth the effort especially since it is mocked.

service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
jogura: What did you have in mind?

service/discussion/src/test/resources/fixtures/updateThread.json
```diff
@@ -0,0 +1,7 @@
+{
+  "id": 1,
+  "title": "Updated Thread Title",
+  "forumId": 1,
+  "siteId": 1,
+  "creatorId": 1
+}
```
armonr: Should this have all non-NULLable field, or else, it would be partial.

service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
armonr: Maybe we can have a function called `isPrivileged`

service/discussion/src/main/java/com/wikia/discussionservice/services/ForumService.java
```diff
@@ -41,7 +41,6 @@ public ForumService(ForumDao forumDao) {
   }
 
   public Optional<Forum> getForums(Integer siteId) {
-    // FIXME
```
jimmordino: OK, but it still needs to be fixed...:smile:

service/discussion/src/test/resources/fixtures/updateThread.json
jimmordino: To be more precise, all writable fields.  For example, create_date shouldn't be nullable, but isn't writable by the application, either.

service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
jogura: The requirements modify the definition of `isPrivileged ` depending on the activity. Hence, why I went for the more verbose yet more flexible approach. I would like to avoid causing security breaches because we modify a common `isPrivileged` method without understanding the ramifications.

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
jimmordino: It's temporary, so it's not critical what's thrown.

service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
armonr: Sure, we can defer refactoring this once we find ourselves repeating these three over and over, 
and there maybe a few more patterns that emerge as we add more authz functions here.

service/discussion/src/test/resources/fixtures/updateThread.json
jimmordino: ```creatorId``` should not be part of the update SQL, by the way.   of the existing columns, we should really only update ```title, container_id, viewable_ind, editable_ind, and held_ind``` in the SQL.  I don't think clients should be able to update other fields directly.

In fact, this brings up the question of checking for privilege before updating status fields - we'll need to figure out which fields can be updated by a logged in user and which fields are mod-only.  Perhaps mod-only deserve a special resource (e.g. hold/unhold, lock/unlock, etc.)

jimmordino:
LGTM, but you need an <update id="updateThread"> element in  your mapper XML ASAP.

PR #190, opened 2015-05-05
Multiple generic dbs

nmonterroso:
@Wikia/services-team @jimmordino @jogura @armonr 
Separates core multi-db classes from jooq-specific implementations. Without knowing the details of MyBatis, if it sources from `DataSourceFactory` and `DataSource` objects then this will hopefully be usable. To add support for a new database library, a few things must be done:

1. create new project in `lib/dropwizard/databases/db-*`
2. extend `AbstractDatabaseConfigs`. This class will be responsible for using database connections to create library-specific database objects. For example, the `JooqConfigs` class creates `DSLContext` objects used to query databases using the Jooq DSL.
3. extend `AbstractDatabaseModule`. This class will configure guice to inject the `AbstractDatabaseConfigs` implemented above.
4. extend `AbstractDatabaseInstaller`. There shouldn't be anything other than implementing `AbstractDatabaseModule getDatabaseModule();`, which should return an instance of your `AbstractDatabaseModule`.

When those are done, an app should add 
```java
  @Override
  public List<Class<? extends FeatureInstaller>> getAdditionalFeatureInstallers() {
    return ImmutableList.<Class<? extends FeatureInstaller>>builder()
        .add(MyDatabaseInstaller.class)
        .build();
  }
```

to their `ConsulGuiceApplication` class. The `Jooq` implementations of the `Abstract*` classes should hopefully provide a good example.


jimmordino:
:+1:  Once you merge I'll get to work on db-mybatis.

lib/dropwizard/databases/db-base/src/main/java/com/wikia/dropwizard/databases/AbstractDatabaseInstaller.java
```diff
@@ -32,22 +35,25 @@ public void install(Environment environment, DatabaseProvider instance) {
...
+    return FeatureUtils.is(type, DatabaseProvider.class);
   }
 
   @Override
-  public boolean matches(Class<?> type) {
-    return FeatureUtils.is(type, DatabaseProvider.class);
+  @SuppressWarnings("unchecked")
+  public <T> void install(Binder binder, Class<? extends T> type, boolean lazy) {
+    AbstractDatabaseModule module = getDatabaseModule();
+    module.setDatabaseProvider((Class<? extends DatabaseProvider>) type);
```
drsnyder: Is the casting necessary here? Is there a way to do this that doesn't require a cast?

nmonterroso: I don't think so. The class is guaranteed to extend `DatabaseProvider` because of the implementation of `matches()`, but in order to correctly override `BindingInstaller.install`, we can't change the signature of this method :(

drsnyder:
:+1:

PR #193, opened 2015-05-08
[DIS-42] Refine getThread() with all valid values of response group implementation

jogura:
* Flesh out both Thread and Post HalMappers to return properties and nested elements
* Update MyBatis mappers for Thread and Posts to return required data to populate the
domain objects
* Modify domain objects to conform to layout
* Add/fix additional unit tests for ThreadResourcesAuthz and HalThreadRepresentationMapper
* Add suppressfbwarnings and other fix-ups for findbugs/pmd

@jimmordino @armonr


service/discussion/src/main/java/com/wikia/discussionservice/domain/PostRevision.java
```diff
@@ -27,9 +31,6 @@
   private Integer creatorId;
 
   @JsonProperty
-  private String title;
```
jimmordino: Why are we deleting this?

service/discussion/src/main/java/com/wikia/discussionservice/resources/ThreadResource.java
```diff
@@ -104,7 +104,7 @@ public Response getThread(
       @ApiParam(value = "The pagination position")
       @QueryParam("offset") @DefaultValue("1") IntParam offset,
       @ApiParam(value = "The responseGroup controls the level of details returned with this call",
-          allowableValues = "small,large")
+          allowableValues = "small,large,full")
```
jimmordino: Do we have use cases for these?  Can we name the response groups after use cases if possible?

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
```diff
@@ -129,7 +129,6 @@ public Boolean unlockPost(Integer siteId, Long postId) {
   private PostRevision buildNewPostRevision(Post post, User creator) {
     return PostRevision.builder()
         .postId(post.getId())
-        .title(post.getTitle())
```
jimmordino: I think we need to keep this until we're certain that there won't be titles.

service/discussion/src/main/java/com/wikia/discussionservice/domain/PostRevision.java
jogura: I don't recall why I thought it was a good idea. But it's been reverted: c0251c5

service/discussion/src/main/java/com/wikia/discussionservice/domain/Post.java
```diff
@@ -74,9 +74,10 @@
   @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS,
       include = JsonTypeInfo.As.PROPERTY,
       property = "@class")
-  @NonNull
```
armonr: We still need this

service/discussion/src/main/java/com/wikia/discussionservice/resources/ThreadResource.java
armonr: I created [DIS-71](https://wikia-inc.atlassian.net/browse/DIS-71) to address that

jogura: :+1:

service/discussion/src/main/java/com/wikia/discussionservice/domain/Post.java
jogura: Please explain, my thought is in the case where we are simply getting the latest revision we don't need the list of revisions.

jimmordino: It wouldn't be unreasonable to initialize this to an empty list and keep it as non-null.  On the other hand, there are nullable fields where the getter will return null, so that's a case that should be handled.

armonr: I'm not sure if this is the best way to go about it because this means that our domain objects directly depend on the parameters on API requests. Sure, we want to return a smaller response faster, but IMO differentiating at DAO level is a premature optimization. Also if we were to have a thin cache layer at this level, multiple copies of the same object has to be cached.
Shouldn't the representation level handle this differentiation?

jimmordino: I understand what you are saying, but I don't agree that fetching only what you need from the database is a premature optimization.  We don't want fetch 100 sets of post revisions to display a thread page on desktop.  Nor do we want to fetch post revisions when the user requests a forum list, etc.  We don't paginate by fetching the entire list and picking the right set in memory, right?  Let's fetch only what we need from the DB.

service/discussion/src/main/resources/com/wikia/discussionservice/dao/DiscussionThreadDao.xml
```diff
@@ -75,17 +80,79 @@
...
         FROM thread
-        JOIN post ON (
-          thread.id = post.thread_id
-        )
+            JOIN post ON (thread.id = post.thread_id)
+            JOIN post first_post ON (thread.first_post_id = first_post.id)
+            JOIN post_revision AS first_post_revision ON (first_post_revision.post_id = first_post.id)
+            JOIN post last_post ON (thread.last_post_id = last_post.id)
+            JOIN post_revision AS last_post_revision ON (last_post_revision.post_id = last_post.id)
+            JOIN post_revision AS rev ON (rev.post_id = post.id)
```
armonr: Isn't this the same as last_post?

jimmordino: That depends on how we implement "revert post to previous version", doesn't it?  :smile:

armonr: And that is TBD. Sure, let's defer this.

armonr:
LGTM

jimmordino:
:+1:

PR #195, opened 2015-05-08
Create New Service Fixes/Improvements

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-294

Updating new service creation to create a runnable service that compiles. Other changes:

* consolidate new service creation gradle script, templates, and tests in `tools/create-new-service`. It's now run using `gradlew :tools:create-new-service:run`
* add test task `:tools:create-new-service:test` that creates a temporary service using `:tools:create-new-service:run` and runs the unit tests
* remove `ModuleFactory` from services created using the script. AFAIK only `mobile-config` has multiple guice `AbstractModule` classes.
* added integration test to created service


tools/create-new-service/build.gradle
```diff
@@ -1,10 +1,28 @@
+apply from: rootProject.file('gradle/common_deps.gradle')
+
 import org.apache.tools.ant.filters.ReplaceTokens
 
+def sanitizeProjectName(String input) {
+    return input.replaceAll(/\s|_/, "-").toLowerCase().replaceAll(/[^a-z0-9-]/, "")
+}
+
+def appendNewServiceToSettings(String projectName) {
```
nmonterroso: Note: This was changed to output the whole file rather than to simply append to `settings.gradle`. The old method would have created a file like so:

```groovy

include 'service:service1'

include 'service:service2'

include 'service:service3'

```

But since this removes all trailing empty lines, the resulting file will looks like:

```groovy
include 'service:service1'
include 'service:service2'
include 'service:service3'
```

tools/create-new-service/templates/src/main/java/com/wikia/services/template/TemplateApplication.java
```diff
@@ -0,0 +1,36 @@
...
+  public static final RepresentationFactory
+      REPRESENTATION_FACTORY = new StandardRepresentationFactory();
+
+  public static void main(String[] args) throws Exception {
+    new @templatePascal@Application().run(args);
+  }
+
+  @Override
+  public String getName() {
+    return "@projectName@"; // no uppercase letters
```
drsnyder: We can probably take the comment out.

drsnyder:
:+1:

tools/create-new-service/templates/src/main/java/com/wikia/services/template/TemplateApplication.java
nmonterroso: removed in 665b9ab

PR #198, opened 2015-05-18
removing notifications service

sqreek:
becuse notifications service is not used on production and its not under develop,
we set tripwire until 8/18/2014 with removing this code from repo

//cc @Wikia/services-team


pchojnacki:
LGTM
/cc @kvas-damian @rafalkalinski

PR #199, opened 2015-05-18
[DIS] DIS-40 Implement view forums behavior

jimmordino:
DIS-40 - As a logged out user (and above), I want to see all visible forums

Implement forum view behavior and supporting DIS and test code.

@jogura @armonr


armonr:
LGTM

PR #202, opened 2015-05-19
DIS-41 remove content type codes - use labels instead

jimmordino:
Having the codes is confusing the issue.  We should remove them unless and until we have a need for them.

@armonr @jogura


jogura:
LGTM

PR #203, opened 2015-05-19
Adjust evictionInterval and minIdleTime for dis

armonr:
This prevents constant eviction of idle connections from the pool, preventing the frequent 5XX responses as a result.

@jimmordino @jogura


jogura:
LGTM, we may need to revisit/tune before we go into production.

PR #208, opened 2015-05-20
Discussion Service - Forum and Thread list with tests

armonr:
* DIS-41 Fetching ThreadList for a Forum
 * Refactor common SQL snippets into reusable MyBatis statements
 * Refactor ForumResource to use new method to get a Forum with a list of threads

* Adjust evictionInterval and minIdleTime
* DIS-41 remove content type codes - use labels instead

* DIS-81 Tests for DIS-41 Display ThreadList
 * Add additional tests for getting specific forums with threads
 * Add methods to convert HalResponses for Threads, Posts, and PostRevisions
 * Add SecureActivity for VIEW_FORUM (stubbed impl ATM)
 * Refactor ForumResourceTest to use Hamcrest assertThat
 * Introduce new generic Hamcrest Matcher for Content comparisons by id from a list
 * Refactor all Mappers to not stringify Integer types (we are only stringifying Long)


jimmordino:
:+1:

PR #209, opened 2015-05-20
Refactor post creation + few post tests

armonr:
Adds a couple of resource and service level tests

/CC @jogura @jimmordino


service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
```diff
@@ -17,46 +17,57 @@
...
+    return createPostRevision(user, post)
+        .map(postRevision -> {
+          createPostRecord(post, postRevision)
+              .orElseThrow(() -> new DiscussionServiceException());
 
-    Post newPost = getPost(post.getSiteId(), post.getId()).map(
-        post1 -> post1
-    ).orElseThrow(() -> new DiscussionServiceException("Post could not be found"));
+          updatePostRevisionId(post, postRevision)
+              .orElseThrow(() -> new DiscussionServiceException());
```
jimmordino: We need to find a way to integration test this - if the update fails here we want to be sure it all gets rolled back, i.e. transactions are working as we expect.

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
```diff
@@ -135,4 +146,75 @@ private PostRevision buildNewPostRevision(Post post, User creator) {
...
+  /**
+   * Same as #getPost, but throws an exception if not found
+   *
+   * @param siteId Integer
+   * @param postId Long
+   * @return Optional[Post]
+   *
+   * @throws DiscussionServiceException if post was not found
+   */
+  private Optional<Post> getPostOrException(Integer siteId, Long postId)
```
jimmordino: Don't like the name - just get post and have it throw if not found.  Let's ban "OrException" from our naming standards

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
armonr: Definitely, a must-have for 'write' tests.

service/discussion/src/test/java/com/wikia/discussionservice/resources/PostResourceTest.java
```diff
@@ -0,0 +1,127 @@
...
+  @Test
+  public void getPost() {
+    Post post;
+    try {
+      post = initPostFromFixture(FIXTURE_CREATE);
+    } catch (IOException e) {
+      e.printStackTrace();
+      fail(String.format("Failed to populate post object: %s", e.getMessage()));
+      return;
+    }
```
jogura: This try-block can be a candidate for refactoring into the `setup()` method.

service/discussion/src/test/java/com/wikia/discussionservice/resources/PostResourceTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+    Post actualPost = new Post();
+    try {
+      BeanUtils.populate(actualPost, (HashMap) response.get("properties"));
+    } catch (IllegalAccessException e) {
+      e.printStackTrace();
+      fail(String.format("Failed to populate post object: %s", e.getMessage()));
+    } catch (InvocationTargetException e) {
+      e.printStackTrace();
+      fail(String.format("Failed to populate post object: %s", e.getMessage()));
```
jogura: You can not catch these exceptions and remove the `fail()` and the test will behave the same.

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
jogura: Should we be throwing an exception without having any context of what failed? In our other services we aren't throwing exceptions at this level. We are just passing back `Optional.empty()`

armonr: If transaction fails, wouldn't it throw an exception?

service/discussion/src/test/java/com/wikia/discussionservice/resources/PostResourceTest.java
armonr: Not sure what you mean. If I remove the `fail()` calls, test passes.

service/discussion/src/test/java/com/wikia/discussionservice/resources/PostResourceTest.java
armonr: These two tests happen to need a post, but future tests may not need it at all.

service/discussion/src/test/java/com/wikia/discussionservice/resources/PostResourceTest.java
jimmordino: James is suggesting you remove the try and don't catch anything.  Just make the call to ```populate``` without any exception handling and if something throws the test will fail.

service/discussion/src/test/java/com/wikia/discussionservice/resources/PostResourceTest.java
jimmordino: Might be worth putting the ```initPostFromFixture``` into a separate method that takes the fixture as an argument.  That would be reusable in the current code and if we create other fixtures.

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
jimmordino: I think ```Optional.empty()``` is just for cases where the service method is specifically trying to fetch a given entity and it can't be found.  This is a case where we're in a transaction, we just wrote a record, and now we're trying to update that record and the update fails.  That to me, requires an exception.  First, we need to roll back the transaction and that's what the exception does.  Second, we're not finding something we just wrote - that's a fatal situation to me.

jogura: I see, in this particular instance if any one of the following update/create fails we have to roll the entire transaction back. Makes sense. One improvement would be to add some messaging on what exactly failed (i.e., failed to create post revision, update post, etc).

jogura:
LGTM, merge when all the tests pass :+1:

PR #216, opened 2015-05-26
SERVICES-325 add prefix for markers

pchojnacki:
as a stop gap measure to overcome bad index maping
This is temporary fix that allows new data to be indexed in ElasticSearch. In future its possible that this prefix would either be changed or completely removed.


sqreek:
LGTM

PR #218, opened 2015-05-26
Services 341 deploy new mobile configuration service

sqreek:
Rename Mobile-Config to Mobile-Config-v2
Revert old Mobile-Config


pchojnacki:
Looks good. But it would be nice to somehow mark this service as obsolete. e.g. mention in Readme
Maybe some special file or warning when building.

sqreek:
Added deprecated description in mobile config and banner

service/mobile-config-v2/src/main/java/com/wikia/mobileconfig/resources/MobileConfigResource.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.concurrent.TimeUnit;
+
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+
+@Path("/configurations/{platform}/apps/{app-tag}")
```
sqreek: Do we want to stay with this pattern or we want to change it?
We can shortcut this to {platform}/{app-tag}
/cc @Wikia/services-team @aniaMu

sqreek:
Also in Mobile Applications:
applications/platform/{platform}/
can be changed to just {platform}

It's because there is two services instead one, and both have only one resource.

/cc @Wikia/services-team @aniaMu

aniaMu:
What are the new urls to the two new services? We are using http://api.wikia-services.com/mobile-config/applications/platform/{platform} and http://api.wikia-services.com/mobile-config/configurations/{platform}/apps/{app_tag}

Also, just a reminder - we still need to maintain the old urls for the released apps.

service/mobile-config-v2/src/main/java/com/wikia/mobileconfig/resources/MobileConfigResource.java
pchojnacki: I would live it as is, IMO its more flexible to have at least one /root/ element in the path since then you will be able to safely  introduce other resources to this service

pchojnacki:
@sqreek 
If you want to reduce ```applications/platform/{platform}/``
then I'd suggest ```platform/{platform}``` since its more flexible if you'd like to add other resources in the future to this Service.

{platform}

pchojnacki:
@aniaMu 
We are deploying the split services to new urls and we will leave old version running until there are no more users. We also have dashboard to show current usage and alerts that will create a ticket when the service is no longer used.
@sqreek will probably send some email once all of that is wrapped up

pchojnacki:
LeGTM

PR #220, opened 2015-05-26
DIS-46 Add pagination for Forum's ThreadList

jogura:
Adds the ability to request specific offset and limits for thread list.

* Modify SQL query to handle offset and limit
* Add links for `first`, `next`, `previous` and `last` in the response
* Add tests for the service, resource, and mappers

@jimmordino @armonr


jimmordino:
LGTM

PR #221, opened 2015-05-27
DIS-45 Add pagination for posts on a single thread

jogura:
* Add pagination links for posts on threads
* Update `getThread` SQL query to add a `limit` and `offset` and `order by`
* Fix `getThread` SQL to use `revision_id` instead of `post.id` for latest revision (removes duplicate posts with different revisions)
* Refactor ThreadService API to be consistent with the ForumService API with `limit` coming before `offset` in method signatures
* Minor checkstyle fixes
* Add tests for service, resource, mapper, and integration

@jimmordino @armonr


service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
```diff
@@ -45,7 +46,15 @@ public ThreadService(DiscussionThreadDao threadDao, PostService postService) {
...
-    return Optional.ofNullable((T) threadDao.getThread(siteId, threadId, limit, offset));
+    return getForumThread(user, siteId, threadId, limit, offset, POST_ORDER_BY_DEFAULT);
+  }
+
+  public <T extends DiscussionThread> Optional<T> getForumThread(User user,
+                                                                 Integer siteId, Long threadId,
+                                                                 Integer limit, Integer offset,
+                                                                 String orderBy) {
+    return Optional.ofNullable((T) threadDao.getThread(siteId, threadId, limit,
+        limit * offset, orderBy));
```
jimmordino: This doesn't seem right.  I don't think offsets in the API should be in pages, they should be in items.  It's ok for us to provide links that have page numbers in the link text, but the offsets in the URLs should be in items.  Also, if a client wants to use the API and grab 100 entities starting with entity 3 for some reason we should support that.  Let's 'cuss this when we have a chance.

service/discussion/src/main/resources/com/wikia/discussionservice/dao/DiscussionThreadDao.xml
```diff
@@ -165,6 +165,16 @@
             <include refid="lastPostAndRevisionJoin"/>
         WHERE thread.site_id = #{siteId}
             AND thread.id = #{threadId}
+        <choose>
+            <when test="orderBy != null and orderBy.equalsIgnoreCase('asc')">
+                ORDER BY post.id asc
+            </when>
```
jimmordino: Let's also discuss how we expose ordering via the API.  I'm not convinced that exposing the SQL values directly is the best option.  We're going to have custom sort orders at some point (oldest first, newest first, trending, etc.) and it would be better to have order values in the API that map to SQL rather than having the SQL values in the API.

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
jogura: That makes for calculating the previous and next links difficult. This is exactly how the threadlist is implemented as well. With all that said I understand your point. Maybe we can rework a more agreeable solution in a future sprint.

service/discussion/src/main/resources/com/wikia/discussionservice/dao/DiscussionThreadDao.xml
jogura: This is the default ATM, we can revisit in future sprints when we get to those stories.

jimmordino:
LGTM for this sprint.  Let's revisit offset units and order values when we have a chance.

PR #222, opened 2015-05-27
DIS-45/DIS-46 Add pagination for threads & posts 

jogura:
* Add pagination links for posts on threads
* Update getThread SQL query to add a limit and offset and order by
* Fix getThread SQL to use revision_id instead of post.id for latest revision (removes duplicate posts with different revisions)
* Refactor ThreadService API to be consistent with the ForumService API with limit coming before offset in method signatures
* Minor checkstyle fixes
* Add tests for service, resource, mapper, and integration

Adds the ability to request specific offset and limits for thread list.

* Modify SQL query to handle offset and limit
* Add links for first, next, previous and last in the response
* Add tests for the service, resource, and mappers

@jimmordino @armonr


jimmordino:
:shipit:

PR #223, opened 2015-05-28
Classpath config for services and basic support for other dw commands

pchojnacki:
/cc: @rybmat @sqreek


sqreek:
LGTM (was offline before online)

PR #227, opened 2015-05-29
DIS-81 run integration tests in a freshly cloned DB every time

jimmordino:
This PR spikes some new gradle buildscript code that will create a new integration schema for each integration test run.  Developers are responsible for ensuring that the test data they need exists in the ```test_master``` schema on the discussion shared dev server.  After any mods to this schema, the developer should run the gradle ```dumpTestDb``` target and commit the output (```sql/integration_db.sql```).  Running the ```disIntegrationTest``` target will clone a fresh schema from the aforementioned dump file, pass the temp schema name down to the test target, and finally drop the temp clone if all test succeed.

We still need to integrate this with whatever targets Jenkins will run.  We may also want to convert the build script code to a plugin to simplify it.
- [x] gradle build script code to dump, clone, and drop schemas
- [x] custom dropwizard command to show DB connection information from consul/dw config

@armonr @jogura


service/discussion/src/test/resources/discussion-test.yaml
```diff
@@ -33,7 +33,7 @@ databaseConfiguration:
     driverClass: com.mysql.jdbc.Driver
     user: ${consul:DB_USER:"discussion"}
     password: ${consul:DB_PASSWORD:""}
-    url: ${consul:DB_URL:"jdbc:mysql://localhost:3306/discussion"}
+    url: ${consul:DB_URL:"jdbc:mysql://localhost:3306/test_default"}
```
armonr: Is it worth creating a different consul key for test db now that we have this capability?

jimmordino: This setting is overridden at line 151 of the build script, so probably not.

service/discussion/build.gradle
```diff
@@ -65,3 +65,107 @@ run {
...
+            "--user=${->project.ext.DB_USER}", "--password=${->project.ext.DB_PASSWORD}"
+
+    def dropScript = 'DROP SCHEMA ' + project.ext.integrationSchema + ';'
+
+    standardInput = new ByteArrayInputStream(dropScript.getBytes())
+    standardOutput = new ByteArrayOutputStream()
+    ext.output = {
+        return standardOutput.toString()
+    }
+}
```
armonr: Is it possible for the integration test related tasks and procedures to live in a separate gradle file?

jimmordino: It would be possible, but as I mentioned in the description if we're going to change things we should consider going to a gradle plugin first.  Using a plugin would probably simplify this down to < 10 lines in the build script.

armonr:
:shipit:

PR #228, opened 2015-05-29
SOC-782 update User Attribute Service to use jooq as db layer

jsutterfield:
When we were prototyping the User Attribute Service, the Services Team had not yet decided on a prescribed DB library so we went with what the DropWizard docs recommended (JDBI). Now that the Pandora repo has a jooq library, we're porting over the Attributes Service's DB layer to use this new library.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-782


service/user-attribute/schema.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE `attribute` (
```
jsutterfield: Note: I stripped down the schema some [from the prototype](https://github.com/Wikia/populo/blob/master/schema.sql). The product requirements for the MVP didn't call for some of the extra fields which we had before (create_date, update_date, visibility), so I removed them to keep it as simple as possible for the initial stages.

service/user-attribute/build.gradle
```diff
@@ -8,9 +8,8 @@ applicationDefaultJvmArgs = [
 ]
```
jsutterfield: Note: All of the files above this (if looking at this PR from [this view](https://github.com/Wikia/pandora/pull/228/files)) are auto-generated code by jooq. You can largely ignore them and just focus on the PR from here down.

garthwebb: This notice should probably have been put above all the files that should be ignored ... ;)

service/user-attribute/schema.sql
```diff
@@ -0,0 +1,15 @@
...
+  PRIMARY KEY (`name`)
+);
+
+CREATE TABLE `user_attribute` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `user_id` int(11) NOT NULL,
+  `attribute_name` varchar(32) NOT NULL,
+  `value` varchar(1024) DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `user_attribute_idx` (`attribute_name`,`user_id`),
```
garthwebb: This would probably be more useful if ```user_id``` and ```attribute_name``` were switched.  That way you'd be able to query on ```user_id``` alone if you wanted to (e.g. to get all attributes for a suer) and still be able to use the index.

service/user-attribute/build.gradle
jsutterfield: Woops. Sorry. I was thinking it would be easy to see in [this view](https://github.com/Wikia/pandora/pull/228), but if you jump straight to the files changed view (like I usually do), then I guess it does get lost way down the page :)

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java
```diff
@@ -0,0 +1,107 @@
...
+
+public class UserAttributeDAO implements UserAttributeTable {
+
+  public static final String DB_NAME = USERATTRIBUTES.getName();
+
+  private DSLContext database;
+
+  @Inject
+  public UserAttributeDAO(JooqConfigs database) {
+    this.database = database.dslFor(DB_NAME);
```
kvas-damian: This code is duplicate in https://github.com/Wikia/pandora/pull/228/files#diff-62cfbed4babefffaa3d7b19994f45d1eR25
Can we extract it into one place?

kvas-damian:
Please resolve merge conflicts.

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java
jsutterfield: Do you have suggestions on where we could put it?

Personally I feel like this isn't really that bad of code duplication. Both of these classes require a database config object, and then are calling a single method on that object to get the db handle for that particular table. Trying to consolidate that method call feels like it could be adding more complexity than it's gaining in maintainability. I'd argue for leaving it as is, but if there's a good way to consolidate it, we could do that too.

kvas-damian: IMO Most valid solution for code duplication like this is creating abstract class that all DAOs inherits from. On the other hand as you pointed there're only two DAOs and we don't expect more in near future. Having that in mind lets leave it as it is.

kvas-damian:
LGTM

nmonterroso:
We add the generated code to `.gitignore` in `jooq-wikicities`, I think we should do that here as well.

jsutterfield:
Good call @nmonterroso https://github.com/Wikia/pandora/commit/549656843d1f45da83cca64052e7c94b95e74b65

PR #229, opened 2015-06-01
DIS-96/DIS-97 Rename parameter `offset` to `page` to avoid confusion

jogura:
* Rename the `offset` query parameter to `page` to align with implementation
* Update Resources, Service and tests
* Fixes minor naming conventions for constants and checkstyle warnings.

The client side fix for this can be found here: https://github.com/Wikia/discussion-client/pull/2

@armonr @jimmordino


jimmordino:
:+1:

PR #230, opened 2015-06-03
More test coverage for sort params on resource level

armonr:
@jogura @jimmordino 

https://wikia-inc.atlassian.net/browse/DIS-92


service/discussion/src/test/java/com/wikia/discussionservice/resources/ForumResourceTest.java
```diff
@@ -275,4 +279,90 @@ public void getForumWithInvalidResponseGroup() {
...
+    assertThat(response.getStatus(), equalTo(Response.Status.OK.getStatusCode()));
+    Map<String, Object> responseMap = response.readEntity(Map.class);
+    Forum resultForum = HalResponseHelper.extractForumFromResponseMap(responseMap);
+
+    // .equals() on Forum doesn't account for differences in thread lists
+    // removing them before doing the comparison and then comparing the
+    // thread lists separately
+    List<Content> expectedForumThreads = (List<Content>)(List<?>)forum.getThreads();
+    List<Content> resultForumThreadList = (List<Content>)(List<?>)resultForum.getThreads();
+    assertThat(expectedForumThreads.size(), equalTo(resultForumThreadList.size()));
```
jimmordino: Does equalTo not work with the lists themselves?  This only compares size.

service/discussion/src/test/java/com/wikia/discussionservice/resources/ForumResourceTest.java
```diff
@@ -275,4 +279,90 @@ public void getForumWithInvalidResponseGroup() {
...
+
+    forum.setThreads(threads);
+
+    resultForum.setThreads(threads);
+    assertThat(resultForum, equalTo(forum));
```
jimmordino: Can you elaborate on why you need to modify your expected and result forums before comparing?

service/discussion/src/test/java/com/wikia/discussionservice/resources/ForumResourceTest.java
armonr: I re-used the logic in `getForumWithThreadWithResponseGroup` test. Because of the way our mappers are wired, we will not get full data in multiple embedded levels; as a result, the data returned from a request has missing pieces in comparison with a conjured up pojo object.

service/discussion/src/test/java/com/wikia/discussionservice/resources/ForumResourceTest.java
armonr: Same as https://github.com/Wikia/pandora/pull/230#discussion_r31670722

service/discussion/src/test/java/com/wikia/discussionservice/resources/ForumResourceTest.java
jimmordino: I see.  Would it be possible to modify the conjured objects before running the code under test and perform direct comparisons after?  If this doesn't crash the response builder code I think it would be a better test.

armonr: Unfortunately it wont work. the nested objects are used in mappers to build the response, it will crash the code if removed beforehand.

jogura: In addition to that when you compare two objects and the lists are two different instances (different hashcodes) then the assertion of `equals()` fails. Because we serialize a new list they are two different references. At [these lines](https://github.com/Wikia/pandora/pull/230/files#diff-40551784e9e0351c6ed0f124618cde0dR321) we do a comparison of the object in the list. So we check for more than just size.

jimmordino: I understand the fact that the response building code is including a subset of the required domain object fields and I'm OK with that.  I don't understand your assertion that 2 lists have to have any instances in common to be equal.  In any case,  this pattern seems to be the best we can do.

jimmordino:
:+1:

PR #231, opened 2015-06-04
DIS-100 Fixes pagingation off by 1 bug

jogura:
This fix is for the thread list and post lists.

If the mod of the post count & limit is 0 then we must reduce the lastPage by one.
If the limit * (page+1) must also be greater the post count to have a next link.

@armonr @jimmordino


armonr:
It'd be nice to have the page calculation which is common to a few places in its own utility function, to stay DRY and have it easily unit tested.

LGTM

PR #233, opened 2015-06-05
DIS-103 Stubbed out method for onSuccessfulCreatePost()

jogura:
* Fixes casting error in User that prevents successfully creating a post
* Refactor post resource to use new service methods to test thread existence

@jimmordino @armonr


armonr:
As long as we revise this soon in order for stats update to become transactional, LGTM

jimmordino:
This is just obfuscating the logic of creating a post.  Can you tell me why we would want to do this?  I now have to look in 10 different places just to figure out the flow of post creation.  It makes no sense to me.

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
```diff
@@ -1,23 +1,32 @@
...
-import lombok.NonNull;
+import java.util.Optional;
 
 import javax.inject.Inject;
-import java.util.Optional;
 
 public class PostService {
 
   private static final String ERROR_MESSSAGE_NOT_FOUND = "Post not found";
+  private static final Logger LOGGER = LoggerFactory.getLogger(PostService.class);
```
armonr: Is this needed? Why can't we use `DiscussionServiceApplication.LOGGER` instance?

service/discussion/src/main/java/com/wikia/discussionservice/domain/Content.java
```diff
@@ -1,15 +1,8 @@
...
-import com.wikia.discussionservice.enums.ContentType;
+  Long getId();
 
-public abstract class Content {
-
-
-  public abstract Long getId();
-
-
-  public static ContentType getType() {
```
armonr: I guess we still need `getType`.

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
jimmordino: Probably not needed, but if we have the ability to vary levels by class or package might be handy.

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
```diff
@@ -53,24 +62,43 @@ public PostService(PostDao postDao) {
...
+      Post newPost = createPostRecord(post, postRevision);
+      updatePostRevisionId(newPost, postRevision);
+
+      // Because post is stale
+      Optional<Post> postOptional = getPost(newPost.getSiteId(), newPost.getId());
+      // update stats
+      postOptional.map(updatedPost -> {
+        updateStats(updatedPost);
+        return true;
+      }).orElseThrow(() -> new DiscussionServiceException(ERROR_MESSSAGE_NOT_FOUND));
```
jimmordino: Note to team:  We're going to have to create a new annotation for all methods that can be called by the resource layer (DIS-89).  We need to implement so that it checks for and uses the existing transactional information before it attempts to create a new transaction.  If not, then calls like those above to other transactional methods will fail (e.g. a ```createPost``` transaction will connect to primary DB and a call within that transaction to ```getPost``` will fail when it tries to connect to a secondary DB).

service/discussion/src/main/java/com/wikia/discussionservice/domain/Content.java
jogura: I did a find usage and cannot find where any of the static `getType()` are called. Can I remove from the domain objects?

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
jogura: I believe this is the recommended pattern. Here is a link that has some opinions on the matter: http://solvedstack.com/questions/log4j-strategies-for-creating-logger-instances

jimmordino:
LGTM

jimmordino:
user prefs service tests are failing, but all discussion tests pass.

PR #234, opened 2015-06-05
Update User Preference in batch

tiffanyperumpail:
@Wikia/services-team https://wikia-inc.atlassian.net/browse/SERVICES-322

Ticket changes:
PUT request of an array of preference objects
added setUserPreferences method to WikiCitiesDAO and PUT to UserPreferenceResource

Swagger UI changes:
centralized Swagger library (lib:swagger)
added the ability to hide ApiParam from swagger documentation (access = hidden)


lib/dropwizard/databases/db-base/src/main/java/com/wikia/dropwizard/databases/AbstractDatabaseConfigs.java
```diff
@@ -12,9 +12,7 @@
     return configurations;
   }
 
-  public void setConfigurations(Map<String, DatabaseSource> configurations) {
-    this.configurations = configurations;
-  }
+  public void setConfigurations(Map<String, DatabaseSource> configurations) { this.configurations = configurations; }
```
nmonterroso: I think this is automatic code folding in IntelliJ - can you make sure this is actually spread over 3 lines?

lib/dropwizard/databases/db-jooq/src/main/java/com/wikia/dropwizard/databases/jooq/JooqConfigs.java
```diff
@@ -10,30 +10,27 @@
...
+        .set(source)
         .set(dialect);
-    config.set(new DefaultExecuteListenerProvider(listener));
 
     return config;
   }
 
-  public DSLContext dslFor(String name) {
-    return dslFor(name, defaultDialect);
+  public DSLContext dslForWrites(String name, SQLDialect dialect) {
```
nmonterroso: I think we should have the functions `dslForWrites` and `dslForReads` that take only a name, and use the `defaultDialect`

lib/service-core/src/main/java/com/wikia/pandora/http/StandardResponses.java
```diff
@@ -44,6 +44,8 @@ public static Response Unauthorized(String title) {
     return buildResponse(Response.Status.UNAUTHORIZED, title);
   }
 
+  public static Response Successful(String title) {return buildResponse(Response.Status.OK, title); }
```
nmonterroso: Can we spread this out over 3 lines?

service/user-preference/src/main/java/com/wikia/userpreference/dao/WikiCitiesDAO.java
```diff
@@ -75,27 +72,56 @@ public static String byteArrayToString(byte[] array) {
...
-      statement.and(USER_PROPERTIES.UP_PROPERTY.in(
-          this.preferenceWhiteList.stream().map(
-              s -> s.getBytes()
-          ).collect(Collectors.toList())
-      ));
-    }
-    int i = statement.execute();
-    //assumption that can be only one pair user/property
-    return (i == 1);
+    throws SQLException {
```
nmonterroso: I think this needs to be indented

service/user-preference/src/main/java/com/wikia/userpreference/resources/UserPreferenceResource.java
```diff
@@ -69,26 +75,75 @@ public Response getUserPreferences(@ApiParam(value = "The id of the user to list
...
       @NotNull @PathParam("userId") Integer userId,
-      @NotNull @PathParam("preferenceName") String preferenceName,
-      @Auth AuthenticatedUser user,
-      @NotNull String updatedValue
+      @Auth @ApiParam(access = "hidden") AuthenticatedUser user,
+      @NotNull @ApiParam(access = "hidden") List<Preference> updatedUserPreferences
   ) throws SQLException {
 
-    checkIsAuthUserIdIsSameAsUserId(userId, user);
+      checkIsAuthUserIdIsSameAsUserId(userId, user);
```
nmonterroso: Can you correctly indent this?

service/user-preference/src/test/java/com/wikia/userpreference/dao/WikiCitiesDAOIntegrationTest.java
```diff
@@ -38,9 +46,52 @@
...
   public static final String PROPERTY_WEBSITE = "website";
+  public static final String PREFERENCE_LANGUAGE = "language";
+  public static final String PREFERENCE_DATE = "date";
+  public static final String PREFERENCE_MARKETINGALLOWED = "marketingallowed";
   private static final Integer VALID_USER_ID = 2;
   private static final Integer INVALID_USER_ID = -1;
 
+  public static ArrayList<Preference> validInitialPreferencesListHelper() throws java.sql.SQLException {
+    ArrayList<Preference> validInitialPreferencesList = new ArrayList<>();
+    Preference INITIAL_PREFERENCE_LANGUAGE = new Preference("language", "English");
```
nmonterroso: Since these aren't class constants, these should be camelCase (`initialPreferenceLanguage`, etc)

service/user-preference/src/test/java/com/wikia/userpreference/dao/WikiCitiesDAOIntegrationTest.java
```diff
@@ -38,9 +46,52 @@
   public static final String NOT_EXISTING_PROPERTY = "NotExistingProperty";
   public static final String PROPERTY_TIMECORRECTION = "timecorrection";
   public static final String PROPERTY_WEBSITE = "website";
+  public static final String PREFERENCE_LANGUAGE = "language";
+  public static final String PREFERENCE_DATE = "date";
+  public static final String PREFERENCE_MARKETINGALLOWED = "marketingallowed";
   private static final Integer VALID_USER_ID = 2;
   private static final Integer INVALID_USER_ID = -1;
 
+  public static ArrayList<Preference> validInitialPreferencesListHelper() throws java.sql.SQLException {
```
nmonterroso: I don't think this method `throws SQLException`

service/user-preference/src/test/java/com/wikia/userpreference/resources/UserPreferenceResourceTest.java
```diff
@@ -277,6 +282,121 @@ private void addHeaderToMockRequest(Integer userId) {
         .thenReturn(String.valueOf(userId));
   }
 
+  @Test()  //START OF NEW TESTS
```
nmonterroso: can we remove this `// START OF NEW TESTS` and the `// LEFT OFF HERE` comment below?

settings.gradle
```diff
@@ -13,35 +23,24 @@ include 'lib:dropwizard:databases:db-base'
...
-include 'service:indexer-base:delete-worker'
-include 'service:mobile-applications'
-include 'service:mobile-config'
-include 'service:mobile-configurations'
-include 'service:service-creation-test'
-include 'service:user-attribute'
-include 'service:user-preference'
-
-//because notifications service is not used on production and its not under develop,
+include 'service:example'
```
nmonterroso: can we remove `service:example`, `service:service-creation-test`, and `git rm` any test service directories that were committed?

nmonterroso:
looks like this also introduced `service/service-test` - is this an accidental commit of a service creation test? if so, can we `git rm` it?

lib/service-core/src/main/java/com/wikia/pandora/http/StandardResponses.java
drsnyder: You may need to install the google formatting schema. See the [guidelines](https://github.com/Wikia/guidelines/tree/master/Java#syntax-and-formatting).

service/service-test/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+```bash
+../../gradlew run # launch application
+```
+
+To run the tests:
+
+```bash
+../../gradlew test # run all tests
+```
```
drsnyder: It looks the the code in `service/service-test` doesn't belong in this PR.

service/user-preference/src/main/java/com/wikia/userpreference/dao/WikiCitiesDAO.java
```diff
@@ -75,27 +72,56 @@ public static String byteArrayToString(byte[] array) {
...
+
+              for (Preference preference : updatedUserPreferences) {
+                  preferenceName = preference.getName();
+                  preferenceValue = preference.getValue();
+
+                  if (!setUserPreference(DSL.using(configuration), userId, preferenceName, preferenceValue)) {
+                      throw new RuntimeException("User Preference Update Failed");
+                    }
+              }
+          });
```
drsnyder: I think we should do this in one `UPDATE` instead of looping over the list of preferences.

nmonterroso: Hm yeah that seems like a good idea. I quickly googled how to do this, and this seems correct..

```sql
UPDATE user_properties 
SET up_value=(case when up_property=prop1 then val1
                   when up_property=prop2 then val2
                   ...
                   when up_property=propN then valN
              end)
WHERE up_user=userId
  AND up_property IN (prop1, prop2, ..., propN)
```

service/user-preference/src/test/java/com/wikia/userpreference/resources/UserPreferenceResourceTest.java
tiffanyperumpail: Definitely, that was just me being sloppy.

lib/service-core/src/main/java/com/wikia/pandora/http/StandardResponses.java
tiffanyperumpail: Thanks for the heads up, just fixed it and installed the google formatting schema.

service/service-test/README.md
tiffanyperumpail: Yes; accidental commit of a service test.

service/user-preference/src/main/java/com/wikia/userpreference/dao/WikiCitiesDAO.java
drsnyder: I think we want something like the following. We may not be able to assume that the property & value exists for the user.

```SQL
INSERT INTO TABLE user_properties (user_id,up_property, up_value) VALUES (user_id, prop1, val1), (user_id, prop2, val2)
ON DUPLICATE KEY UPDATE up_value=VALUES(up_value);
```

See the MySQL [documentation](http://dev.mysql.com/doc/refman/5.1/en/insert-on-duplicate.html) for the exact syntax.

nmonterroso: Hm, yeah if we don't assume that the property needs to be there before calling then we could use inserts as you've described. But iirc there are tests that specifically test the case where you're not able to create a new user preference, only update existing ones. Was there a reason it was done that way?

drsnyder: > Was there a reason it was done that way?

Hm. Maybe because we were while listing them? We might still want to white list them but my guess would be that we want to allow both creation and updating.

nmonterroso: We should create a ticket to change them to do that, until then we should probably maintain that the service only updates

nmonterroso: Actually now that I think about it, it should be less work to make it updateable/creatable than it is to make it only update-able. Also considering that as a developer it'd be strange to have to check if a preference exists before writing, so the flow would be:

```
PUT /2/language -d foo
  // 404, that preference doesn't exist
POST /2/language -d foo
  // 200
```
or 
```
POST /2/language -d foo
// sorry that already exists, can't create
PUT /2/language -d foo
// 200 ok
```
So it probably makes sense (in this service) for both resources to be create/update if exists.

tiffanyperumpail:
@drsnyder @nmonterroso, I've updated the PR with the changes to allow creation of new preferences and address your other suggestions, can you take a look? Thanks in advance

service/discussion/gradle.properties
```diff
@@ -1 +0,0 @@
-dropwizardGuiceVersion = 0.8.1
```
nmonterroso: Hm, I think this file was accidentally removed.

nmonterroso:
Aside from accidentally deleting `service/discussion/gradle.properties` in 718a8e6b65737a35e0a2c4d8c365f473731ea4c8, LGTM! Anything from you, @drsnyder ?

lib/swagger/src/main/java/com/wikia/swagger/configuration/SwaggerConfiguration.java
```diff
@@ -0,0 +1,44 @@
+package com.wikia.swagger.configuration;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.validation.constraints.NotNull;
+
+public class SwaggerConfiguration {
```
drsnyder: You can use the lombok `@Data` decorator here to save having to create the getters and setters.

nmonterroso: @armonr had the same recommendation [here](https://github.com/Wikia/pandora/commit/b034aa4d6ed4932b508629fbefd6fa1e0a0a3f26#commitcomment-11434607). The point is moot, though since in my work to upgrade us to Swagger 2.0 this class no longer exists.

lib/swagger/src/main/java/com/wikia/swagger/SwaggerBundleFactory.java
```diff
@@ -0,0 +1,39 @@
...
+        SwaggerBundleConfiguration bundleConfig = new SwaggerBundleConfiguration();
+
+        if (configuration instanceof SwaggerConfigurationProvider) {
+          SwaggerConfigurationProvider configProvider = (SwaggerConfigurationProvider) configuration;
+          SwaggerConfiguration swaggerConfig = configProvider.getSwaggerConfiguration();
+
+          bundleConfig.setStaticAssetPrefix(swaggerConfig.getStaticAssetPrefix());
+          bundleConfig.setAuthHeader(swaggerConfig.getAuthHeader());
+          bundleConfig.setSecurityTokenType(swaggerConfig.getSecurityTokenType());
+        } else {
```
drsnyder: Is there a way to make this an error or a unreachable state? Should `C` above implement `SwaggerConfigurationProvider`?

service/user-preference/src/main/java/com/wikia/userpreference/dao/WikiCities.java
```diff
@@ -10,7 +13,9 @@
   public Optional<ImmutableMap<String, String>> getUserPreferences(Integer userId)
       throws java.sql.SQLException;
 
-
   boolean setUserPreference(Integer testUserId, String prefKey, String newValue)
       throws SQLException;
+
+  boolean setUserPreferences(Integer testUserId, List<Preference> testNewUserPreferences)
```
drsnyder: Should this be `newUserPreferences` (drop the `test`)?

service/user-preference/src/main/java/com/wikia/userpreference/dao/WikiCitiesDAO.java
```diff
@@ -76,26 +77,60 @@ public static String byteArrayToString(byte[] array) {
...
+          .stream()
+          .filter(p -> !preferenceWhiteList.contains(p.getName()))
+          .collect(Collectors.toList());
+
+      if (!invalidUpdatedUserPreferences.isEmpty()) {
+        return false;
+      }
+    }
+
+    if (updatedUserPreferences.isEmpty()) {
```
drsnyder: I would move this block to the top for clarity.

service/user-preference/src/main/java/com/wikia/userpreference/dao/WikiCitiesDAO.java
```diff
@@ -76,26 +77,60 @@ public static String byteArrayToString(byte[] array) {
...
+          preference.getName().getBytes(),
+          preference.getValue().getBytes());
+    });
+    // JOOQ doesn't properly support DuplicateKeyUpdate() when inserting multiple values
+    // see https://github.com/jOOQ/jOOQ/issues/2134
+    statement
+        .onDuplicateKeyUpdate()
+        .set(
+            field(USER_PROPERTIES.UP_VALUE.getName()),
+            (Object) field("VALUES(" + USER_PROPERTIES.UP_VALUE.getName() + ")"));
```
drsnyder: Minor: this is where string-concatenating SQL libraries break down-- when the complexity goes up their usefulness goes down.

service/user-preference/src/main/java/com/wikia/userpreference/resources/UserPreferenceResource.java
```diff
@@ -91,9 +108,45 @@ public Response updateUserPreference(
...
+      @Auth @ApiParam(access = "hidden") AuthenticatedUser user,
+      @NotNull @ApiParam(access = "hidden") List<Preference> updatedUserPreferences
+  ) throws SQLException {
+
+    checkIsAuthUserIdIsSameAsUserId(userId, user);
+
+    if (wikiCities.setUserPreferences(userId, updatedUserPreferences)) {
+      return StandardResponses.Successful("Successful update of user preferences");
+    } else {
+      return StandardResponses.NotFound("No user preference found");
```
drsnyder: From the `setUserPreferences` code this is probably a `400 Bad Request` and not a `404 Not Found`. This should probably only be a `404` if the user doesn't exist.

service/user-preference/src/test/java/com/wikia/userpreference/resources/UserPreferenceResourceTest.java
```diff
@@ -277,6 +281,77 @@ private void addHeaderToMockRequest(Integer userId) {
...
+
+    addHeaderToMockRequest(VALID_USER_ID);
+
+    Response response = getResponseForUpdateMultiplePreferences(VALID_USER_ID,
+        newPreferencesListHelper());
+
+    assertEquals(HttpStatus.NOT_FOUND_404, response.getStatus());
+
+    verify(wikiCities, times(1)).setUserPreferences(VALID_USER_ID, newPreferencesListHelper());
+  }
```
drsnyder: :+1: This is how tests should read :)

drsnyder:
I think the only change I would make is status code in the controller. OTW, LGTM
:sheep: it, :ship: it, :shipit:

service/user-preference/src/main/java/com/wikia/userpreference/dao/WikiCitiesDAO.java
nmonterroso: :(

lib/swagger/src/main/java/com/wikia/swagger/SwaggerBundleFactory.java
nmonterroso: Yeah, this is done in the new Swagger 2.0 stuff as `<C extends Configuration & SwaggerConfigurationProvider>`. I'm wiser now so I know that a bound can have two Types it's bound to.

PR #236, opened 2015-06-08
Updating Insert Test SQL

jogura:
We were creating a thread on the ROOT Forum, which is not ideal

* Added additional forums for different test scenarios
* Use constants in place of string literals for response codes

@jimmordino @armonr


jimmordino:
:+1:

armonr:
LGTM

PR #237, opened 2015-06-08
Update to the test data causes the tests to fail

jogura:
* Change forum id to forum with threads
* Modifies the expected post count to be greater or equal to 1

@jimmordino @armonr


jimmordino:
:+1: , pending test success

PR #238, opened 2015-06-08
DIS-106 Add stored procedure to update post statistics

jimmordino:
Add stored procedure that properly updates thread and forum statistics when a post is added.  Include locking as appropriate to ensure that concurrent transactions update with accurate stats.

- [x] Add stored proc
- [x] Add mapper implementation that calls method
- [x] Add method to mapper interface
- [x] Call update stats as part of create post transaction 
- [x] Change isolation level to READ_COMMITTED


jogura:
Retargeting to DIS-51

PR #239, opened 2015-06-08
DIS-106 Add stored procedure to update post statistics

jogura:
Add stored procedure that properly updates thread and forum statistics when a post is added.  Include locking as appropriate to ensure that concurrent transactions update with accurate stats.

- [x] Add stored proc
- [x] Add mapper implementation that calls method
- [x] Add method to mapper interface
- [x] Call update stats as part of create post transaction 
- [x] Change isolation level to READ_COMMITTED

@jimmordino @armonr @jogura


service/discussion/sql/create_discussion_db.sql
```diff
@@ -410,6 +410,57 @@ READS SQL DATA
...
+    DECLARE postCount INTEGER DEFAULT -1;
+    DECLARE maxPostId BIGINT DEFAULT 0;
+
+    SELECT thread_id INTO threadId FROM post WHERE id = postId AND site_id = siteId;
+    IF threadId = 0 THEN
+      SIGNAL SQLSTATE '45001' SET MESSAGE_TEXT = 'Post not found';
+    END IF;
+
+    -- lock
+    SELECT post_count, last_post_id FROM thread WHERE id = threadId AND site_id = siteId FOR UPDATE;
```
jogura: Is this performing a row level lock so other threads can READ/WRITE on other threads without getting block?

service/discussion/sql/create_discussion_db.sql
```diff
@@ -410,6 +410,57 @@ READS SQL DATA
...
+    SELECT COUNT(id), MAX(id) INTO postCount, maxPostId
+    FROM post
+    WHERE thread_id = threadId AND site_id = siteId;
+
+    IF postCount = -1 OR maxPostId = 0 THEN
+      SIGNAL SQLSTATE '45002' SET MESSAGE_TEXT = 'Unable to update thread stats';
+    END IF;
+
+    UPDATE thread SET post_count = postCount, last_post_id = maxPostId
+    WHERE id = threadId AND site_id = siteId;
```
jogura: Is the lock from above released after the entire stored proc completes? 
Is the lock on this thread required after this line?  
Could we shorten up the lock time by adding explicit BEGIN/END?

jogura:
LGTM - we can revisit shortening the lock in future if it is required

service/discussion/sql/create_discussion_db.sql
jimmordino: Yes, row level exclusive lock.

service/discussion/sql/create_discussion_db.sql
jimmordino: All locks are released upon commit or rollback and not before.  I haven't seen any way to change that .  These should be very fast queries because they are only reading a fraction of their tables.  If properly indexed we should have no problems with performance.

PR #240, opened 2015-06-09
DIS-105 Use relative URL in links to avoid exposing internal server names

jogura:
* Do not expose the internal server names
* Leaving an unused instance of UriInfo in case there is a need in the future

@jimmordino @armonr


jimmordino:
LGTM, but as we discussed, we should see if we can find the original host somewhere in the incoming request.

jogura:
That could be problematic if the request is coming from an Aggregation/API gateway, then we'd show that server's name as well.

PR #241, opened 2015-06-09
Swagger 2.0

nmonterroso:
@Wikia/services-team @jogura @jimmordino @armonr 
https://wikia-inc.atlassian.net/browse/SERVICES-369

Upgrading Swagger docs to 2.0. Note that previously we were doing `@ApiModelProperty(access = "hidden")` and using a custom schema converter, but this is built into the swagger annotations and can be used with `@ApiModelProperty(hidden = true)`. 

Also adding Swagger to `service-core` and updating the service created using `:tools:create-new-service:run` to be Swagger annotated. 

~~Note that this builds off #234, so the changeset here will be bigger until that's merged, which should be later today.~~


armonr:
LGTM. I think a rebuild on jenkins is needed

PR #243, opened 2015-06-10
DIS-114 Fixes after Swagger 2.0 changes

jogura:
Merges in the changes to Swagger 2.0 and makes specific Discussion Service changes

* Explicitly define default values for @ApiParam declared fields
* Add @JsonProperty annotations with explicity names for output

__NOTE:__ Swagger UI in mesos is not 100% functioning with these changes but @nmonterroso is aware and currently looking into a solution. We will merge those changes in when available.

@jimmordino @armonr


jimmordino:
:+1:

PR #244, opened 2015-06-10
DIS-115 remove fully qualified references to objects in discussion schema

jimmordino:
We're going to use copies of the discussion schema for integration testing and these will be named appropriately.  However, some of the routines in the schema call other routines using fully qualified names, e.g. ```discussion.dis_next_id()```.  Removing the fully qualified references will make things portable and allow the schema to exist under a different name and function as expected.

@armonr @jogura


jogura:
I know this isn't used for integration tests however to be consistent should consider doing the same thing to this file: https://github.com/Wikia/pandora/blob/dis-master/service/discussion/sql/insert_test_data.sql

jimmordino:
Actually, it's consistent now.  We ```use discussion``` in the creation script.  There are not fully qualified references in the script to insert test data, so that data is portable to other schemas after being inserted into discussion.

jogura:
LGTM, the test failures seem to be other services unrelated.

armonr:
LGTM

PR #246, opened 2015-06-10
Swagger fixes

nmonterroso:
@jogura 
Serve static assets from non-prefixed path so swagger ui can access static assets through the api gateway


jogura:
LGTM

tools/create-new-service/templates/src/main/resources/template.yaml
```diff
@@ -38,8 +38,8 @@ server:
     port: 8081
 swaggerConfiguration:
   resourcePackage: com.wikia.services.@templateLower@.resources
+  uriPrefix: ${consul:SWAGGER_PREFIX:/}
```
armonr: Should this have quotes around it?

nmonterroso:
Nope, there are no quotes in yaml

armonr:
Not in plain yaml, but around the "default value" in our conventional consul key/value string: `${consul:KEY_ON_CONSUL:<default_value>`. I remember a case where the default value of `/` would not work properly unless wrapped in quotes.

nmonterroso:
Hm, I haven't run into that issue yet, but I suppose there's no harm in having it there. If you open a PR with that change I'll LGTM it :P

PR #247, opened 2015-06-11
DIS-107 Fix 404 on forums without threads

armonr:
Fixes the issue by returning an empty list of threads from DAO for thread-less forums
Also make a test for this to make sure forums without thread are returned properly

@jimmordino @jogura 

https://wikia-inc.atlassian.net/browse/DIS-107


jogura:
I believe we want an integration test for this as well. The resource test doesn't exercise the modified SQL.

service/discussion/src/main/resources/com/wikia/discussionservice/dao/ForumDao.xml
```diff
@@ -112,7 +113,7 @@
     <include refid="com.wikia.discussionservice.dao.DiscussionThreadDao.firstPostColumns"/>,
     <include refid="com.wikia.discussionservice.dao.DiscussionThreadDao.lastPostColumns"/>
     FROM forum
-    JOIN thread on (
+    LEFT JOIN thread on (
       forum.id = thread.container_id
```
jimmordino: Note:  if a forum goes from allowing threads to not allowing threads we should no longer display the threads in that forum.  So we'll either have to prevent that state transition until all threads are deleted or we'll have to add a where predicate to this query to filter out threads in forums that don't allow them.

armonr: Good point and thanks for following up on it

jogura:
LGTM - Thanks for adding that test :+1:

service/discussion/src/main/resources/com/wikia/discussionservice/dao/DiscussionThreadDao.xml
```diff
@@ -111,18 +111,18 @@
...
-        JOIN post_revision AS first_post_revision ON (first_post_revision.id = first_post.revision_id)
+        LEFT JOIN post first_post ON (thread.first_post_id = first_post.id)
+        LEFT JOIN post_revision AS first_post_revision ON (first_post_revision.id = first_post.revision_id)
     </sql>
 
     <sql id="lastPostAndRevisionJoin">
-        JOIN post last_post ON (thread.last_post_id = last_post.id)
-        JOIN post_revision AS last_post_revision ON (last_post_revision.id = last_post.revision_id)
+        LEFT JOIN post last_post ON (thread.last_post_id = last_post.id)
+        LEFT JOIN post_revision AS last_post_revision ON (last_post_revision.id = last_post.revision_id)
```
jimmordino: I don't think these should always be left joins.  If these are used at the thread level they should be inner joins.  We should substitute ```${joinType}``` for ```LEFT``` and within the ```<include>``` tag add something like ```<property name="joinType" value="LEFT"/>```.  Then we can also specify ```value=""``` or ```value="INNER"``` when we want to use these as inner joins.

PR #249, opened 2015-06-11
DIS-114 Add @ApiParam for input fields for POST/PUT operations

jogura:
The new Swagger 2.0 changes does not automatically document fields

* Add the `@ApiParam` explicitly for input fields
* Clean up some fields in the input that should not be exposed to the clients

@jimmordino @armonr


jimmordino:
:+1:

PR #251, opened 2015-06-12
DIS-118 Add upvote_count column to Post table

jogura:
* Add column to both create db scripts
* Change SQL to use upvote_count instead of ratings

@armonr @jimmordino


service/discussion/sql/create_discussion_db.sql
```diff
@@ -112,6 +112,7 @@ CREATE TABLE IF NOT EXISTS `post` (
   `deleted_position` INT,
   `revision_id` BIGINT,
   `rating` INT,
```
armonr: Are we keeping rating?

armonr:
LGTM

service/discussion/sql/create_discussion_db.sql
jogura: For now, so we don't break existing services deployed we are doing an additive change. Once we have services deployed that do not reference `ratings` then we can delete the column.

armonr: Sounds good!

PR #253, opened 2015-06-12
DIS-95 Add upvote count to post domain object and response groups

jogura:
Exposes the upvote count for posts in the responses.

* Update the db schema to add the `upvote_count` to the `POST` table
* Update SQL to use new query
* Update domain object to include `upvotes` field
* Update mappers to include field
* Update tests to test for new field

@jimmordino @armonr


armonr:
:shipit:

service/discussion/src/test/java/com/wikia/discussionservice/resources/PostResourceIntegrationTest.java
```diff
@@ -110,10 +110,10 @@ public void postReplyToThread() {
     assertThat(newPost.getRawContent(), equalTo(postBody));
 
     testThread = getForumThread(SITE_ID_GLEE, TEST_REPLY_THREAD_ID);
-    assertThat(testThread.getPostCount(), equalTo(new Integer(originalThreadPostCount + 1)));
+    assertThat(testThread.getPostCount(), equalTo(originalThreadPostCount + 1));
     assertThat(testThread.getLastPostId(), equalTo(newPost.getId()));
     testForum = getForum(SITE_ID_GLEE, testThread.getForumId());
-    assertThat(testForum.getPostCount(), equalTo(new Integer(originalForumPostCount + 1)));
+    assertThat(testForum.getPostCount(), equalTo(originalForumPostCount + 1));
   }
```
jimmordino: FYI, I don't like to use auto-boxing.  Explicit is better than implicit, even in Java.

jimmordino:
LGTM

service/discussion/src/test/java/com/wikia/discussionservice/resources/PostResourceIntegrationTest.java
jogura: That's one of the warnings from our static checks. 'Unnecessary auto-boxing'. I changed it to the faster `Integer.valueOf()` and it still complained.

PR #254, opened 2015-06-15
Central Swagger UI

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-358

Introduces a new service `central-swagger-ui` which queries Consul for services with matching tags and serves a Swagger UI for those swagger docs. By default, the tags searched for are `production` and `swagger`, and docs are assumed to live in `/swagger.json` relative to the base path, ex: http://services.wikia.com/community-data/swagger.json.

Note - static files (css, js, images) in `service/central-swagger-ui/src/main/resources/static-assets` are unedited resources provided by [swagger-ui](http://swagger.io/swagger-ui/)


lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/api/ServiceAddress.java
```diff
@@ -0,0 +1,32 @@
+package com.wikia.dropwizard.consul.api;
+
+public class ServiceAddress {
```
drsnyder: Will `@Value` work here?

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/guice/ConsulApiModule.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class ConsulApiModule extends AbstractModule {
+
+  @Override
+  protected void configure() {
+
+  }
+
+  @Provides
+  public ConsulApi getConsulApi(Optional<Consul> consulOptional) {
```
drsnyder: Should this be `Optional`? There is no `empty` logic.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/api/ServiceAddress.java
nmonterroso: Lombok isn't available in this library

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/guice/ConsulApiModule.java
nmonterroso: You mean "should this be `Optional<ConsulApi> getConsulApi(Optional<Consul>)` ?" or "should this be `ConsulApi getConsulApi(Consul)`?"

The parameter is `Optional<Consul>` because that's what's provided [here](https://github.com/Wikia/pandora/blob/master/lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/registration/guice/ConsulProvider.java). I'll change this to do an `orElseThrow`, since that's probably what we want to happen.

nmonterroso: Eh, on second thought, maybe `Optional<ConsulApi>` makes more sense

drsnyder: I was referring to the param `Optional<Consul> consulOptional`. It looks odd passing in an optional and then using it directly without first checking to see if it's set. On the surface it seems unnecessary.

drsnyder:
Minor comments. OTW, lgtm

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/guice/ConsulApiModule.java
nmonterroso: @drsnyder and I talked about this offline, and it came out that it's probably better to have `ConsulApiModule` be opt-in (it's not installed by `ConsulGuiceApplication` by default), so if you opt-in to the module, it's OK to do an `orElseThrow` in the module. See 1a98db3ae305e3fa6d815402e56a78e71ceb0f2b

PR #255, opened 2015-06-16
DIS-124/127 Implement ThreadService to create thread and update SQL

jogura:
* Implement the create thread method in ThreadService
* Add appropriate logging and error handling
* Update SQL to create thread
* Refactor PostService to allow updating stats

@jimmordino @armonr


service/discussion/src/main/java/com/wikia/discussionservice/enums/ContentType.java
```diff
@@ -20,7 +20,7 @@
   private final String label;
 
 
-  private ContentType(@NonNull String label) {
+  ContentType(@NonNull String label) {
```
jimmordino: Why wouldn't this be private?  Are we planning to add enum instances on-the-fly?

service/discussion/src/main/java/com/wikia/discussionservice/resources/ThreadResource.java
```diff
@@ -156,8 +166,17 @@ public Response createThread(
...
+
     Post post = postInput.getPost();
 
+    if (!forumService.isExistingForum(forumId.get())) {
+      throw new NotFoundException(
+          String.format("No forum found for site id: %s with forum id: %s",
+                        siteId.get(),
+                        forumId.get()));
+    }
+
```
jimmordino: Can we generalize this API include authz as well as forum existence?  We don't need to implement all of the authz, per se, but it would be nice to have a single call that checks if the forum exists, accepts threads, isn't locked, and the user can post there, etc.

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
```diff
@@ -58,38 +66,65 @@ public ThreadService(DiscussionThreadDao threadDao, PostService postService) {
...
+          // Update the thread
+          forumThread.setFirstPost(newPost);
+          forumThread.setLastPost(newPost);
+          forumThread.setPosts(Lists.newArrayList(newPost));
+          forumThread.setFirstPostId(newPost.getId());
+          forumThread.setLastPostId(newPost.getId());
+          updateThread(user, forumThread);
+
+          // Update the statistics
+          postService.updatePostStats(newPost);
```
jimmordino: This is being called twice, once here and once in ```createPost```

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
```diff
@@ -115,16 +150,19 @@ public Boolean deleteThread(Integer siteId, Long threadId) {
       executorType = ExecutorType.SIMPLE,
       isolation = Isolation.SERIALIZABLE,
       rethrowExceptionsAs = DiscussionServiceException.class,
-      exceptionMessage = "Unable to create thread")
+      exceptionMessage = "Unable to update thread")
   @SneakyThrows(value = UnauthorizedException.class)
```
jimmordino: I'm wondering how well  ```@SneakyThrows``` plays with ```rethrowExceptionAs``` - the latter may override the former.

service/discussion/src/main/java/com/wikia/discussionservice/enums/ContentType.java
jogura: `private` is redundant in Enums

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
jogura: Good call:  be02bd5

service/discussion/src/main/java/com/wikia/discussionservice/enums/ContentType.java
jimmordino: We'll agree to disagree on this one, but my feeling is that being explicit is better, especially since the compiler is inserting ```private``` if it's not in the code.  In other words, if the constructor is going to be private in the byte code it should be declared private in the source code, even if it's not strictly necessary.

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
```diff
@@ -58,38 +66,65 @@ public ThreadService(DiscussionThreadDao threadDao, PostService postService) {
...
-    forumThread.setFirstPostId(post.getId());
-    forumThread.setLastPostId(post.getId());
-    threadDao.updateThread(forumThread);
-
-    // TODO: Update thread and forum stats
+    try {
+      forumThread.setSiteId(post.getSiteId());
+      forumThread.setForumId(forumId);
+      forumThread.setTitle(post.getTitle());
+      forumThread.setCreatorId(post.getCreatorId());
```
armonr: Setter calls can be moved outside of try block

service/discussion/src/main/java/com/wikia/discussionservice/resources/ThreadResource.java
jogura: The [AuthorizationService](https://github.com/Wikia/pandora/blob/DIS-124/service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java#L10-10) might be a good candidate for this. I was already thinking about going down this line if you read the TODOs

service/discussion/src/main/java/com/wikia/discussionservice/enums/ContainerType.java
```diff
@@ -0,0 +1,16 @@
...
+public enum ContainerType {
+  FORUM("forum");
+
+  @Getter
+  private final String label;
+
+
+  ContainerType(@NonNull String label) {
+    this.label = label;
+  }
```
armonr: Also override `toString` to return lower-case value, for consistency in data.

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
jogura: I've run some tests and if any exception is thrown from this method that is what the caller gets (no translation). I believe that if exceptions are thrown from the MyBatis layer they will get caught and rethrown as a `DiscussionException`

service/discussion/src/main/java/com/wikia/discussionservice/enums/ContainerType.java
jimmordino: MyBatis is using ```name()``` in its ```EnumTypeHandler``` so overriding ```toString``` won't help.  We'll need to lowercase the value in the SQL insert/update statements to be sure the value is lowercased in the DB.  It won't matter on the way out of the DB because there's no setter on the domain object for ```containerType```.

jimmordino:
LGTM, once we lowercase container type in the DB.

service/discussion/src/main/java/com/wikia/discussionservice/enums/ContainerType.java
jogura: ff556fe

armonr:
LGTM

PR #257, opened 2015-06-16
DIS-125 Update stored procedure to update a forum's thread count

jogura:
Update the forum's thread count as part of the post_stats update

@jimmordino @armonr


service/discussion/sql/create_discussion_db.sql
```diff
@@ -451,11 +452,17 @@ CREATE PROCEDURE `update_post_stats` (IN siteId INTEGER, IN postId BIGINT)
     JOIN thread ON (post.thread_id = thread.id)
     WHERE thread.container_id = forumId AND thread.site_id = siteId;
 
-   IF postCount = -1 OR maxPostId = 0 THEN
+    SET threadCount = -1;
```
armonr: Isn't this initialized already?

service/discussion/sql/create_discussion_db.sql
```diff
@@ -451,11 +452,17 @@ CREATE PROCEDURE `update_post_stats` (IN siteId INTEGER, IN postId BIGINT)
...
+    SELECT COUNT(thread.id) INTO threadCount
+    FROM forum
+    JOIN thread on (forum.id = thread.container_id)
```
armonr: `ON` capitalized :)

armonr: And join on container_type = 'forum'?

jimmordino: The second query is not necessary.  Just use
```
SELECT COUNT(post.id), MAX(post.id), COUNT(DISTINCT thread.id) INTO postCount, maxPostId, threadCount
```
in the first query.

jogura: That is way simpler thanks: 7c80c8f

jimmordino:
:+1: , thanks

PR #258, opened 2015-06-17
DIS-129 Implement constraint checks for creating a thread

jogura:
* Check for the existence of the forum
* Check if the thread isViewable, allowsThread and is not deleted
* Refactor authorize method to include `siteId`
* Add unit tests to test authorization method with different conditions

@jimmordino @armonr


service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
```diff
@@ -15,24 +20,56 @@
...
+                              + "The forum does not allow threads.", forumId));
+      }
+
+      if (!forum.getIsViewable()) {
+        throw new DiscussionServiceException(
+            String.format("Unable to create the thread on forum id: %s."
+                              + " The forum is not viewable.", forumId));
+      }
+
+      if (forum.getIsDeleted()) {
```
jimmordino: I think we should order these checks as 1. viewable,  2. deleted,  3. accepts threads
If a forum isn't viewable then it's private, and you shouldn't know anything else about it
If a forum is deleted you should see that exception before you see one about accepting threads

jimmordino:
LGTM

service/discussion/src/main/java/com/wikia/discussionservice/resources/ThreadResource.java
```diff
@@ -170,13 +170,6 @@ public Response createThread(
 
     Post post = postInput.getPost();
 
-    if (!forumService.isExistingForum(forumId.get())) {
-      throw new NotFoundException(
-          String.format("No forum found for site id: %s with forum id: %s",
-                        siteId.get(),
-                        forumId.get()));
-    }
```
armonr: Why was this removed?

jogura: The idea was to move this down into the a central place to avoid unnecessary db round trips.

PR #259, opened 2015-06-17
Swagger UI Per-API Config

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-408

Modifies the `central-swagger-ui` app to allow for per-API configuration overrides in Consul. These overrides should live in K/V `central-swagger-ui/<env>/<service-name>/<property>`. The following properties are override-able:

* `swagger_path` - the path serving Swagger documentation, defaulting to `swagger.json`
* `base_url` - the base url for accessing the swagger docs, defaulting (from YAML config) to `https://services.wikia.com`
* `auth.name` - The swagger doc's [SecurityDefinition (spec 2.0)](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#securitySchemeObject) or [Authorization (spec 1.2)](https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#515-authorization-object) name for the default authorization method used in the SwaggerUI, defaulting to (from YAML config) to `access_token`
* `auth.key` - The `name` (2.0) or `keyname` (1.2) value for the authentication defined by `auth.name`, defaulting to `X-Wikia-AccessToken`
* `auth.location` - The `in` (2.0) or `passAs` (1.2) value for the authentication defined by `auth.name`, defaulting to `header`

This will allow us to serve Helios swagger documentation from `/swagger` (or wherever), and in the future allow us to override any of these other values as necessary. 

In order to leverage the existing `ConsulKeyValue` `Provider` used during application initialization, some of the boot-up logic had to be refactored to avoid code duplication.


ArturKlajnerok:
Lgtm :+1:

PR #260, opened 2015-06-17
Post creation unit tests, couple of fixes

armonr:
Tests for post creation:
* Resource level tests
* Service level tests
* Authz tests

And:
* Adds `Forbidden` response builder to StandardResponses (outside of DIS)
* Implements `isThreadOwner`
* Adds convenience response code constants for auth tests
* Minor styling fixes in PostService in a separate commit

@jogura @jimmordino


service/discussion/src/main/java/com/wikia/discussionservice/resources/ThreadResource.java
```diff
@@ -97,6 +95,8 @@ public ThreadResource(ThreadService threadService,
     this.forumService = forumService;
   }
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(ThreadResource.class);
+
```
jogura: Any reason to declare the `LOGGER` after the constructor?

service/discussion/src/test/java/com/wikia/discussionservice/resources/ThreadResourceAuthzAnnotationTest.java
```diff
@@ -109,10 +112,25 @@ protected UriBuilder getThreadResourceUrl(int siteId, String resourceType, long
...
+    UriBuilder uriBuilder = getThreadsResourceUrl(SITE_ID, forumId);
+    assertPostStatus(RESPONSE_CODE_FORBIDDEN, webTarget(uriBuilder), createThread, null);
+    // Not concerned what the response code is just that the authorization code was exercised
+    Integer[] expectedStatus = new Integer[]{RESPONSE_CODE_CREATED, RESPONSE_CODE_NO_CONTENT,
+                                             RESPONSE_CODE_NOT_FOUND, RESPONSE_CODE_BAD_REQUEST};
+
+    // todo: fix. This is supposed to be 201 but is actually 500
+    //  assertPostStatus(expectedStatus, webTarget(uriBuilder), createThread,
+    //                   userAuthorizationHeader);
+    assertPostStatus(expectedStatus, webTarget(uriBuilder), createThread, adminAuthorizationHeader);
```
jogura: `500` errors aren't a good sign of a successful test. At the very minimum it should be some variant of `400` series error. This is because if the input is malformed then it will throw a `500` well before the auth code is exercised.

service/discussion/src/main/java/com/wikia/discussionservice/resources/ThreadResource.java
armonr: No, I just removed it in a commit and placed it back in another.

service/discussion/src/test/java/com/wikia/discussionservice/resources/ThreadResourceAuthzAnnotationTest.java
armonr: `500` is a result of bad foreign keys created in an intermediate step where a row gets created in the database. So there are a couple of actual problems this test surfaces: 1. broken transactionality, 2. broken auth (a row gets created before a 403 finally gets emitted.)

I hold off on this test until after merging with your refactored code; anything broken can be fixed along with uncommenting this test.

jogura:
Closing in favor of another branch that has most recent changes merged in.

PR #261, opened 2015-06-18
Change WikiCities.getUserPreferences return type

nmonterroso:
@Wikia/services-team 

Changing the `WikiCities.getUserPreferences` return type to `List<Preference>` so the response can be more easily understood by clients. Also adding swagger documentation for the return type.


service/user-preference/src/main/java/com/wikia/userpreference/resources/UserPreferenceResource.java
```diff
@@ -62,9 +62,10 @@ public UserPreferenceResource(WikiCities wikiCities) {
   @ApiOperation(
       value = "Returns all the global user preferences for a user",
       notes = "Returns a white listed subset of the users global preferences",
-      responseContainer = "Map")
+      response = Preference.class,
+      responseContainer = "List",
+      produces = MediaType.APPLICATION_JSON)
   @ApiResponses(value = {
-      @ApiResponse(code = 200, message = "Successful retrieval of user preferences"),
```
sqreek: Why this is removed?

sqreek:
LGTM,

service/user-preference/src/main/java/com/wikia/userpreference/resources/UserPreferenceResource.java
nmonterroso: The `response` and `responseContainer` in the `@ApiOperation` represent the success case as well, but don't show up in the right place in the generated Swagger doc when there is an `@ApiResponse` with `code = 200`. We need the `responseContainer`, which isn't available in `@ApiResponse`.

PR #263, opened 2015-06-19
DIS-132 Implement Correct Transactionality Behavior in DAO

jogura:
All classes that use `@Transactional` annotation must be injected from the MyBatis-Guice module
Add integration tests for rollback tests

@jimmordino @armonr


lib/service-core/src/main/java/com/wikia/pandora/core/auth/ServiceRealm.java
```diff
@@ -84,9 +84,10 @@ public boolean supports(AuthenticationToken token) {
         userInfo.put("permissions", Sets.newHashSet("*"));
         break;
       case "2":
+      case "3":
         userInfo.put("roles", Sets.newHashSet(Role.USER.toString()));
         break;
-      case "3" :
+      case "4" :
```
jimmordino: These users (id 1-n) actually exist in the wikicities DB.  I doubt we would have collisions, but we may want to use negative IDs for special testing.

armonr:
Other than John's comment, LGTM

PR #264, opened 2015-06-19
make sure the Statement used for the query is also closed

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-436

When debugging the memory leak using VisualVM I found that instances of `JDBC4Result` were not being properly GC'd. After digging a bit, I found that the `Statement` created by `conn.createStatement()` needs to also be `close()`d. Adding it to the `try-with-resources` declaration resulted in the `JDBC4Result` instances getting properly cleaned up.


drsnyder:
lgtm

PR #265, opened 2015-06-19
DIS-133 Modify ServiceRealm to support real user ids

jogura:
To support Authentication/Authorization correctly if there is a valid user id present then assume the request is from a logged in user.

@jimmordino @armonr


armonr:
LGTM, let's merge after tests pass

PR #266, opened 2015-06-22
DIS-50 - As a user, I want to create discussion threads

jogura:
This covers the functionality for implementing thread creation on an existing forum.

Aside from the basic thread creation, the main work was around authorization, transactionality and enforcing the business logic of who/when a thread could be created.

Tests were written at each layer.

@jimmordino @armonr


armonr:
:shipit:

PR #268, opened 2015-06-23
JobQueue API

tiffanyperumpail:
@Wikia/services-team https://wikia-inc.atlassian.net/browse/SERVICES-378

This PR centralizes and implements queuing tasks to RabbitMQ behind an API. 
Supports queuing multiple tasks in batch.


service/job-queue/src/main/java/com/wikia/services/jobqueue/guice/QueueBrokerConfiguration.java
```diff
@@ -0,0 +1,8 @@
+package com.wikia.services.jobqueue.guice;
+
+public interface QueueBrokerConfiguration {
```
nmonterroso: I don't think this should live in the `guice` package, since it's not an `AbstractModule`, `Provider`, or anything else that interacts with `guice`

service/job-queue/src/main/java/com/wikia/services/jobqueue/domain/Task.java
```diff
@@ -0,0 +1,48 @@
...
+
+import javax.annotation.Nullable;
+
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+
+public
+@Data
+class Task {
```
nmonterroso: I wonder if this class should expose a method to serialize itself, and then `CeleryTaskFormat` can become `class CeleryTask extends Task`, which would overwrite that serialization method. Then, in `QueueSystemQAO` we can remove the `if/else` around creating the string representation of the `Task` and just call `task->serialize`.

tiffanyperumpail: I think that sounds like a better implementation. I will change that now.

service/job-queue/src/main/java/com/wikia/services/jobqueue/domain/ConnectionBundle.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+
+public
+@Data
+class ConnectionBundle {
```
nmonterroso: Is this class ever json serialized/deserialized? If not, can we remove the `@JsonProperty` tags on the properties?

Also, the word `Bundle` in the class is a bit confusing when used in Dropwizard. I expected it to be an instance of [`Bundle`](https://dropwizard.github.io/dropwizard/manual/core.html#man-core-bundles), so maybe this should be changed to something like `AmqpConnectionProperties` (since it's AMQP specific)

service/job-queue/src/main/java/com/wikia/services/jobqueue/domain/Executor.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.List;
+
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+
+public
+@Data
+class Executor {
```
nmonterroso: maybe this and `CeleryTaskFormat` should move to the `com.wikia.services.jobqueue.domain.celery` package

service/job-queue/src/main/java/com/wikia/services/jobqueue/JobQueueModule.java
```diff
@@ -0,0 +1,39 @@
...
+import com.wikia.services.jobqueue.qao.QueueSystemInterface;
+import com.wikia.services.jobqueue.qao.QueueSystemQAO;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class JobQueueModule extends AbstractModule {
```
nmonterroso: can this be moved to the `guice` package?

service/job-queue/src/main/java/com/wikia/services/jobqueue/qao/MessagingSystemInterface.java
```diff
@@ -0,0 +1,23 @@
+package com.wikia.services.jobqueue.qao;
+
+import com.wikia.services.jobqueue.domain.ConnectionBundle;
+
+import com.rabbitmq.client.ConnectionFactory;
+
+import java.io.IOException;
+import java.util.concurrent.TimeoutException;
+
+public interface MessagingSystemInterface {
```
nmonterroso: It looks like we're declaring this interface and implementing it in `MessagingSystem`, but when we instantiate we use the `MessagingSystem` directly rather than through the interface, so we're not gaining the benefits of [Composition](https://en.wikipedia.org/wiki/Composition_over_inheritance) here.

service/job-queue/src/main/java/com/wikia/services/jobqueue/domain/Task.java
nmonterroso: Another approach would be to keep the `Task` implementations as separate, but change `Task` to be an interface and provide two implementations: `DefaultTask` and `CeleryTask`:

```java
interface Task {
  public String getContentType();
  public String toMessage();
  // other methods
}

class CeleryTask implements Task {
  // celery-specific fields and methods
  public String getContentType() {
    return "application/json"; // through a constant
  }
  public String toMessage() {
    // json serialization stuff
  }
}

class DefaultTask implements Task {
  public String getContentType() {
    // what's the default here?
  }
  public String toMessage() {
    // implementation
  }
}

Comment (nmonterroso) [inline service/job-queue/src/main/java/com/wikia/services/jobqueue/domain/Task.java]: After thinking about it, I think the interface method is better since it allows us to have `CeleryTask` and `DefaultTask` completely independent of one another, so they can evolve independently (if they ever need to)

File: service/job-queue/src/main/java/com/wikia/services/jobqueue/qao/MessagingSystemInterface.java (hunk 6 above)
Comment (nmonterroso) [inline service/job-queue/src/main/java/com/wikia/services/jobqueue/qao/MessagingSystemInterface.java]: So maybe this should be renamed to `MessagingSystem` and the current `MessagingSystem` should be renamed to `AmqpMessagingSystem`

File: service/job-queue/src/main/java/com/wikia/services/jobqueue/qao/QueueSystemInterface.java [hunk 7]
```diff
@@ -0,0 +1,25 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.rabbitmq.client.Channel;
+
+import java.io.IOException;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.concurrent.TimeoutException;
+
+public interface QueueSystemInterface {
```
nmonterroso: See comments about `MessagingSystemInterface`, they also apply here

service/job-queue/src/main/java/com/wikia/services/jobqueue/resources/JobQueueResource.java
```diff
@@ -0,0 +1,112 @@
...
+      @ApiParam(
+          name = "taskList",
+          value = "List of task objects that each include: task type, task arguments, and meta information",
+          required = true)
+      List<Task> taskList
+  ) throws SQLException, IOException, TimeoutException {
+
+    List<String> jobIds = queueSystem.enqueueJobs(queueName, taskList);
+
+    if (jobIds.contains("-1")) {
```
nmonterroso: Can `-1` be a constant somewhere?

service/job-queue/src/main/java/com/wikia/services/jobqueue/qao/MessagingSystemInterface.java
```diff
@@ -0,0 +1,23 @@
...
+  ConnectionBundle openConnection() throws IOException, TimeoutException;
```
nmonterroso: `ConnectionBundle` here is amqp-specific. But since this is only ever used by `QueueSystemQAO`, which is also amqp-specific, that's probably OK.

service/job-queue/src/main/java/com/wikia/services/jobqueue/qao/QueueSystemQAO.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.io.IOException;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.TimeoutException;
+
+public class QueueSystemQAO implements QueueSystemInterface {
```
nmonterroso: I think this can be renamed to `AmqpQueueSystem`

tiffanyperumpail:
@nmonterroso  I've updated the PR with class name changes, removal of the Connector Interface, and addition of Celery-specific queueing methods.  Thanks for all your advice!

service/job-queue/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+mainClassName = "com.wikia.services.jobqueue.JobQueueApplication"
+
+// configure jvm parameters
+applicationDefaultJvmArgs = [
+        "-Xms100m", "-Xmn128m", "-Xmx300m", "-XX:+UseNUMA", "-XX:+UseCondCardMark",
+        "-XX:+UseBiasedLocking", "-XX:+UseConcMarkSweepGC", "-XX:+UseParNewGC"
+]
+
+dependencies {
+    compile project(":lib:swagger")
```
nmonterroso: `:lib:swagger` is now included by default, so you don't need to add it

service/job-queue/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    compile project(':lib:dropwizard:databases:db-jooq')
```
nmonterroso: does this use any database connections? if not can we remove this?

service/job-queue/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    compile "com.rabbitmq:amqp-client:3.5.3"
+}
+
+deployments {
+    production {
+        resources {
+            instances = 2
+        }
+        environment {
+            CONSUL_TAGS = '["production"]'
```
nmonterroso: can you add the `"swagger"` tag to this list?

service/job-queue/src/main/java/com/wikia/services/jobqueue/JobQueueApplication.java
```diff
@@ -0,0 +1,56 @@
...
+public class JobQueueApplication extends ConsulGuiceApplication<JobQueueConfiguration> {
+
+  public static void main(String[] args) throws Exception {
+    new JobQueueApplication().run(args);
+  }
+
+  @Override
+  public List<Class<? extends FeatureInstaller>> getAdditionalFeatureInstallers() {
+    return ImmutableList.<Class<? extends FeatureInstaller>>builder()
+        .add(JooqInstaller.class)
```
nmonterroso: if we don't use database connections in this service we can remove this

service/job-queue/src/main/java/com/wikia/services/jobqueue/qao/AmqpQueueSystem.java
```diff
@@ -0,0 +1,111 @@
...
+  public String enqueueJob(String queueName, Task task)
+      throws IOException, TimeoutException {
+    String jobId;
+
+    try {
+      AMQPConnectionProperties AMQPConnectionProperties = connector.open();
+
+      jobId = sendJob(queueName, task, AMQPConnectionProperties.getChannel());
+
+      if (!jobId.equals(JobQueueConstants.QUEUE_ERROR_ID)) {
```
nmonterroso: I think we always close the connection, regardless of whether or not the task queued successfully or not, so we don't leak connections. Maybe we should use [`try-with-resources`](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)

service/job-queue/src/main/java/com/wikia/services/jobqueue/qao/AmqpQueueSystem.java
```diff
@@ -0,0 +1,111 @@
...
+    List<String> jobIdList = new ArrayList<String>();
+
+    try {
+      AMQPConnectionProperties AMQPConnectionProperties = connector.open();
+
+      for (Task task : taskList) {
+          jobIdList.add(sendJob(queueName, task, AMQPConnectionProperties.getChannel()));
+      }
+
+      if (!jobIdList.contains(JobQueueConstants.QUEUE_ERROR_ID)) {
```
nmonterroso: same reasoning for connections applies here.

nmonterroso: Actually, since this is pretty much duplicated code, can the single queue and multiple queue share the same code?

service/job-queue/src/main/java/com/wikia/services/jobqueue/resources/JobQueueConstants.java
```diff
@@ -0,0 +1,6 @@
+package com.wikia.services.jobqueue.resources;
+
+public class JobQueueConstants {
+  public static final String FAILED_CONNECTION_ID = "-1";
```
nmonterroso: is there a reason for these to be strings?

service/job-queue/src/main/resources/job-queue.yaml
```diff
@@ -0,0 +1,51 @@
...
+        appname: pandora-job-queue
+metrics:
+  reporters:
+    - type: influxdb
+      host: ${consul:INFLUXDB_HOST:127.0.0.1}
+      port: ${consul:INFLUXDB_PORT:5551}
+      prefix: job-queue
+      frequency: 5 seconds
+      includes:
+        - com\.wikia\.jobqueue\.resources\..*
```
nmonterroso: hm i think this needs to be `com\.wikia\.services\.jobqueue.*`

nmonterroso:
some final comments otherwise looks good to me!

service/job-queue/src/main/java/com/wikia/services/jobqueue/resources/JobQueueConstants.java
tiffanyperumpail: If a task if queued without error, it's UUID is a String, so I stored the errorID's as strings as well. Is there a benefit to making the errorIDs of type int?

nmonterroso: Ah ok, that's fine

PR #269, opened 2015-06-23
[DIS] DIS-75 Refactor DB Access

jimmordino:
Refactored DIS DB access to lay the foundation for dynamic routing of requests to the correct shard (database) and instance (slave vs. master).  Implemented and use core DB classes as much as is practical for MyBatis.

- [x] Added core classes for a MyBatis implementation of db-base
- [x] Used the above classes to store DB configurations from the yaml file and add DB health checks
- [x] Added a shard-aware MyBatis guice module that may be pushed up to core eventually
- [x] Moved DIS-specific MyBatis configuration to a subclass of the above module
- [x] Created a first cut of our own @Transaction annotation and method interceptor
- [x] Coded the method interceptor to just use the default DB for now (ignore site id sharding)
- [x] Refactored all services to use a DAO factory to look up their DAOs (needed for dynamic shard routing)
- [x] Verified all existing tests pass
- [x] Added integration tests that verify proper configuration and health check registration


service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/TransactionalMethodInterceptor.java
```diff
@@ -0,0 +1,246 @@
...
+            }
+
+            sqlSessionManager.rollback(true);
+          } else {
+            if (log.isDebugEnabled()) {
+              log.debug(debugPrefix
+                        + " - SqlSession of thread: "
+                        + currentThread().getId()
+                        + " committing");
+            }
```
jogura: Might want to consider removing the `log.isDebugEnabled()` and use `String.format()` to reduce the overhead of concating strings. This was agreed upon as a coding standard unless it is proven to be a huge performance overhead.

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/TransactionalMethodInterceptor.java
```diff
@@ -0,0 +1,246 @@
...
+          interceptedMethod.getDeclaringClass().getAnnotation(Transactional.class);
+    }
+
+    String debugPrefix = null;
+    if (this.log.isDebugEnabled()) {
+      debugPrefix = format("[Intercepted method: %s]", interceptedMethod.toGenericString());
+    }
+
+    // Temporary - just use the default session
+    SqlSessionManager sqlSessionManager = sessionManagerMap.get("shard1-db-master");
```
jogura: Should `"shard1-db-master"` be a constant? In the future how will the name be passed in?

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/TransactionalMethodInterceptor.java
```diff
@@ -0,0 +1,246 @@
...
+  /**
+   * The {@code SqlSessionManager} reference.
+   */
+  @Inject
+  @Named("sessionManagerMap")
+  private Map<String, SqlSessionManager> sessionManagerMap;
+
+  @Inject
+  @Named("siteShardMap")
+  private Map<Integer, Map<String, Object>> siteShardMap;
```
jogura: Where and how is this going to be used?

service/discussion/src/main/java/com/wikia/discussionservice/enums/DeploymentStage.java
```diff
@@ -19,7 +19,7 @@
 public enum DeploymentStage {
   PRODUCTION("prod", ProdDiscussionModule.class),
   QA("qa", QaDiscussionModule.class),
-  TEST("testing", TestDiscussionModule.class);
+  TEST("testing", ProdDiscussionModule.class);
```
jogura: Can you explain the purpose of this change?

service/discussion/src/main/resources/discussion.yaml
```diff
@@ -33,20 +33,45 @@ swaggerConfiguration:
...
+    shardId: 1
+  26337:
+    name: Glee
+    shardId: 1
+databases:
+  configurations:
+    shard1:
+      master:
+        driverClass: com.mysql.jdbc.Driver
+        url: ${consul:DB_URL:"jdbc:mysql://localhost:3306/discussion"}
```
jogura: How will this work with consul? Will we have shard specific `consul:DB_URL`? Same question for the secondary dbs.

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/TransactionalMethodInterceptor.java
jimmordino: It won't - see the comment above.  When we implement DIS-90 we'll look at the argument list of the ```MethodInvocation```, pull out the site ID, use that as a key to look up the shard ID from the ```siteShardMap```, then use the value of the ```readOnly``` attribute on the annotation to build the session manager key as ```{shard ID}-db-{master|slave}```.

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/TransactionalMethodInterceptor.java
jimmordino: See my comment below.  The site ID is available to the method interceptor because the method's arguments are available.  That site ID will be used to look up the shard for that site.

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/TransactionalMethodInterceptor.java
jogura: I see, that would be good to document in the code as well, thanks.

service/discussion/src/main/java/com/wikia/discussionservice/enums/DeploymentStage.java
jimmordino: We're not supposed to have multiple modules - they are actually cruft from bad example code.  We're supposed to have single module (i.e. ```ProdDiscussionModule```) that is dynamically configured from Consul.  Any diffs between environments should be stored there.

service/discussion/src/main/resources/discussion.yaml
jimmordino: TBD.  We're probably going to have to explicitly name the shards and master/slave for each different URL - the other information should all be the same.

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/TransactionalMethodInterceptor.java
jimmordino: For certain, when it's implemented it will be documented.

service/discussion/src/main/java/com/wikia/discussionservice/enums/DeploymentStage.java
jogura: Should we clean up the `QaDiscussionModule.class` as well?

jimmordino: Probably, but I think I recall seeing other services with a second module for QA - we should probably understand the reasoning behind this before we consolidate them all.  It's possible that the other services are in the same boat having followed confusing example code, but let's be sure.

service/discussion/src/main/java/com/wikia/discussionservice/dao/mybatis/TransactionalMethodInterceptor.java
jimmordino: Sure, this was copypasta, but worth cleaning up - 8061ba93b526668eb96b2e480f9f3696dc37aba9

jogura:
LGTM

As we talked before, the tests run slower, and you have an idea as to why. We can address that in the next sprint. The tests failures are due to a different project.

PR #271, opened 2015-06-23
DIS-135 Add responseGroup to the top "_links" in response

jogura:
* Adds the `responseGroup` query parameter for all "_links" in the response
* Add mapper and integration tests

@jimmordino @armonr


armonr:
LGTM

PR #276, opened 2015-06-24
pin to a specific version of the swagger json file

nmonterroso:
@Wikia/services-team @jogura 
https://wikia-inc.atlassian.net/browse/SERVICES-466

Pin to a specific (currently latest) version of the nirvana api swagger json, so that a change in the mediawiki spec will not cause pandora to fail. It now becomes the responsibility of someone wanting to implement a new mediawiki api in pandora to update the pinned version.


sqreek:
LGTM

PR #278, opened 2015-06-25
DIS-137 (WIP) reduce connection pool init size from 10 (default) to 2

jimmordino:
This is for integration tests only - reduce pool sizes.

@armonr @jogura


armonr:
:shipit:

PR #283, opened 2015-06-29
DIS-146 Add upvoteCount to thread and post responses

jogura:
* Update mappers and domain objects
* Minor checkstyle fixes
* Update tests and fixtures

@armonr @jimmordino


service/discussion/src/test/resources/fixtures/createThread.json
```diff
@@ -10,7 +10,7 @@
   "body": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed ",
   "lastEditorId": 1,
   "deletedPosition": 0,
-  "upVotes": 0,
+  "upvoteCount": 0,
```
armonr: Should we add this to `createPost` fixture as well?

armonr:
LGTM

service/discussion/src/test/resources/fixtures/createThread.json
jogura: It's not required to create a post or a reply. But since it was in the file, I had to change to reflect the name changes.

PR #284, opened 2015-06-29
DIS-142 Implementation for upvoting. upvote info in user response to be done separately.

armonr:
/CC @jimmordino @jogura 

* Add table vote	90ca8e7
* Add Vote object in domain	88d5e32
* Add API resource + service + DAO for votes	 2b8a37a
* Add next_id trigger for vote table	02ab7a1
* Register VoteDao and VoteService	353679a
* Update Vote service to return related content, 	53cc821
* Update ContentType - add id, label -> name	e0f62a5
* Remove unused imports of ContentType	2c7097c
* Add content_type table, add FK to tables vote & content_read, 	 19ae58f
* Implement isPostOwner	4906466
* Add VOTE_POST to SecureActivity	 9a904f5
* Anon or banned cannot vote	f6aeb33
* conditional insertion for creating upvote	 7aad5c6
* DIS-142 Add `posts:vote` permission to user role	 51aa8e5
* DIS-148 Use Unauthorized reponse code instead of 403 	 8ea81c1
* Forbidden response on re-upvote request	 587192a
* DIS-146 Add upvoteCount to thread and post responses 	31b2481
* DIS-146 Add upvoteCount to thread and post responses    8d8d6cc

https://wikia-inc.atlassian.net/browse/DIS-142


service/discussion/sql/create_discussion_db.sql
```diff
@@ -461,6 +514,20 @@ CREATE PROCEDURE `update_post_stats` (IN siteId INTEGER, IN postId BIGINT)
...
+    DECLARE upVoteCount INT DEFAULT 0;
+
+    SELECT COUNT(*) INTO upVoteCount FROM vote
+    WHERE vote.content_type = contentType AND vote.content_id = contentId AND vote.value = 1;
+
+    IF contentType = (SELECT id FROM content_type WHERE `name` = 'forum post') THEN
+      UPDATE post SET upvote_count = upVoteCount WHERE post.id = contentId;
+    END IF;
+  END
+$$
```
jogura: Do we need a `SELECT FOR UPDATE` statement to lock the post record?

service/discussion/src/main/java/com/wikia/discussionservice/domain/Vote.java
```diff
@@ -0,0 +1,32 @@
...
+  private Long contentId;
+
+  @JsonProperty
+  private ContentType contentType;
+
+  @JsonProperty
+  private Integer creatorId;
+
+  @JsonProperty
+  private Long creationDate;
```
jogura: We are using `Instant` everywhere else for `creationDate`

service/discussion/src/main/java/com/wikia/discussionservice/domain/hal/input/PostInput.java
```diff
@@ -4,6 +4,7 @@
 import com.wikia.discussionservice.domain.PostRevision;
 
 import com.wordnik.swagger.annotations.ApiModelProperty;
+import org.codehaus.jackson.annotate.JsonIgnoreProperties;
```
jogura: I think I introduced this and it is unnecessary. :frowning:

service/discussion/src/main/java/com/wikia/discussionservice/exceptions/UnauthorizedExceptionMapper.java
```diff
@@ -23,12 +20,8 @@
...
   @Override
   public Response toResponse(UnauthorizedException exception) {
-    DiscussionServiceApplication.LOGGER.info("403 Response - " + exception.getMessage());
-    return StandardResponses.Forbidden(
-        Problem.builder()
-            .title("Forbidden")
-            .instance(URI.create(request.getRequestURI()))
-            .build());
+    DiscussionServiceApplication.LOGGER.info("401 Response - " + exception.getMessage());
+    return StandardResponses.Unauthorized("Unauthorized. " + exception.getMessage());
```
jogura: Do we not want to return the `URI` that was rejected?

service/discussion/sql/create_discussion_db.sql
```diff
@@ -461,6 +514,22 @@ CREATE PROCEDURE `update_post_stats` (IN siteId INTEGER, IN postId BIGINT)
...
+  BEGIN
+    DECLARE upVoteCount INT DEFAULT 0;
+
+    SELECT COUNT(*) INTO upVoteCount FROM vote
+    WHERE vote.content_type = contentType AND vote.content_id = contentId AND vote.value = 1;
+
+    IF contentType = (SELECT id FROM content_type WHERE `name` = 'forum post') THEN
+      -- lock
+      SELECT upvote_count FROM post WHERE id = contentId FOR UPDATE;
+      UPDATE post SET upvote_count = upVoteCount WHERE post.id = contentId;
```
jimmordino: This won't work.  You must lock *before* you select into ```upVoteCount``` or you won't get an accurate count.  A different user can change the count in between the time you select it and the time you lock.  Lock, select into, update - that's the sequence.

service/discussion/src/main/java/com/wikia/discussionservice/services/VoteService.java
```diff
@@ -0,0 +1,88 @@
...
+    SecureActivity.VOTE_POST.authorize(user, siteId, contentId);
+
+    VoteDao voteDao = daoFactory.getDao(VoteDao.class);
+    Integer numberOfRowsInserted =
+        voteDao.createUpVote(user.getUserId(), contentType.getId(), contentId);
+    if (numberOfRowsInserted == 1) {
+      voteDao.updateContentVotes(contentType.getId(), contentId);
+      return postService.getPost(siteId, contentId);
+    } else if (numberOfRowsInserted == 0) {
+      throw new ForbiddenException("User has already upVoted this post");
```
jogura: Is this always true? What if the `postId` is bogus and doesn't exist?

service/discussion/src/main/java/com/wikia/discussionservice/services/VoteService.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    return Optional.empty();
```
jogura: Will this only get here if the `numberOfRowsInserted` is not `0` or `1`? If it is greater than `1` then something unexpected has happened.

armonr: The query is a single-row insertion. I don't see how we can get 2 or more.

jimmordino: The insert will only insert 1, but if you want to be sure you can put a `LIMIT 1` in the appropriate place in the insert query.  Bear in mind that if you do, MySQL will start complaining of replication because it always warns about replicating any inserts or updates with a `LIMIT` clause.

service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
```diff
@@ -33,6 +36,16 @@
       throw new UnauthorizedException("This user is not authorized to update the thread.");
     };
 
+  public static final Authorizable authorizeUpVotePost = (user, siteId, postId) -> {
+      if (user.isAnonymous() || user.isBanned()) {
+        throw new UnauthorizedException("This user is not authorized to vote on posts.");
+      }
+      if (postService.isPostOwner(siteId, user.getUserId(), postId)) { // NOPMD
+        throw new UnauthorizedException("This user cannot vote on own posts.");
```
jogura: Add log messages for the different failed attempts.

service/discussion/src/main/java/com/wikia/discussionservice/services/PostService.java
```diff
@@ -281,4 +282,21 @@ private PostRevision updatePostRevisionId(Post post, PostRevision postRevision)
...
+   * @param siteId Integer
+   * @param userId Integer
+   * @param contentId Long
+   * @return Boolean
+   */
+  public boolean isPostOwner(Integer siteId, Integer userId, Long contentId) {
+    PostDao postDao = daoFactory.getDao(PostDao.class);
+    Post post = postDao.getPost(siteId, contentId);
+    if (post == null) {
+      throw new NotFoundException("Post not found.");
```
jogura: Log this scenario

jogura:
Other than my few minor comments, LGTM. 

It looks like test failures are caused by another service.

service/discussion/src/main/java/com/wikia/discussionservice/domain/Vote.java
```diff
@@ -0,0 +1,34 @@
...
+  private ContentType contentType;
+
+  @JsonProperty
+  private Integer creatorId;
+
+  @JsonProperty
+  private Instant creationDate;
+
+  @JsonProperty
+  private Byte value;
```
jogura: Why is this a `Byte`?

armonr: `Byte` in Java is the equivalent of `TINYINT` in mysql

PR #290, opened 2015-07-02
DIS-142 Decouple javax.ws exceptions with Service layer

jogura:
* Introduces specific DiscussionService exceptions that will be translated into the proper HTTP responses
* Adds new exception mappers
* Updates method protocols
* Updates tests to expect the new exceptions

@armonr @jimmordino


jimmordino:
Given this, can we remove the throws clauses from service methods?  Will we always be throwing a subclass of DiscussionServiceException once we refactor to use these exception everywhere?

jogura:
>Given this, can we remove the throws clauses from service methods? Will we always be throwing a subclass of DiscussionServiceException once we refactor to use these exception everywhere?

Already done [here](https://github.com/Wikia/pandora/pull/290/files#diff-5eee0cf281e8ad8835fd79cd49ff0fa5L41). Just needed to add the new exception mappers and clean up missed references. 90d853c

jimmordino:
Yes, I'm asking because I've been adding throws clauses as part of my service refactoring - I need to remove those now.

service/discussion/src/main/java/com/wikia/discussionservice/auth/Authorizable.java
```diff
@@ -9,6 +10,7 @@
 @FunctionalInterface
 public interface Authorizable {
 
-  boolean authorize(User user, Integer siteId, Long contentId) throws UnauthorizedException;
+  boolean authorize(User user, Integer siteId, Long contentId)
+      throws ForbiddenException, NotFoundException;
```
armonr: So can this just be `throws DiscussionServiceException`?

armonr:
LGTM

service/discussion/src/main/java/com/wikia/discussionservice/auth/Authorizable.java
jogura: I chose to be explicit.

jogura:
The test failures are outside of the discussion service.

service/discussion/src/main/java/com/wikia/discussionservice/auth/Authorizable.java
jimmordino: :clap:

PR #291, opened 2015-07-02
DIS-142 Properly returns the upvoteCount after a create/delete

jogura:
Uses a Map for the parameters so MyBatis can populate the `OUT` parameter properly.

* Update the dao interface
* Update the VoteService
* Update the tests

@jimmordino @armonr


armonr:
:shipit: (after DIS tests pass)

PR #292, opened 2015-07-03
DIS-159 Add authz tests for UpVote functionality

jogura:
* Test the create and delete functionality for upvotes

@jimmordino @armonr


jogura:
Going to roll up the two PR's into one.

PR #294, opened 2015-07-06
add more instances for discussion service

sqreek:
if one of instance fall down, we still have two more for handle traffic, until mesos restart fallen instance


pchojnacki:
LGTM

PR #295, opened 2015-07-06
Introduction of aggregation service layer + add user upvote data

armonr:
This adds an aggregation service in order to combine more user-specific data to the payload of the response; user upvote data now gets added via this layer.

CC @jogura @jimmordino 

https://wikia-inc.atlassian.net/browse/DIS-164


service/discussion/src/main/java/com/wikia/discussionservice/services/AggregationService.java
```diff
@@ -0,0 +1,68 @@
...
+                            VoteService voteService,
+                            PostRepresentationMapper postMapper,
+                            AggregateRepresentationMapper aggregateMapper) {
+    this.postService = postService;
+    this.voteService = voteService;
+    this.postMapper = postMapper;
+    this.aggregateMapper = aggregateMapper;
+  }
+
+  public Representation postRepresentation(Integer siteId, User user, Long postId, UriInfo uriInfo,
```
jimmordino: please add a verb up front like "get" or "build", assuming this doesn't "post" anything.

service/discussion/src/main/java/com/wikia/discussionservice/services/AggregationService.java
```diff
@@ -0,0 +1,68 @@
...
+
+            return aggregateMapper.addUserSpecifics(uriInfo, siteId, representation, userSpecifics);
+          } else {
+            return representation;
+          }
+        })
+        .orElseThrow(() -> new NotFoundException(
+            String.format("No post found for site id %s with forum id %s", siteId, postId)));
+  }
+}
```
jimmordino: I think we should reserve operations that involve **externa**l services for the AggregationService.  DIS services should just call each other to get this kind of information - aggregating DIS services means separate transactions and all of the associated overhead in addition to the overhead of another round trip.  Let leave this in place for now, but in the future this operation should migrate down to the Post service.

service/discussion/src/main/java/com/wikia/discussionservice/services/VoteService.java
```diff
@@ -82,6 +82,12 @@ private Integer updateUpvoteCount(ContentType contentType, Long contentId, VoteD
     params.put("contentId", contentId);
     params.put("upVoteCount", upVoteCount);
     voteDao.updateContentVotes(params);
-    return (Integer)params.get("upVoteCount");
+    return (Integer) params.get("upVoteCount");
+  }
+
+  public Boolean hasUserUpvoted(Integer siteId, User user, ContentType contentType,
```
jimmordino: This should throw NotFoundException if the specified content has been soft or is otherwise not visible.

service/discussion/src/main/java/com/wikia/discussionservice/resources/PostResource.java
```diff
@@ -109,16 +116,10 @@ public Response getPost(@Context @Auth User user,
     ResponseGroup responseGroup = ResponseGroup.getResponseGroup(requestedResponseGroup);
     Preconditions.checkNotNull(responseGroup, "Invalid response group");
 
-    Optional<Post> optionalPost = postService.getPost(siteId.get(), postId.get());
+    Representation postRepresentation = aggregationService.postRepresentation(
+        siteId.get(), user, postId.get(), uriInfo, responseGroup);
```
jimmordino: This is a problem. IMO.  We should not have HAL at the AggregationService layer.  Remember, we want to be able to throw away REST and go with some other front end for DIS as needed.  The AggregationService should survive that move relatively unscathed.  So the Aggregation service should return domain objects just as any other service would, and the resource layer is the only place we should see HAL.  It should also makes calls to external services (eventually) through an interface, so that if the external service changes from REST to some other access we don't need to change the AggregationService.

service/discussion/src/main/java/com/wikia/discussionservice/services/VoteService.java
armonr: That would cost yet another roundtrip to the database; is it a must?

service/discussion/src/main/java/com/wikia/discussionservice/resources/PostResource.java
armonr: As I understand, we wanted the aggregation service to help us with both external and some internal data fetching/merging; or at least I think this sprint would allow for this much. Sure we can rethink this design for future sprints. Let's discuss this tomorrow.

service/discussion/src/main/java/com/wikia/discussionservice/services/VoteService.java
jimmordino: The truth of it is that this should not be a separate call at all.  The call to fetch posts should optionally (if the user is not anon) left join with vote to return value or zero if the user has not voted.  Fetching the user-specific data should be part of the entity fetch, not separate, IMO.  We can figure out how to separate the user-specific data from the generic data for caching later.

jogura: We actually wanted to make this a separate call for the eventuality of reusing the anon call for caching reasons. We'd only decorate the cached anon call results with user specific. This would depend on the access pattern of the service (more heavily anon vs logged in) and we'd adjust the strategy.

service/discussion/src/main/resources/com/wikia/discussionservice/dao/VoteDao.xml
```diff
@@ -13,6 +13,14 @@
         <result column="value" property="value"/>
     </resultMap>
 
+    <select id="hasUserUpvoted" resultType="java.lang.Boolean">
+        SELECT COUNT(*) = 1
+        FROM vote
+        WHERE content_id = #{contentId}
+        AND content_type = #{contentType}
+        AND created_by = #{creatorId}
+    </select>
```
jogura: Having a bulk API for this may be more optimal to avoid multiple db roundtrips.

jogura:
We need to have the same decoration done with a user requests a thread and all of the posts underneath.

service/discussion/src/main/java/com/wikia/discussionservice/services/VoteService.java
jimmordino: I understand this reasoning, but we don't understand how that cache works yet.  Does it only hold visible posts?  Does it hold deleted and held posts so it can also serve admins?  When a new reply comes in is it appended to the cache or is the cache invalidated?  We should understand how that cache works before we make design decisions about how data is fetched.

jimmordino: Also, mybatis will support separation - if we create a pseudo entity that contains a collection of replies (posts) and a collection of user specific decorations we should be able to populate that pseudo entity with a single round trip.  We can then do the association between the two lists within the service after caching (as needed) the list of replies.

service/discussion/src/main/java/com/wikia/discussionservice/mappers/HalAggregateRepresentationMapper.java
```diff
@@ -0,0 +1,43 @@
...
+  @Override
+  public Representation addUserSpecifics(UriInfo uriInfo, Integer siteId,
+                                         Representation representation,
+                                         HashMap<String, Object> userSpecifics) {
+
+    Representation userSpecificsRepresentation = representationFactory.newRepresentation();
+
+    for (Map.Entry<String, Object> userSpecific : userSpecifics.entrySet()) {
+      userSpecificsRepresentation.withProperty(
+          userSpecific.getKey(), String.valueOf(userSpecific.getValue()));
```
jogura: I don't think we want to unilaterally convert the values to a `String` things like booleans and numbers should be left alone.  `Long`s should be converted because those get rounded.

service/discussion/src/main/java/com/wikia/discussionservice/services/AggregationService.java
```diff
@@ -0,0 +1,72 @@
...
+    return optionalPost
+        .map(post1 -> {
+            Representation representation =
+                postMapper.buildRepresentation(siteId, post1, uriInfo, responseGroup);
+            if (user.getUserId() > 0) {
+              Boolean hasUpvoted = voteService.hasUserUpvoted(siteId, user, ContentType.FORUM_POST,
+                                                              postId);
+
+              HashMap<String, Object> userSpecifics = new HashMap<>();
+              userSpecifics.put("hasUpvoted", String.valueOf(hasUpvoted));
```
jogura: The `boolean` value should be left alone, IMO, not converted to a `String`

jimmordino:
Assuming there's nothing in this branch that hasn't already been reviewed as part of a subtask review,
LGTM

jogura:
:+1:

PR #296, opened 2015-07-07
[Do not Merge] proof of concept add tag to logs

sqreek:
For add information about instance that sending information to influxdb


sqreek:
ping @Wikia/services-team @harnash

pchojnacki:
Good Idea but if its possible please try to merge your changes to upstream source first instead of importing foreign source code.

sqreek:
Question: Is upstream is still foreign? becuse most of code is written by @nmonterroso  :)
Maybe we want to take care about it as fork in Wikia repositories?

pchojnacki:
Im talking about this: https://github.com/influxdb/influxdb-java and the metrics sending file you overshadowed and not about dropwizard 'plugin' that is indeed Nelsons doing.

sqreek:
actually we are using https://github.com/davidB/metrics-influxdb

ArturKlajnerok:
@sqreek shouldn't we switch to official and actively developed: https://github.com/influxdb/influxdb-java as https://github.com/davidB/metrics-influxdb seems to not be actively maintained ?

sqreek:
I was looked at that working on issue, in this case, we need to rewrite reporter and use writed from us. In metrics we had that, and we add only small change.
So, yes, we can switch it, but it will be a little more work with this,

PR #297, opened 2015-07-07
DIS-169 Handle userId of `0` as an anonymous request

jogura:
If a `0` is explicitly passed in then treat the user as anonymous.

@jimmordino @armonr


jimmordino:
:+1:  for the code, but those are discussion test failures

jogura:
I cherrypicked a change from another branch that addresses the failed test and rerunning the suite.

PR #298, opened 2015-07-08
DIS-170/174 If single representation is present then write out single element otherwise a list

jogura:
* Hal RepresentationFactory has the ability to configure if we want this behavior by using the `COALESCE_ARRAYS` flag. Set this flag in a guice provider method.
* Remove the stringification of the vote count
* Minor findbug fixes
* Update tests to reflect change

@jimmordino @armonr


armonr:
LGTM

armonr:
LGTM

PR #303, opened 2015-07-08
DIS-170/174 Reverting the coalescing of single element arrays

jogura:
After weighing the pro's and con's of using the coalescing single element arrays we've decided it was better to be consistent across the API.

@armonr @jimmordino


armonr:
:shipit:

PR #307, opened 2015-07-09
Improved exception handling in unit tests for static assets service

jcellary:
As pointed out by @ludwikkazmierczak unit test functions where declared as throws Exception - changed that to IOException and deleted them in some cases.


ludwikkazmierczak:
LGTM

PR #308, opened 2015-07-09
DIS-166 Add the `requesterId` to all `GET` calls

jogura:
To aide clients in cases where the authn session may have expired, the response with contain a `requesterId` so clients can verify/validate that the user's id is what the client expected it to be. If there is a mismatch then the clients can prompt the user to authenticate again.

@jimmordino @armonr


jimmordino:
:+1:  Seems straightforward

service/discussion/src/main/java/com/wikia/discussionservice/mappers/HalAggregateRepresentationMapper.java
```diff
@@ -0,0 +1,45 @@
+package com.wikia.discussionservice.mappers;
```
armonr: This files should be deleted. I think it came back in a merge

service/discussion/src/test/java/com/wikia/discussionservice/mappers/HalAggregateRepresentationMapperTest.java
```diff
@@ -0,0 +1,76 @@
+package com.wikia.discussionservice.mappers;
```
armonr: Same with this file. should be removed

armonr: https://github.com/Wikia/pandora/pull/309

PR #310, opened 2015-07-09
Add sorting function for replies

armonr:
And make the default "newest first"

https://wikia-inc.atlassian.net/browse/DIS-143


jogura:
Should we also change the default at the [ThreadService](https://github.com/Wikia/pandora/blob/DIS-143/service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java#L37-37)?

jogura:
:+1:

PR #312, opened 2015-07-09
DIS-166 Add requesterId to Upvote response

jogura:
Missed adding the `requesterId` to the Upvote response

* Reorder argument parameters to be consistent with other HAL mappers
* Minor checkstyles changes
* Add UserFactory to VoteResourceTest

@jimmordino @armonr


jimmordino:
:+1:

PR #313, opened 2015-07-09
[DIS] DIS-175 Add user data to relevant entity queries and push up through responses

jimmordino:
Fetches upvote data as part of the relevant forum, thread/post, and post/reply queries.  Adds embedded user data to responses.  

@armonr @jogura


service/discussion/src/main/java/com/wikia/discussionservice/dao/DiscussionThreadDao.java
```diff
@@ -9,6 +9,7 @@
 public interface DiscussionThreadDao<T extends DiscussionThread> {
 
   T getThread(@Param("siteId") Integer siteId, @Param("threadId") Long threadId,
+              @Param("userId") Integer userId,
```
armonr: Added [DIS-182](https://wikia-inc.atlassian.net/browse/DIS-182) to update the parameter orders everywhere.

service/discussion/src/main/java/com/wikia/discussionservice/services/ThreadService.java
```diff
@@ -67,8 +72,31 @@ public ThreadService(DaoFactory daoFactory, PostService postService) {
                                                                  SortDirection sortDir) {
     // TODO: Check user permissions
     DiscussionThreadDao threadDao = daoFactory.getDao(DiscussionThreadDao.class);
-    return Optional.ofNullable((T) threadDao.getThread(siteId, threadId, limit, limit * page,
-                                                       sortKey.name(), sortDir.name()));
+    return Optional.ofNullable((T) threadDao.getThread(siteId, threadId, user.getUserId(),
+                                                       limit, limit * page,
+                                                       sortKey.name(), sortDir.name()))
+        .map(thread -> {
+            if (!user.isAnonymous()) {
```
armonr: Will `isAnonymous` work as expected right now, or should it check the id explicitly?

jimmordino: I think the concern with ```isAnonymous``` is with regard to testing - the role that it checks isn't always set up unless you come through the servlet layer.  @jogura is going to fix that, but I think this method is fine for production and at the moment it's not affecting our tests.

service/discussion/src/main/java/com/wikia/discussionservice/domain/DiscussionThread.java
```diff
@@ -81,6 +80,13 @@ public Integer getUpvoteCount() {
     return firstPost.getUpvoteCount();
   }
 
+  public Boolean hasUserUpvoted() {
+    return firstPost.getUserData().getHasUpvoted();
+  }
+
+  @JsonProperty
+  private List<PostUserData> postUserDataList;
```
armonr: Why is this needed at all?

service/discussion/src/main/java/com/wikia/discussionservice/dao/DiscussionThreadDao.java
jimmordino: Fixed in 8a0be32f62b9b2bde3422d27eafc5df4ba495881

service/discussion/src/main/java/com/wikia/discussionservice/domain/DiscussionThread.java
jimmordino: When we fetch the user data from the db we need to store it somewhere at this level.  The query is thread-driven and it fetches the thread, first post, last post, and child posts.  This is the correct place to also fetch the user data per child post - it melds in well with the existing query.  So the user data is stored in this collection when the thread is returned from MyBatis.  We keep it separate because we'll probably cache the thread, posts, or both in memcache at some point and we want to cache anonymous data only.  We then move the user data from this collection into and underneath the post data (first, last, children).

service/discussion/src/test/java/com/wikia/discussionservice/resources/VoteHelper.java
```diff
@@ -0,0 +1,77 @@
...
+
+public class VoteHelper {
+
+  public static Integer performUpvote(DropwizardAppRule<DiscussionServiceConfiguration> rule,
+                                      Client client,
+                                      Integer siteId, Integer userId, Long postId,
+                                      Integer[] expectedCodes) throws Exception {
+    UriBuilder uriBuilder = getVoteResourceUrl(rule, siteId, postId);
+    MultivaluedMap<String, Object> headerMap = new MultivaluedHashMap<>();
+    if (userId != null) {
```
armonr: minor: also check for `> 0`?

armonr:
:ship: it!

service/discussion/src/test/java/com/wikia/discussionservice/resources/VoteHelper.java
jimmordino: Good call. 84417f1ac4d106fb7f0ed381080fefee116a0134

PR #321, opened 2015-07-13
Whitelist Regexes

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-502

Introduces a `Whitelist` class which compares inputs to both a list of string literals and regular expressions, so that dynamic preferences (ex: `some-pref-$wikiId`) can be `set` and are returned as part of the `get`.


service/user-preference/src/main/java/com/wikia/userpreference/domain/Whitelist.java
```diff
@@ -0,0 +1,71 @@
...
+  public void setLiterals(List<String> literals) {
+    this.literals = Optional.ofNullable(literals);
+  }
+
+  public List<String> getRegexes() {
+    return regexes.orElse(new ArrayList<>());
+  }
+
+  public void setRegexes(List<String> regexes) {
+    this.regexes = Optional.ofNullable(regexes);
```
sqreek: As far as I know, it will be setup once. Maybe consider to merge regexes into one regex pattern?

nmonterroso: That's a good idea. I'll play around with that.

service/user-preference/src/main/resources/user-preference.yaml
```diff
@@ -55,4 +55,132 @@ databases:
...
+    - .*-toolbar-contents$
+    - ^EditorMainPageNotificationHidden_.*
+    - ^adoptionemails-.*
+    - ^forum_sort_.*
+    - ^founderemails-complete-digest-.*
+    - ^founderemails-edits-.*
+    - ^founderemails-joins-.*
+    - ^founderemails-views-digest-.*
+    - ^searchNs.*
+    - ^wall_sort_.*
```
ArturKlajnerok: I think that, it would be good to not use the wildcard but a defined scope for all the regexes.

service/user-preference/src/main/java/com/wikia/userpreference/domain/Whitelist.java
nmonterroso: da6a1d5

service/user-preference/src/main/resources/user-preference.yaml
nmonterroso: good call. i made them a bit more specific in 580a0fe, but the spreadsheet doesn't have details for all of the regexes, so i left those as `.*`

ArturKlajnerok:
Lgtm :+1:

PR #323, opened 2015-07-13
DIS-204 Implement Trending Sort in SQL

jogura:
* Add an additional `case/when` for `trending`
* Add an additional `thread_trending_score` column (that is computed) on the threads query
* Add `trendingScore` in HalThreadMapper/ThreadResponse
* Add field in HalThreadMapperTest

@jimmordino @armonr


jimmordino:
LGTM

PR #324, opened 2015-07-13
JobQueue Delayed Queuing 

tiffanyperumpail:
@Wikia/services-team https://wikia-inc.atlassian.net/browse/SERVICES-378

~~Don't review this code at this time. I improperly merged and deleted all of the classes that read / wrote to the jobqueue_service database. I need to fix this first.~~

This PR supports queueing tasks for delayed execution to the job queue. This PR also merges the "queueCeleryJob" and "queueJob" methods. This PR also supports deduplication of tasks when the a Task's `deduplicationFlag` is set to `true.`


service/job-queue/src/main/java/com/wikia/services/jobqueue/dao/AmqpQueueSystemDAO.java
```diff
@@ -0,0 +1,157 @@
...
+  public static final String DB_NAME = JOBQUEUE_SERVICE.getName();
+
+  private JooqConfigs database;
+  private RabbitMQConfiguration rabbitMQConfig;
+  private HashMap<String, Integer> validDelayTimes;
+
+  public AmqpQueueSystemDAO(
+      JooqConfigs database,
+      RabbitMQConfiguration config)
+      throws IOException {
```
nmonterroso: I don't think a constructor should have the possibility to `throw` an exception. Can `validDelayTimes` be provided by the applications `Module`?

service/job-queue/src/main/java/com/wikia/services/jobqueue/dao/AmqpQueueSystemDAO.java
```diff
@@ -0,0 +1,157 @@
...
+    this.rabbitMQConfig = config;
+
+    ObjectMapper mapper = new ObjectMapper();
+    String validDelayTimesJsonString = rabbitMQConfig.getValidDelayTimes();
+    this.validDelayTimes = mapper.readValue(validDelayTimesJsonString, new TypeReference<HashMap<String, Integer>>(){});
+  }
+
+  public void addTaskToDatabase(Task task) {
+
+    SimpleDateFormat dateFormat = new SimpleDateFormat("dd-MMM-yyyy HH:mm:ss");
```
nmonterroso: looks like we use the same date format below - can the string format be a `static final` of the class?

service/job-queue/src/main/java/com/wikia/services/jobqueue/dao/AmqpQueueSystemDAO.java
```diff
@@ -0,0 +1,157 @@
...
+    DateTime dateQueued = new DateTime();
+    dateQueued = dateQueued.parse(r.getValue(JOB_STATUSES.DATE_QUEUED), DateTimeFormat.forPattern("dd-MMM-yyyy HH:mm:ss"));
+
+    Integer delayTimeInMiliseconds = validDelayTimes.get(r.getValue(JOB_STATUSES.DELAY_TIME));
+    DateTime executionDate = dateQueued.plusMillis(delayTimeInMiliseconds);
+
+    DateTime currentTime = new DateTime(DateTimeZone.UTC);
+
+    if (executionDate.isBefore(currentTime)) {
+      database.dslForReads(DB_NAME)
```
nmonterroso: if a query is doing a delete it should use `dslForWrites`

service/job-queue/src/main/java/com/wikia/services/jobqueue/dao/AmqpQueueSystemDAO.java
```diff
@@ -0,0 +1,157 @@
...
+        .select()
+        .from(JOB_STATUSES)
+        .where(JOB_STATUSES.UNIQUE_ID.eq(uniqueId));
+
+    for (Record r : statement) {
+      deleteIfExecutedTask(r);
+    }
+  }
+
+  public void deleteIfExecutedTask(Record r) {
```
nmonterroso: I wonder if this method is really needed. I think we don't need to call this on every dedup call, we can take care of that in the SQL statement, and we can let the deletion mechanism take care of removing records

service/job-queue/src/main/java/com/wikia/services/jobqueue/dao/AmqpQueueSystemDAO.java
```diff
@@ -0,0 +1,157 @@
...
+    updateJobStatusforUniqueId(uniqueTaskId);
+
+    SelectJoinStep statement =  database.dslForReads(DB_NAME)
+        .select()
+        .from(JOB_STATUSES);
+
+    return statement.fetch(JOB_STATUSES.UNIQUE_ID).contains(uniqueTaskId);
+  }
+
+  public void updateAllJobStatuses()
```
nmonterroso: I think this method should be called `deleteDelayedTasks()` and it should do the `delete` directly (limited by delay time), rather than iterating over many tasks and checking them individually

service/job-queue/src/main/java/com/wikia/services/jobqueue/domain/Task.java
```diff
@@ -1,23 +1,84 @@
...
+  @JsonProperty
+  private String uniqueId;
+
+  @Nullable
+  @JsonProperty
+  private String hashId;
+
+  @Nullable
+  @JsonProperty
+  private boolean deduplicateFlag;
```
nmonterroso: i think this can just be called `deduplicate`, since a `boolean` is already an indicator of a `flag`

service/job-queue/src/main/java/com/wikia/services/jobqueue/exception/DuplicateTaskException.java
```diff
@@ -0,0 +1,7 @@
+package com.wikia.services.jobqueue.exception;
+
+public class DuplicateTaskException extends Exception{
+  public DuplicateTaskException(String message) {
```
nmonterroso: should this be:

```java
public DuplicateTaskException(String message, String taskId)
```

where `taskId` is the id of the task that's already queued?

service/job-queue/src/main/java/com/wikia/services/jobqueue/qao/AMQPDelayQueueCreator.java
```diff
@@ -0,0 +1,68 @@
...
+  private AmqpConnector connector;
+  private RabbitMQConfiguration rabbitMQConfig;
+  private HashMap<String, Integer> validDelayTimes;
+  private int delayQueueExpiration;
+
+  @Inject
+  public AMQPDelayQueueCreator(
+      RabbitMQConfiguration config,
+      AmqpConnector connector)
+      throws IOException {
```
nmonterroso: Same comment as above - now that the same logic is duplicated, let's provide the delay times in the `Module`

service/job-queue/src/main/java/com/wikia/services/jobqueue/qao/AmqpQueueSystem.java
```diff
@@ -21,65 +28,87 @@
...
-  public String enqueueJob(String queueName, Task task)
-      throws IOException, TimeoutException {
-    String jobId;
+  public boolean isJobQueued(String uniqueTaskId)
+      throws IOException, InvalidDelayTimeException {
+    return amqpQueueSystemDAO.isQueued(uniqueTaskId);
+  }
+
+  @Override
+  public QueueResponse enqueueJob(String queueName, Task task)
```
nmonterroso: I think that this function will leak connections if an exception is throws before `connector.close` is called.

service/job-queue/src/test/java/com/wikia/services/jobqueue/resources/JobQueueResourceIntegrationTest.java
```diff
@@ -56,11 +79,46 @@
...
+    rabbitMQConfig = getConfigFromRule();
+    database = getDataSourceFromRule();
+
+    AmqpQueueSystemDAO amqpQueueSystemDAO = new AmqpQueueSystemDAO(database, rabbitMQConfig);
+    amqpConnector = new AmqpConnector(rabbitMQConfig);
+    amqpQueueSystem = new AmqpQueueSystem(amqpConnector, amqpQueueSystemDAO, rabbitMQConfig);
+  }
+
+  @After
+  public void tearDown() {
```
nmonterroso: I think that the possibility exists here for one person's tests to delete data that another person's tests depends on. Maybe we should create the tasks with a specific id and then delete those tasks afterwards?

nmonterroso:
I think the `ValidDelayTimesProvider` class isn't used. If so, we should remove it. Otherwise LGTM

PR #325, opened 2015-07-13
SOC-787 Add authentication to write routes of User Attribute Service

jsutterfield:
This PR adds authentication to the write routes of the User Attributes (PUTs, POSTs, and DELETEs). GETs remain unauthenticated.

It emulates the approach taken by the User Preferences Services in that it simply verifies that the value of the X-Wikia-UserId header matches the user id of the user we want to modify. An explanation of the request routing and authentication handling can be found [here](https://github.com/Wikia/pandora/wiki/Service-Request-Routing). Essentially all requests exist behind the API-Gateway which forwards authentication requests to Helios. Helios will validate the credentials, and if successful sets the X-Wikia-UserId header with the user's id. This request is then sent back to the API-Gateway which proxies it to the service. Since the service has to be accessed through the gateway, we know that if the request has an X-Wikia-UserId header, it's authenticated.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-787


service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/AttributeResource.java
```diff
@@ -77,57 +72,4 @@ public Response getUserAttributeByName(@PathParam("name") String name)
         .orElseGet(() -> StandardResponses
             .NotFound(String.format("No attribute found with name %s", name)));
   }
-
-  @POST
-  public Response createAttribute(
```
jsutterfield: The ability to create, edit, and delete attributes aren't part of the product requirements for the MVP and should probably be restricted to a certain set of clients if/when we have a use case for it.

Rather than deal with the complexity of whitelisting certain users to allow them to perform these operations (which affects the data of many users), I'm just going to remove them for now until we have a specific case for it. These operations can be done directly in the db on an ad-hoc basis if need be.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -95,17 +99,17 @@ public Response getAttributeForUser(
...
-  // TODO, this doesn't do any checking that the user actually exists, make sure this is
-  // handled when we add the auth library
   @POST
   @Path("/attr")
   public Response createAttributeForUser(
+      @Auth AuthenticatedUser user,
       @PathParam("userID") int userID,
       @FormParam("attrName") String attrName,
       @FormParam("value") String value
   ) throws java.sql.SQLException {
```
mklucsarits: These functions can throw WebApplicationException now. Should that be added here?

jsutterfield: Nope, since this is a run time exception we don't have to declare it in the method signature: http://stackoverflow.com/questions/583973/jax-rs-jersey-how-to-customize-error-handling#answer-599131

Also, in the Jersey examples (as well as the User Preference Service), the throws is not included: https://jersey.java.net/documentation/latest/representations.html (Section 7.3, there aren't any ids I can link to).

mklucsarits: OK, I see.

mklucsarits:
LGTM

service/user-attribute/src/main/java/com/wikia/services/userattribute/UserAttributeApplication.java
```diff
@@ -46,8 +49,10 @@ public String getName() {
 
   @Override
   public void run(UserAttributeConfiguration configuration, Environment environment)
-      throws Exception {
-    environment.jersey().register(JaxRsHalBuilderSupport.class);
+          throws Exception {
+      environment.jersey().register(JaxRsHalBuilderSupport.class);
+      configurePublicCors(environment);
+    environment.jersey().register(APIGatewayAuthFactory.create());
   }
```
garthwebb: The indentation on L53,54 looks off

garthwebb:
LGTM

PR #328, opened 2015-07-14
[DIS] DIS-90 Refactor shard selection and session management

jimmordino:
Refactor the code that determines which shard to use and whether or not a shard session already exists.  Previously the code used shard name to determine if a session was in progress, but this does not work when a read-only method (slave shard) is nested within a read-write method (master shard).  This change caches the site ID along with the current shard information so that it can compare requested sessions by site, which is common to both read-only and read-write requests.  
- [x] cache and compare site ID when validating nested session requests
- [x] use the dao factory (instead of the MyBatis session manager) as the authoritative source of session in progress
- [x] update unit tests to reflect code changes
- [x] update DB configurations to use shard-specific Consul keys for DB urls

@jogura @armonr


armonr:
:ship:

PR #329, opened 2015-07-14
DIS-207 Add Service level tests for new thread sort and user specific post data

jogura:
@jimmordino @armonr


service/discussion/src/test/java/com/wikia/discussionservice/services/ThreadServiceTest.java
```diff
@@ -320,4 +333,112 @@ public void createThread_emptyTitle() {
...
+    threadService.getForumThread(adminUser, siteId, threadId, expectedLimit, expectedPage,
+                                 expectedSortKey, expectedSortDirection);
+
+    verify(threadDao).getThread(adminUser.getUserId(), siteId, threadId,
+                                expectedLimit, expectedOffset,
+                                expectedSortKey.name(),
+                                expectedSortDirection.name());
+
+    expectedSortKey = SortKey.TRENDING;
+    threadService.getForumThread(adminUser, siteId, threadId, expectedLimit, expectedPage,
```
jimmordino: I think we should probably be resetting mocks after we verify the behavior of a production method call.  In this case the verify calls are different, but in future tests we might be verifying the same call again and if we don't reset it may appear to have been called, but the call may be from a previous method invocation.

jimmordino:
:+1:

armonr:
LGTM

wtf github? don't show already merged changes!

PR #330, opened 2015-07-14
DIS-206 Add Integration test for trending sort order

armonr:
Also add setter call in the map helper for extracting trendingScore

@jogura @jimmordino 

https://wikia-inc.atlassian.net/browse/DIS-206


jimmordino:
:+1:

PR #331, opened 2015-07-14
SERVICES-508 Fixed trailing slash bug for Central Swagger UI

connietruong:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-508
When using swagger, the link:
https://services.wikia.com/central-swagger-ui/?url=https://services.wikia.com/user-preference/swagger.json
works, but does not work if the slash after central-swagger-ui is removed:
https://services.wikia.com/central-swagger-ui?url=https://services.wikia.com/user-preference/swagger.json

Fixed the bug so that swagger works with or without the trailing slash.


nmonterroso:
LGTM

PR #336, opened 2015-07-15
DIS-181 Remove restriction on upvoting own post

armonr:
Also exception thrown on attempt to upvote anything other than posts (until supported)

@jogura @jimmordino 

https://wikia-inc.atlassian.net/browse/DIS-181


jimmordino:
:+1:

PR #337, opened 2015-07-15
DIS-180 IsAnonymous should also check userId is 0

armonr:
CC @jogura @jimmordino


service/discussion/src/main/java/com/wikia/discussionservice/auth/User.java
```diff
@@ -72,7 +70,7 @@ public boolean isBanned() {
   }
 
   public boolean isAnonymous() {
-    return checkSpecificRole(Role.ANON);
+    return checkSpecificRole(Role.ANON) && getUserId() == 0;
```
jimmordino: Is Role.ANON always set for all test variants?  Maybe ```&&``` should be ```||```?

jimmordino:
:+1:

PR #338, opened 2015-07-15
DIS-172 Add real trending_score column and update sproc

jogura:
* Update the thread's trending score only if the first_post is upvote or undone
* Add integration tests to prove sproc is working correctly
* Minor renaming of test names

@jimmordino @armonr


jimmordino:
:+1: , pending tests

PR #340, opened 2015-07-16
DIS-203 Add Cache headers to response

armonr:
* Adds `Vary` Header (with `Cookie` and `Accept-Encoding`)
* Adds `Cache-Control` set to `private`
* Also removes local CORS call (should be handled by the API gateway now)

@jimmordino @jogura 

https://wikia-inc.atlassian.net/browse/DIS-203


service/discussion/src/main/java/com/wikia/discussionservice/utils/CacheControlFilter.java
```diff
@@ -0,0 +1,54 @@
...
+      HttpServletResponse httpResponse = (HttpServletResponse) response;
+      Collection<String> headers = httpResponse.getHeaders("Vary");
+      Set<String> varyHeaderValues = new HashSet<>();
+      if (headers == null || headers.isEmpty() || !headers.contains("Cookie")) {
+        varyHeaderValues.add("Cookie");
+      }
+      if (headers == null || headers.isEmpty() || !headers.contains("Accept-Encoding")) {
+        varyHeaderValues.add("Accept-Encoding");
+      }
+      httpResponse.addHeader("Vary", String.join(", ", varyHeaderValues));
```
jimmordino: This won't do what we need.  The end result should be a single ```Vary``` header with all values that were present prior to the filter and ```Cookie``` and ```Accept-Encoding``` appended if they weren't present to start with.  So you need initialize ```varyHeaderValues``` with the values in the ```headers``` collection so that they don't get thrown away and after tests for the other 2 values, you need to call ```setHeader``` with the comma separated values.  Calling addHeader will add another Vary header, which is legal, but not what we want.

armonr: Fixed.
On my local, `Vary` is initially empty, but a filter after this sets another `Vary` header. I have not found anyone saying this violates the RFC specs, so this should likely work.

service/discussion/src/main/java/com/wikia/discussionservice/DiscussionServiceApplication.java
```diff
@@ -92,7 +96,9 @@ public void run(DiscussionServiceConfiguration configuration, Environment enviro
       throws Exception {
 
     // Allow cross-origin requests, at least until the API gateway handles this.
-    configurePublicCors(environment);
+    //configurePublicCors(environment);
```
jogura: Please delete this and the comment above.

service/discussion/src/main/java/com/wikia/discussionservice/utils/CacheControlFilter.java
```diff
@@ -0,0 +1,52 @@
...
+  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+      throws IOException, ServletException {
+
+    if (request instanceof HttpServletRequest) {
+      HttpServletResponse httpResponse = (HttpServletResponse) response;
+      Collection<String> headers = httpResponse.getHeaders("Vary");
+      Collection<String> varyHeaderValues = headers;
+      if (headers == null || headers.isEmpty() || !headers.contains("Cookie")) {
+        varyHeaderValues.add("Cookie");
+      }
```
jogura: Won't this result in an NPE if `headers` is null? Isn't `varyHeaderValues` the same reference as `headers` in this code (look at how `varyHeaderValues` is initialized)?

service/discussion/src/main/java/com/wikia/discussionservice/utils/CacheControlFilter.java
jimmordino: It's not a question of the RFC specs.  Varnish at one point only processed a single, comma separated Vary header.  I'm not sure if this is still the case, but we have the ability to create a single header, so that's the safest way to do this.  If there's another filter setting Vary headers we may need to change the order of the chain or disable the other filter.

service/discussion/src/main/java/com/wikia/discussionservice/utils/CacheControlFilter.java
jimmordino: This was caused by unclear instructions from me, sorry.  I asked for ```varyHeaderValues``` to be initialized with ```headers```, which is unclear and @armonr interpreted that as assignment - perfectly reasonable.  The javadoc states that ```getHeaders``` returns a Collection, possibly empty.  It's not clear whether or not that collection will always be mutable, so we should probably not assign it directly here, but rather use either ```Collections.copy``` or ```Collections.addAll```

jimmordino:
:+1:

PR #341, opened 2015-07-16
DIS-131 Add the constraint checks for creating a post/reply

jogura:
* Checks the status of the thread: isHeld, isViewable, isDeleted and if exists
* Create unit tests to cover new constraints

@jimmordino @armonr


service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
```diff
@@ -116,7 +117,48 @@
...
+    }
+
+    if (forumThread.getIsHeld()) {
+      LOGGER.info(String.format("An attempt to add post to a thread that does is held "
+                                    + "was performed by user id: %s site id: %s thread id: %s",
+                                user.getUserId(), siteId, threadId));
+      throw new DiscussionServiceException(
+          String.format("Unable to create the post on thread id: %s. "
+                            + "The thread is held.", threadId));
+    }
```
jimmordino: If you're going to check ```isHeld``` (a post-MVP feature), you also need to check ```isEditable```.  Threads can be locked/unlocked, which would change the value of this flag.

jogura: I thought `isEditable` was just on the thread's attributes (titles, etc). Does it extend to the posts?

jimmordino: it's everything about the thread - you can't modify the post list if the thread is not editable

jimmordino:
:+1: , pending tests

PR #343, opened 2015-07-17
[DIS] DIS-210 Refactor discussion test db cloning mechanism

jimmordino:
- [x] When dumping test_master use a separate config from the test config to allow for local customizations to the test config
- [x] update all yaml files to use the new consul keys (DB_URL should now be obsolete)
- [x] update ShowTestConfig command to use new property structure (shards, master/slave)
- [x] Update build script to properly pass the URL of the newly cloned integration test DB down to the integration test process
- [x] dump latest schema, sprocs, and data from test_master to integration_db.sql, and commit
- [x] clean up check bugs and test data in tests

@armonr @jogura


armonr:
:shipit:

PR #344, opened 2015-07-20
DIS-131 Few more authz policy adjustments on replies

armonr:
* 404 if not viewable or held
* 403 If not editable

@jimmordino @jogura


service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
```diff
@@ -172,6 +172,6 @@
     if (user.isStaff() || user.isAdmin() || user.isLoggedInUser()) {
       return true;
     }
-    throw new ForbiddenException("This user is not authorized to create post.");
+    throw new ForbiddenException("This user is forbidden to create post.");
```
jimmordino: I think the original language was correct.  The response may be ```403 Forbidden```, but the reason is a lack of authorization.  "This user is forbidden" implies an explicit ban or denial of privilege as opposed to a lack of privilege.

armonr: I will revert :point_up:

jimmordino:
LGTM

PR #345, opened 2015-07-20
DIS-215 Add editable constraint to thread creation

armonr:
Also thread create on forum disallowing threads should return a 403 Forbidden

@jimmordino @jogura 

https://wikia-inc.atlassian.net/browse/DIS-215


jimmordino:
:+1:

PR #348, opened 2015-07-21
add '//no tests' annotation to disable project with no tests in jenkins

sqreek:
optimize test runner in jenkins


harnash:
On remark that with this filter hardcoded into the gradle.settings we risk skipping test when they are added later. Otherwise LGTM

PR #365, opened 2015-07-29
standardize on dev/prod

nmonterroso:
@Wikia/services-team @jsutterfield @jogura @jimmordino @armonr 
https://wikia-inc.atlassian.net/browse/SERVICES-534

add configurations for all services to standardize on `dev` and `prod` for environments, the same way that non-pandora applications work at Wikia. Upon merge, this will require teams to update their `gradle.properties` and to switch from using `deployTesting` to `deployDev` and `deployProduction` to `deployProd`. Further, they'll have to update their consul configurations to match the new environment when they find the time to do so. We'll make the switch at the api-gateway level to query for the `prod` tag once each service is in production with both `prod` and `production` tags, and then finally remove the `production` tag completely. Same will be done in the (not yet complete) dev mesos/marathon cluster.

DO NOT merge until:
- [x] update the Pandora wiki with the new `gradle.properties` settings required to deploy, and the new deploy tasks
- [x] send email to engineering org stating the changes that need to take place now (updating `gradle.properties`, `deployDev/deployProd` tasks instead of `deployTesting/deployProduction`) and the steps they should take the next time they deploy (setting up consul configuration)
- [ ] plan for mediawiki to use the correct tags for user-preference and (possibly) user-attribute


README.md
```diff
@@ -1,4 +1,4 @@
-# Pandora
+# Pandora
```
drsnyder: Did you mean to add the white space at the end of the line?

nmonterroso: Yes and no. I didn't want to, but there were 30 Travis checks that I could either restart manually or "bump" the build by adding a non-impactful change.

I'll revert this

harnash:
@nmonterroso: I did rerun the tests for this branch and there are basically two types of errors:
1. timeouts due to a long running tests (can't think why there are tests running longer than 1m though)
2. tests are out of date and needs to be updated due to code changes
3. tests needs to be fixed because assumptions they rely on are out of date (WikiCitiesDAOIntegrationTest::getUserPreferencesForInvalidUserWithoutWhiteList() expects properties for user with id=0 be absent which is not the case in our dev database) - we should mock the DB calls and/or use fixtures

Either way I don't think there are any PR-specific errors and those problems should be fixed on master.

harnash:
UPDATE: Rerun on timed-out tests temporary fixes this problem. Only one which need some intervention left. This issue with unpredictable test behaviour needs to be fixed ASAP

nmonterroso:
Yeah, something somewhere is adding user info to the db with user id = 0, so it causes user pref tests to fail sometimes. I definitely agree we should do something about that practice, but that shouldn't be part of this pull request

nmonterroso:
tests seem to randomly be failing.. i run these locally and they work :(

drsnyder:
lgtm

harnash:
@nmonterroso:
Yeah they are timing out and I think I have a clue:
```
20:33:41 Download http://pkg-s2.wikia-prod:81/repository/internal/com/wikia/gradle/marathon/0.5.14/marathon-0.5.14.pom
20:33:43 Download http://pkg-s2.wikia-prod:81/repository/internal/com/wikia/gradle/marathon/0.5.14/marathon-0.5.14.jar
20:43:19 Build timed out (after 10 minutes). Marking the build as aborted.
20:43:19 Build was aborted
```
For some reason downloading marathon plugin package stalls.

PR #367, opened 2015-07-30
Solr connector fix

rybmat:


lib/solr/src/main/java/com/wikia/solr/SolrConnection.java
```diff
@@ -82,7 +83,7 @@ public static SolrConnectionError errorStatus(String methodKey, Response respons
       if (response.body() != null) {
         try {
           message += Util.toString(response.body().asReader());
-        } catch (IOException e) {
+        } catch (IOException ignored) {
```
pchojnacki: It would be nice to also log this exception

pchojnacki:
One suggestion otherwise LGTM

lib/solr/src/main/java/com/wikia/solr/BatchUpdate.java
```diff
@@ -119,6 +120,7 @@ private void sendUpdate() {
     Map<String, String> params = new HashMap<>();
     params.put("fl", "id,_version_,deleted_b");
     params.put("q", String.format("id:(%s)", idsStr));
+    params.put("rows", "9999999");
```
SebastianMarzjan: The response payload will be huge, do we need so many rows?

PR #372, opened 2015-08-03
SOC-678 add script to migrate user attribute data

jsutterfield:
Migrate data from the `wikicities.user_properties` table into the `userattributes.user_attribute ` table. Note, these are in different databases, but on the same database server.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-678


rafalkalinski:
One question comes to my mind: how hard would it be (or how easy), to update the attribute names for fit to one name convention (like camel case)?
And generally: LGTM

jsutterfield:
Hmm, good question @rafalkalinski. I don't think that wouldn't be too difficult, but I also think that it would add some complication. Since these user attributes are used throughout MW, if we did update their names in the db we'd either have to then update all their usage throughout the code, or transform those names when they enter the Attribute Service to conform to the new naming convention. In the end I think the confusion that mapping could cause outweighs the benefit of having all our attributes following a single naming convention.

That being said, I do think all new attribute we add should try and follow a single naming convention. That would be nice :)

mklucsarits:
LGTM

PR #379, opened 2015-08-06
case-insensitive sorting and add missing prefs

nmonterroso:
@Wikia/services-team 

re-syncing service whitelist with what's in MediaWiki


drsnyder:
lgtm

PR #380, opened 2015-08-07
GeoIP service

pchojnacki:
Simple service that returns GeoIP headers provided by our Edge as a JSon object.

currently deployed as /geoip

TODO:
* [x] change name to geoip
* [x] add tests
* [x] create Jira ticket


service/geoip/README.md
```diff
@@ -0,0 +1,25 @@
+# GeoIp
+Please enter a short, general description about what this service does
```
macbre: Please do so :)

service/geoip/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+        "-XX:+UseBiasedLocking", "-XX:+UseConcMarkSweepGC", "-XX:+UseParNewGC"
+]
+
+deployments {
+    prod {
+        resources {
+            instances = 2
+        }
+        environment {
+            CONSUL_TAGS = '["prod", "production", "swagger"]'
```
jcellary: BTW I do the same in my new service, but didn't we want to remove production tag and just leave prod?

service/geoip/src/main/java/com/wikia/services/geoip/resources/GeoIpResource.java
```diff
@@ -0,0 +1,45 @@
...
+  @Path("/location")
+  @Timed
+  @CacheControl(isPrivate = true, maxAge = 0, mustRevalidate = true)
+  @Produces
+  @ApiOperation(
+      value = "returns location from geoip headers",
+      notes = "returns location from geoip headers",
+      produces = MediaType.APPLICATION_JSON)
+  @ApiResponses(value = {
+      @ApiResponse(code = 200, message = "successfully sent geo ip", response = GeoIp.class)
```
jcellary: Maybe it's worth mentioning what is returned when required headers are not provided? Bad request, etc. (I'm assuming Bad Request is then returned and not 500 because of internal exception)

jcellary:
I believe the problem of consul tag is because the create-new-service gradle template has both production and prod specified. Maybe if you're at it, you could correct that there as well? (Assuming this is what we want)

jcellary:
Looks good, just some minors, feel free to merge after correcting

service/geoip/src/main/java/com/wikia/services/geoip/resources/GeoIpResource.java
pchojnacki: Headers are implementation specific detail. I.e. they are provided by Fastly when the request traverses our stack. The api client is not expected to provide them. Although I now think that a disclaimer that this service shouldn't be used in current implementation without going through internet.

pchojnacki:
@jcellary this service was actually created a long time (month or more) ago. Since then the new service creation is already fixed... I presume.

jcellary:
I checked it's not :D But no worries, I can also change that in my next PR

PR #391, opened 2015-08-14
SERVICES-586 Added options functionality to Swagger client download

connietruong:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-586

The initial Swagger client download created a zip file with the service API's with the selected language's default options. This updates the client download such that when an API and language are selected, options will appear. You can add options and the resulting zip file will include the options inputted.


service/central-swagger-ui/src/main/java/com/wikia/services/centralswaggerui/domain/SwaggerClientDownload.java
```diff
@@ -121,6 +124,23 @@ public StreamingOutput getZipFile(String language, SwaggerDoc currentDoc,
...
+  private CodegenConfig setConfigurations(String language, String tmpFolder,
+                                          HashMap<String, String> options) {
+    CodegenConfig config = new SwaggerLanguageConfigs().getLanguageConfig(language)
+        .orElseThrow(
+            () -> new IllegalArgumentException("invalid language"));
+    config.setOutputDir(new File(tmpFolder).getAbsolutePath());
+
+    for (CliOption langCliOption : config.cliOptions()) {
+      String optionName = langCliOption.getOpt();
+      if (options.containsKey(optionName) && !options.get(optionName).isEmpty()) {
```
nmonterroso: should we work in `.trim()`, to be safe?

service/central-swagger-ui/src/main/java/com/wikia/services/centralswaggerui/domain/SwaggerClientDownload.java
```diff
@@ -74,7 +76,8 @@ private SwaggerDoc getSwaggerDoc(String serviceName) {
             () -> new IllegalArgumentException("invalid service name"));
   }
 
-  public StreamingOutput getZipFile(String serviceName, String language)
+  public StreamingOutput getZipFile(String serviceName, String language,
+                                    HashMap<String, String> options)
```
nmonterroso: Let's take in a `Map<String, String>` where we're expecting `HashMap<String, String>` (there are a few places in this class)

nmonterroso:
LGTM

PR #393, opened 2015-08-14
remove deprecated mobile config from sonar

sqreek:
Technical debt was split into mobile-configurations and mobile-applications.
Deprecated code don't need to be showed in sonar


harnash:
LGTM

PR #397, opened 2015-08-18
Exclude the junit group from "compile" dependencies 

jogura:
This was pulling in an older version of hamcrest that was higher in the classpath which causes the IDE to report erroneous errors.

@jimmordino @armonr


jimmordino:
LGTM

armonr:
Nice! LGTM too

PR #401, opened 2015-08-19
[DIS] Fix stats proc so that it works on an empty forum

jimmordino:
Fix the case where there is no max post ID because there are no posts.  This edge case arose when we removed first post from the counts and became visible when we created and empty forum and tried to create the first thread (i.e. no posts because there were not yet any replies)


jogura:
LGTM

service/discussion/sql/create_discussion_db.sql
```diff
@@ -490,17 +490,18 @@ CREATE PROCEDURE `update_post_stats` (IN siteId INTEGER, IN postId BIGINT)
...
     WHERE thread_id = threadId AND post.site_id = siteId
           AND post.viewable_ind IS TRUE AND post.deleted_ind IS FALSE AND post.held_ind IS FALSE;
 
-    IF postCount = -1 OR maxPostId = 0 THEN
+    IF postCount = -1 OR maxPostId = -1 THEN
       SIGNAL SQLSTATE '45002' SET MESSAGE_TEXT = 'Unable to update thread stats';
     END IF;
 
-    UPDATE thread SET post_count = postCount, last_post_id = maxPostId
+    UPDATE thread SET post_count = postCount, last_post_id = NULLIF(maxPostId, 0)
```
armonr: So `if maxPostId == 0 -> NULL else -> maxPostId`. Is that the intention?

jimmordino: Correct.  The target column is nullable and we never want to have an invalid ID in there

PR #405, opened 2015-08-20
SERVICES-249 Edited unit test running script for pandora so it only runs tests for dependent projects of files changed

connietruong:
https://wikia-inc.atlassian.net/browse/SERVICES-249
@Wikia/services-team 

The change is in unit-test-selector.py in the pandora repo. The files takes in a string of the list of files altered, and creates numbered .param files each containing the name of the unit tests to run just like the original bash script. The selector looks through all of the projects with tests and looks if they're dependent on any of the projects that have been changed by the pull request. The selector then takes that lists and creates the .param files of only the files dependent on it.


nmonterroso:
LGTM

PR #417, opened 2015-08-28
Use a super account when dumping and cloning DBs

jimmordino:
MySQL requires SUPER permission to create stored procs, so we can't use the same account to dump/clone DBs that we use within the service to read and write data.  This change also involves adding two new keys to our consul config in DEV/TESTING, which has already been done.

@jogura @armonr


jimmordino:
Feel free to merge and then pull changes into your PR branches

armonr:
Thanks! :ship: it!

PR #418, opened 2015-08-28
DIS-279 Add statistics check to existing Post Delete Integration Tests

jogura:
Implement ForumHelper to get forum
Refactor tests to use ForumHelper
Refactor tests to use helper method to create expected response codes
Add check for parent forum counts, and for thread's post count
Add //NOSONAR to quiet down test code false positives

Merged in dis-master to get the changes to make the tests run properly.

@jimmordino @armonr


jimmordino:
:+1:  very clean.  deletes are going to change http operation and response so we should merge this asap

PR #425, opened 2015-08-31
Disable consul registration by default

drsnyder:
This disables consul registration by default. This should prevent developer instances of services from registering globally.

Service registration is then enabled (by default) in the dev and prod environments.

https://wikia-inc.atlassian.net/browse/SERVICES-691

/cc @Wikia/services-team @pchojnacki


lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/ConsulGuiceApplication.java
```diff
@@ -35,15 +35,8 @@
    *  find current port the app is exposing and will register itself in Consul so that other
    *  other services will be able to find it ip/port address.
    */
-  protected Boolean enableConsulAutoregistration;
+  protected Boolean enableConsulAutoregistration = Boolean.valueOf(System.getenv("ENABLE_CONSUL_REGISTRATION"));
```
pchojnacki: Hmm

pchojnacki: IIRC there is actually another level of Guice available here to provide predictable configuration for Consul itself.
e.g. in ConsulManager you have the @Configuration annotation. It actually is configured through Guice/Governator Netflix configuration library.
```
public class ConfigurationModule implements BootstrapModule {
  @Override
  public void configure(BootstrapBinder binder) {
    binder.bind(PropertiesConfigurationProvider.class)
        .toProvider(DefaultsFromPropertiesConfigurationProvider.class);
    binder.bindConfigurationProvider().toProvider(ConfigurationBootstrapProvider.class);
  }
}
```
Its configured through both Environment variables and properties file: consul_defaults.properties.
See
https://github.com/Netflix/governator/wiki/Configuration-Mapping

I think you could reuse/extend this functionality to provide the endresult you look for.

drsnyder: It's not clear to me how you would do this here. The modules haven't been loaded at this point in the code. We would probably have to move the `enableConsulAutoregistration` into `ConsulManager` but I think that would break some of the assumptions in the child classes.

drsnyder: I tried to push this down the stack but I wasn't able to do so because some of the functionality is in `ConsulManager`, `Consular` and `ConsulRegistrationInstaller`. In theory we could add `@Configuration` injection to all three but that would result in a lot of `if/else` blocks.

drsnyder: This might work if all of the functionality was moved from the installer to the `ConsulManager`. That is, if some form of `ConsulManager.setup` is called from the installer that does the heavy lifting:

```
      environment.lifecycle()
          .addServerLifecycleListener(getHostingPortRegistrationListener());

      scheduleHeartbeats(environment.lifecycle()
                             .scheduledExecutorService("consul-heartbeat-scheduler")
                             .threads(1).build());

      environment.healthChecks().register("check-heartbeats", getHeartbeatHealthCheck());
      environment.lifecycle().manage(getManagedBean());
```

'cause I can't seem to get `Configuration` to work in the installer but it does work in the `ConsulManager`.

Also, I'm unable to override with environment variables. It looks like the way this works is you only get the configuration provided from `DefaultsFromPropertiesConfigurationProvider`. I didn't see anything about environment overrides in the [documentation](https://github.com/Netflix/governator/wiki/Configuration-Mapping).

Also, why did we introduce another configuration scheme in `lib/dropwizard/consul/src/main/resources/consul_defaults.properties`? It doesn't seem right to have two configuration files in `service.yaml` and `consul_defaults.properties` (and consul).

drsnyder:
Closing in favor of #428 and focusing solely on he flapping.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/ConsulGuiceApplication.java
pchojnacki: I need some form of configuration to setup consul before even DropWizard starts so that DW plugin could fetch data from Consul. Also data in consul_defaults.properties is as you seen global so you change this here and it will be propagated to all services.

This configuration is pretty generic and can be used even outside of DropWizard

tbh I'd like to depracate (in some unspecified future) service.yaml for most of configuration options because of this very reason and because it causes many problems with how guice lifecycle is integrated. Governator based configuration also supports dynamic reconfiguration so that would enable 0-restart configuration changes.

jogura:
As a short-term bandaid, LGTM

drsnyder:
Ok, I tested it in dev with user-preference (was flapping) and community-data and it registered / deregistered as expected. I also tested on my laptop (with no changes in environment) and it did not register.

@Wikia/services-team any objection to merging?

nmonterroso:
LGTM shipit

PR #426, opened 2015-08-31
DIS-281 attempt to minimize thread updates when calculating vote stats

jimmordino:
Don't attempt to update thread unless the upvote is on a first post.  If we do attempt to update, lock first.

@jogura @armonr


service/discussion/sql/create_discussion_db.sql
```diff
@@ -563,9 +565,27 @@ CREATE PROCEDURE `update_content_votes` (IN contentType SMALLINT UNSIGNED, IN co
...
+
+      IF threadToUpdate != -1 THEN
+        SELECT trending_score
+        INTO trendingScore
+        FROM thread
+        WHERE id = threadToUpdate
+        FOR UPDATE;
+
+        IF trendingScore = -1 THEN
+          SIGNAL SQLSTATE '45014' SET MESSAGE_TEXT = 'Thread not found';
```
armonr: At this point, existence of thread is already verified (line 573). Maybe "invalid trending score"?

jimmordino: This doesn't validate the thread, it validates the query and the lock.   The column isn't important.

jogura:
LGTM

PR #428, opened 2015-09-01
Increase consul heartbeat ttl to reduce flapping

drsnyder:
Increasing `consul.heartbeat.ttl` to 30s. The previous 2s timeout appears to be too short and causes flapping.

I opened this in favor of #425 to decouple the registration toggle. Decoupling the registration will be more involved.

https://wikia-inc.atlassian.net/browse/SERVICES-651

/cc @pchojnacki @Wikia/services-team


pchojnacki:
30s seems pretty arbitrary it also reduces the usefulness of heartbeats every 1s.
I'd suggest following setup: e.g. allow application to miss 3 heartbeats but also extend the 1s interval to e.g. 2 or 5s.

and set TTL to be 3x of heartbeat 6s or 15.

drsnyder:
> allow application to miss 3 heartbeats 

Is that configurable?

pchojnacki:
Not directly. If you set TTL to 15s and send heartbeats every 5s  then if application will miss 3 heartbeats in a row it will be marked as failed.

drsnyder:
Ok, sounds reasonable. I'm going to test 5 & 15:

```
+consul.heartbeat.ttl=15000
+consul.heartbeat.interval=5000
```

If that works ok with no flapping I'll proceed with that config.

drsnyder:
Hmm. I'm seeing some flapping. I'll try lowering the interval.

drsnyder:
Ok, 15+3 seems to be stable.

ArturKlajnerok:
Lgtm :+1:

pchojnacki:
LGTM :+1:

PR #429, opened 2015-09-01
DIS-245 update positions that are stale OR null

jimmordino:
Update post positions from stale values (e.g. after delete) or from null values (e.g. initial migration).

@jogura @armonr


jogura:
LGTM

armonr:
:+1:

PR #436, opened 2015-09-02
use new version of marathon plugin

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-679

use the latest version of the marathon plugin that has better error messages for conflicting deployments (see Wikia/marathon_gradle_plugin#12)


ArturKlajnerok:
Lgtm :+1:

PR #438, opened 2015-09-03
SOC-1215 DIS Service - Remove environment-specific guice modules

armonr:
https://wikia-inc.atlassian.net/browse/SOC-1215

@jimmordino @jogura


WikiaTech:
SonarQube analysis reported 1 issue:
* MAJOR 1 major

Watch the comments in this conversation to review them.
Note: the following issues could not be reported as comments because they are located on lines that are not displayed in this pull request:
* MAJOR [Make the "LOGGER" logger private static final.](https://github.com/Wikia/pandora/blob/117d24dfe8cf524b48a50a1e1439855c7567e864/service/discussion/src/main/java/com/wikia/discussionservice/DiscussionServiceApplication.java#L46) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1312)

jimmordino:
LGTM, as long as it behaves gracefully when ```ENVIRONMENT``` is not set.

armonr:
@drsnyder @nmonterroso: Should a service set a default `ENVIRONMENT` variable (as in https://github.com/Wikia/pandora/pull/438/files#diff-9ac1a781236977441113646e742dc98eL64)?

pchojnacki:
Hi,guys. You should set environment according to the environment you are running in (defaults are: dev and prod, but you can add more if you wish).  Consul uses this information to build the key to fetch appropriate data.

Additionally it would be nice if the code that you just removed was also removed from example service because its actually only really used in mobile-config service and for some reason it was copied into example.

service/discussion/src/main/java/com/wikia/discussionservice/DiscussionServiceApplication.java
```diff
@@ -64,12 +64,12 @@ public String getName() {
 
   @Override
   public List<Module> getAdditionalModules() {
-    return Arrays.asList(new DiscussionMyBatisModule());
+    return Collections.singletonList(new DiscussionMyBatisModule());
```
pchojnacki: FYI you can also set this by ```instal(new DiscussionMyBatisModule())``` in DiscussionModule configure method.
This is usally where I'm adding additional modules.

But if you like it here than please ignore my comment.

pchojnacki:
Also if you'll have some spare time please look into the LOGGER object being public if this instance is used anywhere it should be replaced with new static instance so that each class that needs Logger has its own instance.

BTW. https://projectlombok.org/features/Log.html

drsnyder:
> Consul uses this information to build the key to fetch appropriate data.

@pchojnacki where does that happen?

pchojnacki:
@drsnyder 
https://github.com/Wikia/pandora/blob/master/lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/config/ConsulKeyValueConfigModule.java

although I see there are some changes in consul kv itself. But it will be incompatible once we merge clusters and/or have more different config options. In that case I don't understand the changes nor the code that is leftover.

armonr:
Ticket created for the logger [DIS-289](https://wikia-inc.atlassian.net/browse/DIS-289)

jogura:
Thanks for opening the logger ticket. LGTM

armonr:
@pchojnacki Thanks for code review. I created a ticket to fix the logger, and one ticket for updating the example service [SERVICES-697](https://wikia-inc.atlassian.net/browse/SERVICES-697).

WikiaTech:
SonarQube analysis reported 1 issue:
* MAJOR 1 major

Watch the comments in this conversation to review them.
Note: the following issues could not be reported as comments because they are located on lines that are not displayed in this pull request:
* MAJOR [Make the "LOGGER" logger private static final.](https://github.com/Wikia/pandora/blob/117d24dfe8cf524b48a50a1e1439855c7567e864/service/discussion/src/main/java/com/wikia/discussionservice/DiscussionServiceApplication.java#L46) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1312)

service/discussion/build.gradle
```diff
@@ -59,12 +59,6 @@ deployments {
     }
 }
 
-run {
-    environment << [
-        ENVIRONMENT: System.getenv("ENVIRONMENT") ?: "production",
-    ]
-}
-
```
drsnyder: I wonder if this is/was causing some of the discussion service instances to start up & register in production. That _could_ explain why we have so many orphans in production now.

armonr: It is a possibility, and that is why we tackled this now to either stop creating orphaned instances or rule this out as the cause.

pchojnacki: @drsnyder good intuition but afaik if you use agent api to register instance (which we do use) then the service will always be registered in the same DC as the agent is configured. Its also how consul knows the ip address of service it just takes IP configured in the agent the service is registering in. So if I were to guess the cause if orphans is that either someone/something has CONSUL_HOSTNAME configured to e.g. consul-s1, consul.service.sjc.consul. Or that service itself is constantly killed in a manner that makes the unregister code not fire (e.g. oomkill).

pchojnacki:
@armonr no problem, I just noticed PR with some code I'm actually familiar with.

service/discussion/build.gradle
drsnyder: > So if I were to guess the cause if orphans is that either someone/something has CONSUL_HOSTNAME configured to e.g. consul-s1, consul.service.sjc.consul. 

I think that might be happening.

> Or that service itself is constantly killed in a manner that makes the unregister code not fire (e.g. oomkill).

But if that was the case, wouldn't the orphaned service be associated with the node it registered on (e.g. `mesos-s1`)?

pchojnacki: Yes it should be associated. Although theoretically speaking its also possible to deploy to Mesos and override consul hostname if someone wishes so.

PR #439, opened 2015-09-03
MOB-5455 Add index to improve query performance

jogura:
* To avoid full table scans of threads and votes add an index and an extra parameter to utilize existing index

@jimmordino @armonr


jimmordino:
:shipit:

armonr:
LGTM

PR #445, opened 2015-09-08
[PLATFORM-1442] Added reconnection to rabbit and thread sync up

jcellary:
Added checking if connection to rabbit is closed and if so reconnecting. Also added a wrapper which synchronizes operations on rabbit connection which per default aren't thread safe.

Pinging @pchojnacki and @rybmat


lib/rabbit/src/main/java/com/wikia/rabbit/RabbitConnectionSyncWrapper.java
```diff
@@ -0,0 +1,33 @@
...
+
+  private RabbitConnection rabbitConnection;
+
+  public RabbitConnectionSyncWrapper(RabbitConnection rabbitConnection) {
+    this.rabbitConnection = rabbitConnection;
+  }
+
+  public QueueingConsumer getQueuingConsumer(String bindingKey, String queueName)
+      throws IOException {
+    synchronized (this) {
```
pchojnacki: I'm not sure synchronizing only the consumer access is the right way to go.
What would happen if two threads access this block. And after the block is executed? will queuing consumer be able to be shared among two threads

lib/rabbit/src/main/java/com/wikia/rabbit/RabbitConnection.java
```diff
@@ -110,7 +110,7 @@ public boolean publish(String routingKey, String messageText) {
   }
 
   private Channel getPublishChannel() throws IOException {
-    if (publishChannel == null) {
+    if (publishChannel == null || !publishChannel.isOpen()) {
```
pchojnacki: :+1:

pchojnacki:
I'm  not sure how workers rabbitmq works but maybe its possible to use some kind of connection pooling?

lib/rabbit/src/main/java/com/wikia/rabbit/RabbitConnectionSyncWrapper.java
jcellary: You're right I provided the wrapper with publisher in mind - to sync the publish method, the queuing consumer is again not thread safe, I provided this getter  just for consistency to have all public method of rabbit connection thread safe. So if someone wanted to use the queueing consumer from multiple threads he would need to implement a similar wrapper or use separate instances, but we don't need it for now as consumers are single threaded, so I can remove this method to avoid confusion and just leave publish or leave it as is with and added comment that when someone wants to build multi threaded clients this needs to be considered/extended

jcellary:
I haven't found a built in worker pool. I could implement one, but I'm not sure if its worth the effort for our case and expected traffic. I could also create a separate RabbitConnection instance per service class instance, but I think that would not be very efficient.

jcellary:
Per offline conversation, removed syncing as it's not needed for now and merging the rest

PR #453, opened 2015-09-14
add DependentOn groovy task for filter tests to run

sqreek:
modify dependenciesInsight task and create dependentOn task that filter projects to run only dependent tests.
//cc @Wikia/services-team


buildSrc/src/main/groovy/DependentOnTask.groovy
```diff
@@ -0,0 +1,149 @@
...
+    protected VersionSelectorScheme getVersionSelectorScheme() {
+        throw new UnsupportedOperationException()
+    }
+
+    @Inject
+    protected VersionComparator getVersionComparator() {
+        throw new UnsupportedOperationException();
+    }
+
+//    /**
```
pchojnacki: Uncomment or remove

pchojnacki:
1 minor and failing tests Otherwise LGTM

PR #457, opened 2015-09-15
MOB-5589 Change the mobile config service so that it accepts new json format

jogura:
The new json format contains the `languages` attribute that was causing an exception to be thrown when the json is parsed. This changes defines the `languages` attribute in the `MobileConfiguration` to avoid the exception.

This PR also enables the mobile-configurations service to be swagger enabled.
Fixes checkstyles and findbug errors.

RFCR @sqreek @jimmordino


service/mobile-configurations/src/main/java/com/wikia/mobileconfig/MobileConfigConfiguration.java
```diff
@@ -1,14 +1,20 @@
...
+import lombok.Getter;
 import org.hibernate.validator.constraints.NotEmpty;
 
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 
-public class MobileConfigConfiguration extends Configuration {
+@Getter
+public class MobileConfigConfiguration extends Configuration implements
+                                                             SwaggerConfigurationProvider {
```
jimmordino: Trivial: this indentation seems strange.  Line break before ```implements``` instead?

jimmordino:
:+1:  Looks reasonable to me

sqreek:
Mobile-config is (deprecated)[https://github.com/Wikia/pandora/tree/master/service/mobile-config#deprecated-mobile-config]

Please move changes to mobile-configurations.
Also if you use somewhere old mobile-config, please let us know, we want to remove mobile-config, but there is still some traffic for this service

jogura:
@sqreek because there is still traffic to `Mobile-Config` I had to make the change to both services ( `Mobile-Config` and  `Mobile-Configurations`)

sqreek:
LGTM

PR #461, opened 2015-09-17
WIP: MOB-5603 Add additional operations to allow for decommissioning of old mobile-config service

jogura:
~~DO NOT MERGE:~~
- [x] ~~Need to confirm that the applications will follow the redirect correctly~~ - __Confirmed works__

* Add operation to respond to old operation path
* Add operation to redirect for application list
* Requires API-GATEWAY configuration changes to point traffic to the mobile-config service's to the mobile-configurations service

@drsnyder @jimmordino


jimmordino:
LGTM

PR #465, opened 2015-09-18
fixing healthcheck

nmonterroso:
@Wikia/services-team 
forgot to change this when we switched to globals/locals. oops


ArturKlajnerok:
Lgtm :+1:

PR #466, opened 2015-09-21
SOC-1282 increase timeouts

jsutterfield:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-1282


service/user-attribute/src/main/resources/user-attribute.yaml
```diff
@@ -36,8 +36,8 @@ swaggerConfiguration:
     authKey: ${consul:SWAGGER_AUTH_HEADER:X-Wikia-UserId}
     authKeyLocation: ${consul:SWAGGER_SECURITY_TOKEN_TYPE:header}
 httpClientConfiguration:
-  timeout: 500ms
-  connectionTimeout: 500ms
+  timeout: 2s
+  connectionTimeout: 2s
```
jsutterfield: I'll be honest, I'm not sure what kind of science to apply here. 2 seconds seems like a reasonable amount of time. We had it set at 500ms before and only 1/10th of 1% of requests had exceptions, so very few requests will be affected by this timeout change.

If anyone has any strong opinions about why this should be higher or lower, or knows best practices when it comes to defining this kind of timeout, I'm happy to hear it.

armonr: I think @macbre's suggestion (in the ticket) is the preferable solution.
As a stopgap measure, though, this LGTM; also `retries` might help as long as the timeouts are not happening during traffic surges when, for instance, connection limit is reached.

jsutterfield: Agreed, using the purger queue would be prefereable, but right now there's no way to talk to it. See [this comment](https://wikia-inc.atlassian.net/browse/SOC-1282?focusedCommentId=196882&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-196882)

garthwebb:
LGTM

PR #470, opened 2015-09-21
add migration info with sample queries

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-657

Some information on the migration:
* federated table schema
* sample literal preference
* sample regex preference


drsnyder:
Thanks for committing!

PR #471, opened 2015-09-21
local preference whitelist

nmonterroso:
@Wikia/services-team 

Since on the new resource, local preferences do not come in with the wiki id attached, we need to account for them as literals, while also leaving in the regexes for the legacy resource.


drsnyder:
LGTM:SHIPIT:

PR #475, opened 2015-09-22
adding custom field for environment

nmonterroso:
@Wikia/services-team 

we don't have a way to separate environment logs from production logs in Kibana. This change adds the `env` field to every log. I've added the production config into Consul already.


drsnyder:
lgtm

PR #476, opened 2015-09-22
updating whitelist

nmonterroso:
@Wikia/services-team 

seeing some invalid preference names that should pass, like `searchNs-1` and `searchNsNS_MAIN` (the second is probably a bug on the mediawiki side, but imo we shouldn't try to fix/migrate that yet). Since we potentially want to get rid of the whitelist at some point I don't think there's an issue making it more inclusive.


service/user-preference/src/main/resources/user-preference.yaml
```diff
@@ -215,5 +215,5 @@ whitelist:
       - ([a-z]+)-toolbar-contents$
       - ^forum_sort_.*
       - ^gadget-.*
-      - ^searchNs([0-9]+)$
+      - ^searchNs.*$
```
drsnyder: Minor: should we be a more conservative with `^searchNs[a-zA-Z0-9_.-]*$`?

drsnyder:
Minor comment, OTW, :shipit:

service/user-preference/src/main/resources/user-preference.yaml
nmonterroso: hmm, i thought about doing it that way, but since we want to do away with the whitelist at some point i just thought "let's make it accept any suffix to avoid any potential errors"

drsnyder: Ok, sounds good.

PR #477, opened 2015-09-22
[DIS] Be explicit about character collation in creation and import scripts

jimmordino:
MySQL will default to utf8mb4_general_ci collation, which does not use strict Unicode rules for sorting and comparison.  Be specific about using utf8mb4_unicode_ci collation in the create and import scripts so that we adhere to unicode standards and don't cause errors when trying to compare strings of different utf8mb4 collations.

@armonr @jogura @jsutterfield


armonr:
:shipit:

PR #480, opened 2015-09-24
Made square validation optional

jcellary:
Made square validation optional, due to some issues in mobile apps. In future we will turn it on via consul.

Pinging @sqreek


service/user-avatar/src/main/java/com/wikia/services/useravatar/UserAvatarConfiguration.java
```diff
@@ -45,6 +45,9 @@
   @JsonProperty
   private int maxImageSizeBytes;
 
+  @JsonProperty
+  private boolean validateIsSquare;
```
sqreek: please rewrite it to use `@Data` from lombok

service/user-avatar/src/main/java/com/wikia/services/useravatar/resources/UserAvatarResource.java
```diff
@@ -77,18 +77,22 @@
 
   private final int maxImageSizeBytes;
 
+  private final boolean validateIsSquare;
+
   @Inject
   public UserAvatarResource(
       UserAttributeService userAttributeService,
```
sqreek: This constructor is started to have too much parameters.
Consider to use Inject on field's, and remove it from constructor.

service/user-avatar/src/main/java/com/wikia/services/useravatar/resources/UserAvatarResource.java
```diff
@@ -77,18 +77,22 @@
...
       ImageService imageService,
       ImageUrlBuilder imageUrlBuilder,
       AvatarValidator avatarValidator,
-      @Named("max-image-size-bytes") int maxImageSizeBytes) {
+      @Named("max-image-size-bytes") int maxImageSizeBytes,
+      @Named("validate-is-square") boolean validateIsSquare) {
```
sqreek: validateIsSquare? what this is mean?

service/user-avatar/src/main/java/com/wikia/services/useravatar/resources/UserAvatarResource.java
```diff
@@ -189,6 +193,10 @@ public Response getAvatarForUser(@PathParam("userId") int userId,
   private void validateAvatar(int userId,
                               InputStream fileInputStream,
                               FormDataContentDisposition fileDetail) {
+    if (!validateIsSquare) {
```
sqreek: above comment about naming, is mostly for understand what happen here.

service/user-avatar/src/test/resources/user-avatar-test.yaml
```diff
@@ -57,3 +57,4 @@ fixedStaticAssetsServiceUri: http://localhost:8811
 fixedUserAttributeServiceUri: http://localhost:8812
 thumbnailerHostname: static.wikia.com
 maxImageSizeBytes: ${consul:MAX_IMAGE_SIZE_BYTES:10485760} #defaults to 10 MB
+validateIsSquare: true
```
sqreek: is this should be in  consul?

jcellary: This config is only used for unit tests, so I don't think this should be in consul. The user-avatar.yaml file uses consul however.

sqreek: :+1:

service/user-avatar/src/main/java/com/wikia/services/useravatar/resources/UserAvatarResource.java
jcellary: I changed the name

service/user-avatar/src/main/java/com/wikia/services/useravatar/resources/UserAvatarResource.java
jcellary: Per offline conversation that would cause trouble due to the way out unit tests are built - it's much easier to handle direct parameters than injected fields in a jersey unit tests

service/user-avatar/src/main/java/com/wikia/services/useravatar/UserAvatarConfiguration.java
jcellary: I deleted all unneeded getters

PR #485, opened 2015-09-29
SOC-1368 | User Atrribute change marathod upgradeStategy to not release all instances at once

kvas-damian:
In order to improve service performance during the deploy (we're deploying new instance with empty cache) we decided to create new instances of service one by one (or actually by 10% in every step) instead all at once.

https://mesosphere.github.io/marathon/docs/rest-api.html#upgradestrategy


kvas-damian:
ping @pchojnacki

pchojnacki:
:+1: Looks good

Other options you could consider is maximumOverCapacity = 0
As this is supposed to cause 1 for one deployment. I.e if you have 10, or 100 services then deployment will always happen only 1 instance at a time

kvas-damian:
For now it's one by one ;) 
I wanted to avoid releasing one by one when number of instances grows to really big numer. ```10%``` just scale better.

pchojnacki:
:+1: cool LGTM

jsutterfield:
LGTM!

PR #486, opened 2015-09-30
Fix geoip deployment on production

pchojnacki:


sqreek:
No description :)

but LGTM, :sheep:

PR #492, opened 2015-10-05
SOC-1053 support for requesting all attributes for multiple users

bekcunning:
https://wikia-inc.atlassian.net/browse/SOC-1053

using /user/bulk?id=1&id=2&id=3 will return all attributes for those user id's, identified by userId in the response.  Invalid userID's will be ignored and only responses for valid userId's will be returned.


service/user-attribute/src/main/java/com/wikia/services/userattribute/core/UserWithAllAttributes.java
```diff
@@ -0,0 +1,22 @@
...
+  public int userId;
+
+  public List<UserAttribute> userAttributeList;
+
+  public UserWithAllAttributes(int userId, List<UserAttribute> userAttributeList) {
+    this.userId = userId;
+    this.userAttributeList = userAttributeList;
+  }
+
+  public int getUserId() { return userId; }
```
WikiaTech: MAJOR Move this closing curly brace to the next line. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3ARightCurlyBraceStartLineCheck)

service/user-attribute/src/main/java/com/wikia/services/userattribute/core/UserWithAllAttributes.java
```diff
@@ -0,0 +1,22 @@
...
+  public List<UserAttribute> getUserAttributeList() { return userAttributeList; }
```
WikiaTech: MAJOR Move this closing curly brace to the next line. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3ARightCurlyBraceStartLineCheck)

service/user-attribute/src/main/java/com/wikia/services/userattribute/core/UserWithAllAttributes.java
```diff
@@ -0,0 +1,22 @@
+package com.wikia.services.userattribute.core;
+
+import java.util.List;
+
+/**
+ * Created by rcunningham on 9/23/15.
+ */
+public class UserWithAllAttributes {
+
+  public int userId;
```
WikiaTech: MAJOR Make userId a static final constant or non-public and provide accessors if needed. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AClassVariableVisibilityCheck)

service/user-attribute/src/main/java/com/wikia/services/userattribute/core/UserWithAllAttributes.java
```diff
@@ -0,0 +1,22 @@
...
+
+  public List<UserAttribute> userAttributeList;
```
WikiaTech: MAJOR Make userAttributeList a static final constant or non-public and provide accessors if needed. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AClassVariableVisibilityCheck)

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,56 @@ public Response getAllAttributes(
...
+  public Response getAllAttributesForMultipleUsers(
+      @QueryParam("ids") String arr
+  ) throws java.sql.SQLException {
+
+    String[] items = arr.split(",");
+    Set<Integer> userIds = new HashSet<>();
+    for (int i = 0; i < items.length; i++) {
+      try {
+        userIds.add(Integer.parseInt(items[i]));
+      } catch (NumberFormatException nfe) {
```
WikiaTech: MAJOR Either remove or fill this block of code. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS00108)

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,56 @@ public Response getAllAttributes(
...
+      }
+      ;
```
WikiaTech: MAJOR Remove this empty statement. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AEmptyStatementUsageCheck)

WikiaTech:
SonarQube analysis reported 6 issues:
* MAJOR 6 major

Watch the comments in this conversation to review them.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,108 @@ public Response getAllAttributes(
...
+      @ApiParam(value = "The name of the attribute to be retrieved")
+      @PathParam("attrName") String attrName
+  ) throws java.sql.SQLException {
+
+    String[] items = arr.split(",");
+    Set<Integer> userIds = new HashSet<>();
+    for (int i = 0; i < items.length; i++) {
+      try {
+        userIds.add(Integer.parseInt(items[i]));
+      } catch (NumberFormatException nfe) {
```
WikiaTech: MAJOR Either remove or fill this block of code. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS00108)

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,108 @@ public Response getAllAttributes(
...
+      }
+      ;
```
WikiaTech: MAJOR Remove this empty statement. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AEmptyStatementUsageCheck)

WikiaTech:
SonarQube analysis reported 4 issues:
* MAJOR 3 major
* MINOR 1 minor

Watch the comments in this conversation to review them.
Note: the following issues could not be reported as comments because they are located on lines that are not displayed in this pull request:
* MINOR [Define a constant instead of duplicating this literal "properties" 3 times.](https://github.com/Wikia/pandora/blob/cfffd95897bac65269273f51ab257f3d116f66fd/service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java#L132) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1192)
* MAJOR [2 duplicated blocks of code must be removed.](https://github.com/Wikia/pandora/blob/cfffd95897bac65269273f51ab257f3d116f66fd/service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=common-java%3ADuplicatedBlocks)

service/user-attribute/src/main/java/com/wikia/services/userattribute/core/UserWithAllAttributes.java
WikiaTech: MAJOR Move this closing curly brace to the next line. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3ARightCurlyBraceStartLineCheck)

service/user-attribute/src/main/java/com/wikia/services/userattribute/core/UserWithAllAttributes.java
WikiaTech: MAJOR Move this closing curly brace to the next line. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3ARightCurlyBraceStartLineCheck)

service/user-attribute/src/main/java/com/wikia/services/userattribute/core/UserWithAllAttributes.java
WikiaTech: MAJOR Make userId a static final constant or non-public and provide accessors if needed. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AClassVariableVisibilityCheck)

service/user-attribute/src/main/java/com/wikia/services/userattribute/core/UserWithAllAttributes.java
WikiaTech: MAJOR Make userAttributeList a static final constant or non-public and provide accessors if needed. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AClassVariableVisibilityCheck)

WikiaTech:
SonarQube analysis reported 4 issues:
* MAJOR 4 major

Watch the comments in this conversation to review them.

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java
```diff
@@ -71,6 +74,47 @@ public UserAttribute map(Record record) {
...
+
+    List<UserWithAllAttributes> userWithAllAttributesList = new ArrayList<>();
+
+    Result<Record> userAttributes = database.dslForReads(USER_ATTRIBUTES_DB_NAME)
+        .select()
+        .from(USER_ATTRIBUTE)
+        .where(USER_ATTRIBUTE.USER_ID.in(userIds))
+        .fetch();
+
+    HashMap<Integer, UserWithAllAttributes> map = new HashMap<>();
```
kvas-damian: Please use more descriptive variable names

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java
```diff
@@ -71,6 +74,47 @@ public UserAttribute map(Record record) {
     return Optional.of(attributeList);
   }
 
+
+  public Optional<List<UserWithAllAttributes>> getAllAttributesMultipleUsers(Set<Integer> userIds)
+      throws java.sql.SQLException {
+
+    List<UserWithAllAttributes> userWithAllAttributesList = new ArrayList<>();
```
kvas-damian: You don't have to initialize this variable with value when you always override it.

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java
```diff
@@ -71,6 +74,47 @@ public UserAttribute map(Record record) {
...
+      List<UserAttribute> userAttributeList = new ArrayList<>();
+      Optional<UserAttribute> username = getUsername(userId);
+      userAttributeList.add(username.get());
+      userWithAllAttributes = new UserWithAllAttributes(userId, userAttributeList);
+      map.put(userId, userWithAllAttributes);
+    }
+    for (Record r : userAttributes) {
+      int userId = r.getValue(USER_ATTRIBUTE.USER_ID);
+
+      UserAttribute userAttribute = new UserAttribute(r.getValue(USER_ATTRIBUTE.ATTRIBUTE_NAME),
```
kvas-damian: Login for creation of ```UserAttribute``` looks pretty the same as the one in ```getAllAttributes```. Maybe we could extract it and use it in both places?

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java
```diff
@@ -71,6 +74,47 @@ public UserAttribute map(Record record) {
...
+                                                      byteArrayToUnicode(
+                                                          r.getValue(USER_ATTRIBUTE.VALUE)));
+
+      if (map.containsKey(userId)) {
+        userWithAllAttributes = map.get(userId);
+        List<UserAttribute> userAttributeList = userWithAllAttributes.getUserAttributeList();
+        userAttributeList.add(userAttribute);
```
kvas-damian: Maybe we could create a setter in ```UserWithAllAttributes```?

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java
```diff
@@ -71,6 +74,47 @@ public UserAttribute map(Record record) {
...
+      } else {
+        return Optional.empty();
```
kvas-damian: I think we should have better exception handling here. On the other hand I'm not sure if code can reach this condition.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,56 @@ public Response getAllAttributes(
   @GET
   @Timed
   @ApiOperation(
+      value = "Returns all available attributes for the specified list of users",
+      response = AllUserAttributesHalResponse.class
+  )
+  @ApiResponses(value = {
+      @ApiResponse(code = 404, message = "No attributes found for this userId")
```
kvas-damian: Please update message to reflect login in this method

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
kvas-damian: Is it required here?

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,56 @@ public Response getAllAttributes(
...
+      @QueryParam("ids") String arr
+  ) throws java.sql.SQLException {
+
+    String[] items = arr.split(",");
+    Set<Integer> userIds = new HashSet<>();
+    for (int i = 0; i < items.length; i++) {
+      try {
+        userIds.add(Integer.parseInt(items[i]));
+      } catch (NumberFormatException nfe) {
+      }
```
kvas-damian: Please add some exception handling for this case. IMO service should return ```400 bad request``` when input is not valid.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,56 @@ public Response getAllAttributes(
...
+    }
+    for (Integer userId : userIds) {
+      checkIfUserExists(userId);
+    }
+
+    Optional<List<UserWithAllAttributes>>
+        userWithAllAttributesList =
+        userAttributeDAO.getAllAttributesMultipleUsers(userIds);
+    if (!userWithAllAttributesList.isPresent()) {
+      return StandardResponses.NotFound(String.format("No attributes found for one or more users"));
```
kvas-damian: I'm not sure if we should return 404. Users exists but they don't have attributes connected to them, so we could return List of empty ```userWithAllAttributes```? This requires change in ```getAllAttributes```, @jsutterfield what do you think?
BTW message should say something like ```No attributes found for selected users```

bekcunning: Actually I don't think that this condition will ever be hit because if the user exists we create an attribute from their username.  So there shouldn't ever be a case where a userId exists without a list of attributes.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,56 @@ public Response getAllAttributes(
...
+    Set<Integer> userIds = new HashSet<>();
+    for (int i = 0; i < items.length; i++) {
+      try {
+        userIds.add(Integer.parseInt(items[i]));
+      } catch (NumberFormatException nfe) {
+      }
+      ;
+    }
+    for (Integer userId : userIds) {
+      checkIfUserExists(userId);
```
kvas-damian: I'm not sure if it's covered by any of our guidelines. But we return 404 if one from the list of users doesn't exist, which makes response not usable for end client. instead we could return a response when that particular user is marked as not existing and API Client still can use data about other users.
@jsutterfield any thoughts?

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java
bekcunning: I agree I don't think that the code can reach this condition.  Do you think it is better not to have an else in this case?

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,56 @@ public Response getAllAttributes(
...
+  )
+  @ApiResponses(value = {
+      @ApiResponse(code = 404, message = "No attributes found for this userId")
+  })
+  @Path("/bulk")
+  public Response getAllAttributesForMultipleUsers(
+      @QueryParam("ids") String arr
+  ) throws java.sql.SQLException {
+
+    String[] items = arr.split(",");
```
kvas-damian: I'm not an expert in REST bulk operation but did you consider passing userIds in following format: ```id=1&id=2&id=3```?
This sounds more natural for me, but I'm not sure if it isn't against rest standard...
```@QueryParam("id") final List<int> userIds``` should make a deal here.

Comment (bekcunning) [inline service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java]: Yes, I did consider it and actually did the extra work to make it work this way, as that was specified in the ticket.  I can go back to the other way if it is preferred.  @jsutterfield what do you think?

File: service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java [hunk 19]
```diff
@@ -127,6 +144,56 @@ public Response getAllAttributes(
   @GET
   @Timed
   @ApiOperation(
+      value = "Returns all available attributes for the specified list of users",
+      response = AllUserAttributesHalResponse.class
```
kvas-damian: I know this type of response was already defined but from API Client side it's very hard to consume, because client needs to iterate through the list to find right User. Instead of that our response could be based on HashMap where ```userId``` is a key. What do you think about it?

We could do the same with attributes, but this is topic for separate discussion - it's not backward compatible change. Ping @jsutterfield

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java
kvas-damian: I'd say, let's remove the ```if``` statement.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
kvas-damian: I see @garthwebb suggested this format in the ticket. @garthwebb, Was that on purpose or can we change it?

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
kvas-damian: BTW ```AllUserAttributesHalResponse``` provides ```List<UserAttributeHalResponse>``` under ```properties``` field. This is slightly misleading.

bekcunning: changed AllUserAttributesHalResponse to use a more relevant variable name.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
garthwebb: @kvas-damian, I think I was thinking about this line in "Restful Web Services" (https://docs.google.com/a/wikia-inc.com/file/d/0ByU0UlAOC541Mk1VRDJkVXk3TGc/edit, p223):
```
Use punctuation characters to separate multiple pieces of data at the same level of a
hierarchy. Use commas when the order of the items matters, as it does in latitude and
longitude: /Earth/37.0,-95.2. Use semicolons when the order doesn’t matter: /color-
blends/red;blue.
```
However many other places suggest `foo=1&foo=2` and if its easier to do it this way in Java, then we should do that.

kvas-damian: Punctuation char makes much more sense if we decide to put ```userIds``` as PathParams. Which sounds good for me. @garthwebb @bekcunning what do you think?

jsutterfield: It's interesting, for me using punctuation actually feels the most natural. Having multiple keys with the same name `id=1&id=2&id=3` seems awkward, and is a pattern you don't really encounter anywhere else. Twitter uses a similar approach for their [multiple users lookup](https://dev.twitter.com/rest/reference/get/users/lookup), and I believe I've seen similar solutions used by our vendors in their video APIs.

Personally I like having bulk operations and single resource operations separate, but I'm not expert on REST either and don't have super strong feelings either way. Either would be fine with me.

bekcunning: ok @garthwebb @kvas-damian @jsutterfield  where does that leave us?  The least amount of work is to use `foo=1&foo=2` so that there is no string parsing to get the values, but that work is already done if the punctuation format is preferred.  Who makes the call?

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
jsutterfield: I think using a hash map is a good idea. It will make it easier to consume. I also think we should call out whether a user exists or not. Currently if one of the users in the set doesn't exist, we return a 404 for the complete response. We should probably return data for all users that do exist, and mark those that don't. This is what I'm thinking the API response should look like:

```json
{
	_links: {
		self: {
			href: "/user/bulk"
		}
	},
	_embedded: {
		1: {
			exists: true,
			_links: {
				self: {
					href: "/user/1"
				}
			},
			userAttributeList: [
				{
					name: "username",
					value: "Jasonr"
				},
				{
					name: "avatar",
					value: "http://dev-vignette.wikia-dev.com/6a1f89df-07a5-41f5-939e-2ad834dd1b86"
				}
			],
		},
		4: {
			exists: true,
			_links: {
				self: {
					href: "/user/4"
				}
			},
			userAttributeList: [
				{
					name: "username",
					value: "MattisManzel"
				},
				{
					name: "nickname",
					value: ""
				}
			],
		},
		6: {
			exists: false
		}
	}
}
```

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
jsutterfield: So, I've actually come around and think we should use `userId=1&userId=2&userId=3`.

Doing so lets jersey take care of a lot of error handling for us. Currently if a client passes in `user/bulk?ids=` we get a null pointer exception on this line. Jersey will also wont' map to this method if the client tries to pass in `user/bulk?ids=1,2,three` so we don't have to worry about the error handling on L165.

If Java makes it easy to do it this way, and it's used elsewhere, then let's do that.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,57 @@ public Response getAllAttributes(
...
+    Set<Integer> userIds = new HashSet<>();
+    for (int i = 0; i < items.length; i++) {
+      try {
+        userIds.add(Integer.parseInt(items[i]));
+      } catch (NumberFormatException nfe) {
+        return StandardResponses.BadRequest("Invalid input");
+      }
+    }
+    for (Integer userId : userIds) {
+      checkIfUserExists(userId);
```
jsutterfield: See [this comment](https://github.com/Wikia/pandora/pull/492/files#r41376075). We can wrap each of these `checkIfUserExists` in a `try/catch`. If it throws an `WebApplicationException`, then we can mark that user doesn't exist.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,57 @@ public Response getAllAttributes(
...
+    }
+
+    Optional<List<UserWithAllAttributes>>
+        userWithAllAttributesList =
+        userAttributeDAO.getAllAttributesMultipleUsers(userIds);
+    if (!userWithAllAttributesList.isPresent()) {
+      return StandardResponses.NotFound(String.format("No attributes found for selected users"));
```
jsutterfield: No need to call `String.format` here, there's no interpolation taking place. Just using the string itself is fine.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
kvas-damian: @jsutterfield what do you think about dropping ```exists``` and just don't include that key in hashmap?

service/user-attribute/src/main/java/com/wikia/services/userattribute/core/UserWithAllAttributes.java
```diff
@@ -0,0 +1,32 @@
+package com.wikia.services.userattribute.core;
+
+import java.util.List;
+
+/**
+ * Created by rcunningham on 9/23/15.
+ */
```
kvas-damian: This data is available in git history ;)

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
kvas-damian: After small research I think that ```foo=1&foo=2``` makes most sense currently. It's not ideal but for now I don't think we have better option. http://www.l1ghtm4n.com/post/53259404576/patterns-for-rest-api-bulk-operations

service/user-attribute/src/main/java/com/wikia/services/userattribute/domain/responses/AllUserAttributesHalResponse.java
```diff
@@ -22,6 +22,6 @@
   public class UserAttributes {
 
     @NotNull
-    private List<UserAttributeHalResponse> properties;
+    private List<UserAttributeHalResponse> userAttributes;
```
kvas-damian: Shouldn't this be a ```List<AllAttributesHalResponse>```?

Or if we agree on https://github.com/Wikia/pandora/pull/492/files#r41165225
Map<Integer, AllAttributesHalResponse>

bekcunning: Actually given the comments from @jsutterfield  above I think we need a new HalResponse so we can add `boolean exists`   and also have `Map<Integer, AllAttributesHalResponse>`

kvas-damian:
@bekcunning Please merge dev into this branch. You'd need to update ```MemcachedUserAttributeDao``` to reflect your changes in ```UserAttributeDaoInterface```

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java
```diff
@@ -71,6 +70,33 @@ public UserAttribute map(Record record) {
...
+    for (Integer userId : userIds) {
+      Optional<UserAttribute> username = getUsername(userId);
+      userAttributeList = new ArrayList<>();
+      userAttributeList.add(username.get());
+      usersMap.put(userId, userAttributeList);
+    }
+    for (Record record : userAttributes) {
+      userAttributeList = usersMap.get(record.getValue(USER_ATTRIBUTE.USER_ID));
+      userAttributeList.add(createUserAttribute(record));
+    }
```
kvas-damian: Can we merge those two for loops into one?

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,62 @@ public Response getAllAttributes(
...
+      @QueryParam("id") Set<Integer> userIds
+  ) throws java.sql.SQLException {
+
+    Set<Integer> temp = new HashSet<>();
+
+    for(Integer id : userIds) {
+      try {
+        checkIfUserExists(id);
+        temp.add(id);
+      } catch(WebApplicationException e) {
```
kvas-damian: Can we think of a solution that doesn't require exception handling?

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,62 @@ public Response getAllAttributes(
...
+  public Response getAllAttributesForMultipleUsers(
+      @QueryParam("id") Set<Integer> userIds
+  ) throws java.sql.SQLException {
+
+    Set<Integer> temp = new HashSet<>();
+
+    for(Integer id : userIds) {
+      try {
+        checkIfUserExists(id);
+        temp.add(id);
```
kvas-damian: Maybe we could just remove missing user ids from ```userIds``` set?

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,62 @@ public Response getAllAttributes(
...
+        userAttributeDAO.getAllAttributesMultipleUsers(userIds);
+    if (!userWithAllAttributesMap.isPresent()) {
+      return StandardResponses.NotFound("No attributes found for selected users");
+    }
+
+    Representation representation = UserAttributeApplication.REPRESENTATION_FACTORY
+        .newRepresentation(createAllUserAttributeSelfUrl());
+
+
+    for (Integer userId : userWithAllAttributesMap.get().keySet()) {
```
kvas-damian: Maybe we could use ```userWithAllAttributesMap.map()``` for iterating?

kvas-damian:
I think we should update PR description.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
jsutterfield: That was my suggestion, since `checkIfUserExists` does what we want, it just throws exceptions rather than returning something like a boolean instead. Seems like since we have the code, we should try and re-use it?

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,62 @@ public Response getAllAttributes(
...
+
+      representation.withRepresentation("properties",
+                                        UserAttributeApplication.REPRESENTATION_FACTORY
+                                            .newRepresentation(
+                                                createAllAttributeSelfUrl(userId))
+                                            .withProperty("userId", userId)
```
jsutterfield: @bekcunning, I thought we decided userIds should be a key in the response? From [this](https://github.com/Wikia/pandora/pull/492#discussion_r41165225) conversation chain. These userIds are still being returned as just a property which means the client has to iterate over the entire list in {{properties}} to find a specific user. Or did something change?

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -55,7 +59,7 @@
             value = SwaggerBundleProvider.EXTERNAL_AUTH_NAME,
             type = "apiKey")
     })
-@Path("/user/{userId}")
+@Path("/user")
```
jsutterfield: I think this change broke the self URL for the `/user/{{userId}}` response. Now its missing the ```{{userId}}``` portion:

``` curl localhost:8080/user/1```
```json
{  
   "_links":{  
      "self":{  
         "href":"/user"
      }
   },
...
```

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
```diff
@@ -127,6 +144,62 @@ public Response getAllAttributes(
...
+      @ApiResponse(code = 400, message = "Invalid input provided")
+  })
+  @Path("/bulk")
+  public Response getAllAttributesForMultipleUsers(
+      @QueryParam("id") Set<Integer> userIds
+  ) throws java.sql.SQLException {
+
+    Set<Integer> temp = new HashSet<>();
+
+    for(Integer id : userIds) {
```
jsutterfield: Can we run reformat code on this file? There should be a space after this `for` since it's a control structure, not a function call. There may be a couple other small things like this it'll catch.

service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java
kvas-damian: maybe it could be just merged to:
````.withProperty(userId, userWithAllAttributesMap.get().get(userId)));````
?

File: service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java (hunk 25 above)
Comment (kvas-damian) [inline service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java]: While this method is throwing ```WebApplicationException``` I don't think we should be using it in this context. It could be more confusing in future if we for example decide to handle blocked users in different way.

File: service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java (hunk 26 above)
Comment (garthwebb) [inline service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java]: When @bekcunning and I tried using `.remove` it was causing an exception, I think because we were modifying the array while the loop was using it / had a reference to the element.

Comment (kvas-damian) [inline service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java]: You can use iterator and remove items then.

File: service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java (hunk 25 above)
Comment (jsutterfield) [inline service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java]: OK. Maybe we could do something like break up `checkIfUserExists()` into 2 methods, `userExists()` which just returns a boolean, and `assertUserExists()` which uses `userExists()` and throws an exception.

Comment (kvas-damian) [inline service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java]: Sounds good

File: service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java (hunk 24 above)
Comment (bekcunning) [inline service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/UserAttributeDAO.java]: I could not find a good way to do that since the records returned could possibly be zero if the user has no attributes but still exists in the wikiacities db - we create a username attribute in this case.  I think we need them both, but happy to hear any better solution that you have.

File: service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java (hunk 28 above)
Comment (bekcunning) [inline service/user-attribute/src/main/java/com/wikia/services/userattribute/resources/UserResource.java]: I've tried to find a way to return a key alongside a list of userAttributes but haven't found a way to do it.  .withProperty only allows string coupled with an object.  I suppose we can add the string value of the userId but that I think would be insufficient.  I'm open to hear more ideas.

File: service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/MemcachedUserAttributeDao.java [hunk 31]
```diff
@@ -97,13 +126,35 @@ protected String getMemcacheKey(int userId) {
...
+  @SuppressWarnings("unchecked")
+  protected HashMap<Integer, List<UserAttribute>> getAllAttributesMultipleUsersFromCache(
+      String cacheKey) {
+    HashMap<Integer, List<UserAttribute>> result = null;
+    Object cacheValue = memcachedClient.get(cacheKey);
+
+    if (cacheValue != null) {
+      try {
+        result = (HashMap<Integer, List<UserAttribute>>) cacheValue;
+      } catch (ClassCastException ex) {
```
WikiaTech: MAJOR Either log or rethrow this exception. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1166)

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/MemcachedUserAttributeDao.java
```diff
@@ -97,13 +126,35 @@ protected String getMemcacheKey(int userId) {
       } catch (ClassCastException ex) {
```
WikiaTech: MAJOR Either log or rethrow this exception. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1166)

WikiaTech:
SonarQube analysis reported 2 issues:
* MAJOR 2 major

Watch the comments in this conversation to review them.

WikiaTech:
SonarQube analysis reported 2 issues:
* MAJOR 2 major

Watch the comments in this conversation to review them.

WikiaTech:
SonarQube analysis reported 1 issue:
* MAJOR 1 major

Watch the comments in this conversation to review them.

bekcunning:
after discussion with @kvas-damian I removed the code to cache the multiple users request since it possibly requires many calls to memcache so would be similar in response time to the one db call required on the fly.

kvas-damian:
code LGTM, can you make sure that tests failure is not connected to code changes we introduced?

service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/HalResponseHelper.java
```diff
@@ -42,4 +43,56 @@ public static UserAttribute extractUserAttributeFromResponseMap(Map<String, Obje
         .add(extractUserAttributeFromResponseMap(resource2))
         .build();
   }
+
+  public static Integer extractUserIdFromResponseMap(Map<String, Object> resourceMap) {
+    Integer userId = 0;
+    for (String name : resourceMap.keySet()) {
+      if (name != "_links") {
```
WikiaTech: CRITICAL Replace "==" and "!=" by "equals()" and "!equals()" respectively to compare these strings. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AStringEqualityComparisonCheck)

service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/HalResponseHelper.java
```diff
@@ -42,4 +43,56 @@ public static UserAttribute extractUserAttributeFromResponseMap(Map<String, Obje
...
+
+        }
+
+      }
+    }
+    return userAttributeList;
+
+  }
+
+  public static HashMap<Integer, List<UserAttribute>> extractAllUserAttributeListFromResponseMap(
```
WikiaTech: MAJOR The return type of this method should be an interface such as "Map" rather than the implementation "HashMap". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1319)

service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/HalResponseHelper.java
```diff
@@ -42,4 +43,56 @@ public static UserAttribute extractUserAttributeFromResponseMap(Map<String, Obje
...
+      if (name != "_links") {
+        List<HashMap<String, String>>
+            userMapList =
+            (List<HashMap<String, String>>) resourceMap.get(name);
+
+        for (HashMap<String, String> usersMap : userMapList) {
+          UserAttribute
+              userAttribute =
+              new UserAttribute(usersMap.get("name"), usersMap.get("value"));
+          if (!userAttributeList.contains(userAttribute)) {
```
WikiaTech: MINOR Refactor this code to not nest more than 3 if/for/while/switch/try statements. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS134)

service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/HalResponseHelper.java
```diff
@@ -42,4 +43,56 @@ public static UserAttribute extractUserAttributeFromResponseMap(Map<String, Obje
...
+    }
+    return userId;
+  }
+
+  public static List<UserAttribute> extractUserAttributeListFromResponseHashMap(
+      Map<String, Object> resourceMap) {
+    List<UserAttribute> userAttributeList = new ArrayList<>();
+    for (String name : resourceMap.keySet()) {
+
+      if (name != "_links") {
```
WikiaTech: CRITICAL Replace "==" and "!=" by "equals()" and "!equals()" respectively to compare these strings. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AStringEqualityComparisonCheck)

WikiaTech:
SonarQube analysis reported 6 issues:
* CRITICAL 2 critical
* MAJOR 3 major
* MINOR 1 minor

Watch the comments in this conversation to review them.
Note: the following issues could not be reported as comments because they are located on lines that are not displayed in this pull request:
* MAJOR [2 duplicated blocks of code must be removed.](https://github.com/Wikia/pandora/blob/f85b1c5f951ee3695c70751f51d3e0bc7c58e20c/service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/UserResourceTest.java) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=common-java%3ADuplicatedBlocks)

service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/HalResponseHelper.java
```diff
@@ -42,4 +43,56 @@ public static UserAttribute extractUserAttributeFromResponseMap(Map<String, Obje
         .add(extractUserAttributeFromResponseMap(resource2))
         .build();
   }
+
+  public static Integer extractUserIdFromResponseMap(Map<String, Object> resourceMap) {
+    Integer userId = 0;
+    for (String name : resourceMap.keySet()) {
+      if (!name.equals("_links")) {
```
WikiaTech: MAJOR Move the "_links" string literal on the left side of this string comparison. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1132)

service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/HalResponseHelper.java
```diff
@@ -42,4 +43,56 @@ public static UserAttribute extractUserAttributeFromResponseMap(Map<String, Obje
...
+    }
+    return userId;
+  }
+
+  public static List<UserAttribute> extractUserAttributeListFromResponseHashMap(
+      Map<String, Object> resourceMap) {
+    List<UserAttribute> userAttributeList = new ArrayList<>();
+    for (String name : resourceMap.keySet()) {
+
+      if (!name.equals("_links")) {
```
WikiaTech: MAJOR Move the "_links" string literal on the left side of this string comparison. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1132)

WikiaTech:
SonarQube analysis reported 2 issues:
* MAJOR 2 major

Watch the comments in this conversation to review them.

service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/UserResourceTest.java
```diff
@@ -153,6 +171,65 @@ public void testGetAttributesForInvalidUser() throws SQLException {
...
+    assertThat(response.getStatus(), equalTo(Response.Status.OK.getStatusCode()));
+  }
+
+  @Test
+  public void getAllAttributesMultipleUsersOneInvalid() throws SQLException {
+    Response
+        response =
+        resources.client()
+            .target(String
+                        .format("/user/bulk?id=%s&id=%s&id=%s", VALID_USER_ID, INVALID_USER_ID,
```
jsutterfield: So this isn't a blocker, but just as an FYI since we're interpolating integers into this string, we should probably be using `%d` which handles digits, rather than `%s` which is meant for strings. `String.format` uses a [printf format string](https://en.wikipedia.org/wiki/Printf_format_string) which takes all kinds of options other than just `%s`. You can also find more info in the [Java docs](http://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html).

service/user-attribute/src/main/java/com/wikia/services/userattribute/daos/MemcachedUserAttributeDao.java
```diff
@@ -103,13 +114,15 @@ protected String getMemcacheKey(int userId) {
...
```
WikiaTech: MAJOR Either log or rethrow this exception. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1166)

service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/HalResponseHelper.java
```diff
@@ -42,4 +43,56 @@ public static UserAttribute extractUserAttributeFromResponseMap(Map<String, Obje
...
+
+        }
+
+      }
+    }
+    return userAttributeList;
+
+  }
+
+  public static HashMap<Integer, List<UserAttribute>> extractAllUserAttributeListFromResponseMap(
```
WikiaTech: MAJOR The return type of this method should be an interface such as "Map" rather than the implementation "HashMap". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1319)

service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/HalResponseHelper.java
```diff
@@ -42,4 +43,56 @@ public static UserAttribute extractUserAttributeFromResponseMap(Map<String, Obje
...
+      if (!name.equals("_links")) {
+        List<HashMap<String, String>>
+            userMapList =
+            (List<HashMap<String, String>>) resourceMap.get(name);
+
+        for (HashMap<String, String> usersMap : userMapList) {
+          UserAttribute
+              userAttribute =
+              new UserAttribute(usersMap.get("name"), usersMap.get("value"));
+          if (!userAttributeList.contains(userAttribute)) {
```
WikiaTech: MINOR Refactor this code to not nest more than 3 if/for/while/switch/try statements. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS134)

WikiaTech:
SonarQube analysis reported 7 issues:
* MAJOR 5 major
* MINOR 2 minor

Watch the comments in this conversation to review them.
Note: the following issues could not be reported as comments because they are located on lines that are not displayed in this pull request:
* MAJOR [2 duplicated blocks of code must be removed.](https://github.com/Wikia/pandora/blob/663a9130bc9522b98873bafd6bba82edaf37c0a2/service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/UserResourceTest.java) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=common-java%3ADuplicatedBlocks)
* MINOR [Define a constant instead of duplicating this literal "PATCH" 4 times.](https://github.com/Wikia/pandora/blob/663a9130bc9522b98873bafd6bba82edaf37c0a2/service/user-attribute/src/test/java/com/wikia/services/userattribute/resources/UserResourceTest.java#L339) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1192)

jsutterfield:
LGTM

PR #496, opened 2015-10-09
New Service "Template Classification"

sqreek:
Init PR with new template classification service (name is under development), with boiler plate for new service.

//cc @Wikia/services-team @lgarczewski @SebastianMarzjan @idradm


SebastianMarzjan:
As we talked, naming being the hardest part of starting this work... The name suggests it's doing template classification, the main purpose is to store the classification information so maybe actually template-classes would be a name that aligns with existing names: user-attributes, etc.

drsnyder:
Please provide some details in the description.

sqreek:
Template classification was provided by @lgarczewski :)
We should schedule meeting (at least 2 hour) that everyone provide at least three name proposition.
Another aproach is to start using md5(new service name) for naming, and write it in our guidelines as a only one way to name new services :)

We will talk about name in monday, and we rename it.

drsnyder:
```
$ echo "template classification v1" | shasum
a4579acbbe9a4b52ca733f58a3ecf3adca4c363e
```

sqreek:
We want to work on separate branch until finish
https://github.com/Wikia/pandora/pull/499

PR #498, opened 2015-10-12
Reverse lookup and preference deletion resource

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-732

Implementing reverse-lookup for finding wikis where any user has a given local preference set to a given value. We need this in [`FounderEmails`](https://github.com/Wikia/app/blob/dev/extensions/wikia/FounderEmails/FounderEmails.class.php#L51), which returns that list when sending out founder emails. 

Also implementing a resource to delete all of a user's preferences in `DELETE /{userId}`. We use this [RemoveUserBase](https://github.com/Wikia/app/blob/dev/maintenance/wikia/RemoveUserBase.class.php#L65), which gets run when renaming a user.

TODO:
- [x] tests for `DELETE` resource
- [ ] add index on `local_preference` table to make reverse lookups faster?


service/user-preference/src/main/java/com/wikia/userpreference/resources/ReverseLookupResource.java
```diff
@@ -0,0 +1,64 @@
...
+      @PathParam("preferenceName")
+      String preferenceName,
+      @ApiParam(value = "The preference value", defaultValue = "1")
+      @QueryParam("value")
+      @DefaultValue("1")
+      String value
+  ) {
+
+    return storedPreferences.findWikisWithLocalPreferenceValue(preferenceName, value)
+        .map(l -> Response.ok(l).build())
```
sqreek: letter l is very confusing (or is it 1)

service/user-preference/src/main/java/com/wikia/userpreference/dao/StoredPreferences.java
```diff
@@ -10,4 +11,8 @@
   boolean add(int userId, UserPreferences preferences);
 
   boolean set(int userId, UserPreferences preferences);
+
+  boolean delete (int userId);
+
+  Optional<List<String>> findWikisWithLocalPreferenceValue(String preferenceName, String value);
```
sqreek: why using string?

service/user-preference/src/main/java/com/wikia/userpreference/dao/StoredPreferences.java
```diff
@@ -10,4 +11,8 @@
   boolean add(int userId, UserPreferences preferences);
 
   boolean set(int userId, UserPreferences preferences);
+
+  boolean delete (int userId);
```
sqreek: I think about deleteAllPreference to be more verbose, but deleteAll should works well too

service/user-preference/src/main/java/com/wikia/userpreference/dao/StoredPreferences.java
nmonterroso: After speaking with @cmiller-wikia we decided that all IDs should be of type `string`, not `int`. We have an item in our backlog to update this in Helios, and we should do the same for the user preference service. This is just getting ahead of it for this use-case.

service/user-preference/src/main/java/com/wikia/userpreference/resources/ReverseLookupResource.java
nmonterroso: it definitely cannot be a `1` since you can't start a variable name with a number. But I see your point. I updated this in 93786f1

service/user-preference/src/main/java/com/wikia/userpreference/dao/StoredPreferences.java
nmonterroso: updated in 8651a71

service/user-preference/src/main/java/com/wikia/userpreference/dao/StoredPreferencesDAO.java
```diff
@@ -119,6 +119,29 @@ public boolean set(int userId, UserPreferences preferences) {
     return true;
   }
 
+  @Override
+  public boolean delete(int userId) {
+    return set(userId, new UserPreferences());
```
sqreek: I'm not sure that this is a valid aproach, 
but in this moment I can not provide a valid point why not too.

I agree with this solution after tests will be provided that covers distinguish between delete and set empty.

ArturKlajnerok:
Lgtm :+1:

PR #500, opened 2015-10-12
[SERVICES-895] New external-auth service

ArturKlajnerok:
This introduces base code for new `External Auth` service. We can start building several parts of functionality based on this service.

/cc @Wikia/services-team


service/external-auth/README.md
```diff
@@ -0,0 +1,25 @@
+# ExternalAuth
```
nmonterroso: Can we fill in these parts?

service/external-auth/src/main/resources/banner.txt
```diff
@@ -0,0 +1,13 @@
+███████╗██╗  ██╗████████╗███████╗██████╗ ███╗   ██╗ █████╗ ██╗          █████╗ ██╗   ██╗████████╗██╗  ██╗
```
nmonterroso: YES

settings.gradle
```diff
@@ -62,4 +62,4 @@ include 'workers:xwiki-feeder'//no tests
 include 'workers:xwiki-worker'//no tests
 include 'workers:article-feeder'//no tests
 include 'workers:article-snippet-worker'
-
+include 'service:external-auth'
```
nmonterroso: hm as part of this can you put the `workers` entries above the `service` entries? it'd be nice to keep these all grouped together.

nmonterroso:
minor comments LGTM

settings.gradle
ArturKlajnerok: Moved and added more order to services list here : e3120f0

service/external-auth/README.md
ArturKlajnerok: Filled description and will fill other paragraphs when specific functionality will arrive : 4f64692

PR #502, opened 2015-10-12
[SERVICES-901] Table schema for external auth with google.

ArturKlajnerok:
SQL create table schema for linking google account.

/cc @Wikia/services-team


nmonterroso:
LGTM

service/external-auth/sql/schema.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE `google_account` (
+  `wikia_user_id` varchar(255) NOT NULL,
```
sqreek: do we realy need 255 chars here?
9 999 999 999 is 11 chars. 
If we ever want to go GUID we have 32, or 64.

With assumption that we will have 100 000 users in 1 year this will be a ~43 MB difference.

ArturKlajnerok: The general recommendation is to store id as string and we can assume that they can also be some random unique hashes there. I could make it shorter but to be honest, I'm not sure what's the optimal length at the moment.

service/external-auth/sql/schema.sql
```diff
@@ -0,0 +1,7 @@
...
+  `google_user_id` varchar(255) NOT NULL,
+  `connection_date` datetime DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`wikia_user_id`),
```
drsnyder: Should this be `PRIMARY KEY ('wikia_user_id', 'google_user_id')`? Will we ever do a lookup only on `google_user_id`?

nmonterroso: I think it's fine as `wikia_user_id`, since if we did a combination of the two a user could technically have multiple google accounts linked, which i don't think we want to support

PR #511, opened 2015-10-13
upgrading version to latest and adding support for library option

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-899

Now that the core swagger-codegen library supports different "libraries" per language, we should upgrade to the latest and use that to generate a helios client. This upgrades the `central-swagger-ui` to the latest version and accounts for the `library` configuration properly.


ArturKlajnerok:
Lgtm :+1:

PR #513, opened 2015-10-14
Drop memory and CPU allocation for UAS

jsutterfield:


service/user-attribute/build.gradle
```diff
@@ -25,8 +25,8 @@ deployments {
             }
         }
         resources {
-            cpus = 2
-            mem = 500
+            cpus = 1
+            mem = 300
```
jsutterfield: `applicationDefaultJvmArgs` already set to 300MB

PR #514, opened 2015-10-14
Dat 3022

rybmat:


workers/xwiki-feeder/src/main/java/com/wikia/workers/xwikifeeder/XwikiFeeder.java
```diff
@@ -28,6 +32,20 @@ protected JSONObject getData(Integer wid, Integer pid) throws FeederException {
         .target(WikiaPhp.class, String.format("http://%s/", domain))
         .get().body()
         .toString();
-    return new JSONObject(data);
+
+    JSONObject jsonmsg = new JSONObject(data);
+    Integer wikiId = jsonmsg.getInt("wid");
+    String indexingDate = this.currentIsoDate();
```
SebastianMarzjan: We probably don't need those 2 vars - each is used only once below

workers/abstract-solr-worker/src/main/java/com.wikia.workers.abstractsolrworker/AbstractSolrWorker.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+    return false;
+  }
+
+  private void scheduleUpdate() {
+    this.scheduledUpdate = this.timer.schedule(new Runnable() {
+      @Override
+      public void run() {
+        updateLock.lock();
+        getBatchUpdate().send();
```
SebastianMarzjan: What do you think about logging the timed updates vs regular updates

rybmat: It makes sense, but when I add logging on the worker level then all updates will be logged twice because of logging in BatchUpdate class. Should I keep both logging?

SebastianMarzjan: Frankly speaking we need only one place, maybe just updating the logging in BatchUpdate will be sufficient (changing this send() to sendUpdate() and logging sendUpdate callee there)

PR #516, opened 2015-10-14
Dat 3381 add swagger docs

idradm:
https://wikia-inc.atlassian.net/browse/DAT-3381


service/template-classification-storage/src/main/java/com/wikia/services/templateclassificationstorage/resources/TemplateClassificationStorageResource.java
```diff
@@ -61,9 +91,18 @@ public Response getTemplateDetails(@PathParam("wiki_id") int wikiId,
...
-  public Response insertTemplateDetails(@PathParam("wiki_id") int wikiId,
-                                        @PathParam("page_id") int pageId,
-                                        TemplateTypeProvider data
+  @ApiOperation(value = "Save template type data",
+      notes = "Save template types with information about provider and its origin",
+      consumes = MediaType.APPLICATION_JSON,
+      produces = MediaType.APPLICATION_JSON,
+      response = boolean.class)
+  @ApiResponses(value = {
+      @ApiResponse(code = 201, message = "Template data saved", response = boolean.class),
```
sqreek: If we want to return 201 created, we need to return url for new resource that was created.
It's a little confusion, because in some situation we doesn't get same value what we just pushed.

PR #458, opened 2015-09-17
Use latest Swagger without using Current Swagger bundle

pchojnacki:
Its my intution that we would benefit long term from integration of most of current swagger bundle into pandora codebase. As that would make updating upstream Swagger code much easier.

Additionally this PR is first step in configuring Building both swagger specs and swagger generated source code during gradle build.
This will allow for compile time validation of interface compatibility.

This change was mostly inspired by missing support for File uploads in old swagger.
Old swagger bundle is still maintained as a means to provide seamless and gradual upgrade path for other services. Also inbetween version swagger changed their package names: https://github.com/swagger-api/swagger-core/pull/1128/files

Currently the biggest feature missing is integrated swagger ui, so right now its only for those services that are OK with that.

/cc: @nmonterroso @drsnyder @sqreek @harnash  @ArturKlajnerok

PR #521, opened 2015-10-14
Service Gateway Support

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-899

Adding auto-generated client for Helios to `lib/service-gateway/helios-gateway/src/main/java/com/wikia/swagger/generated`, which should serve as an example of how to provide gateways to services within the Pandora repo. To do this, I chose to move service resolution from the `lib/dropwizard/consul` project into the new `lib/service-gateway/service-resolution` project, since it makes more sense to be there, and importing the consul resolution stuff should not require importing all of Dropwizard.

For now, since we haven't yet created the functionality required for fetching a token for a user, I've added `login` to the `HeliosGateway` interface and it's available at `GET /login?username=<username>&password=<password>` in the `external-auth` service, and it will output an access token. We can remove this once Helios contains the functionality we need.

TODO:
- [x] integrate work @pchojnacki did in Wikia/pandora#458 to get a generated client into HeliosGateway


lib/service-gateway/helios-gateway/helios-swagger-client/resources/swagger.json
```diff
@@ -0,0 +1,910 @@
+{
```
sqreek: Can we set a one practice for code generation?
In database aproach, where we don't change it a lot, we download schema from database.
Here, where helios is change sometime (and it is more possible helios will change than database), we put source for code generation into repository.

sqreek: btw. I prefer this aproach

nmonterroso: I went back and forth on this a few times, here's where I ended up:

There's no reason we couldn't commit the code to the Pandora repo, but if we do that we have to know which version of the Helios API we're pinned to, and how would we do that? We'll probably have to have the `swagger.json` file included in the repo so we know if it's behind or not. Assuming that, why have the generated files in the repo, since they will always be generated the same way, and don't require an external resource to do the generation (as opposed to jooq, which currently needs a db connection)?

lib/service-gateway/helios-gateway/src/main/java/com/wikia/gateway/helios/SwaggerHeliosGateway.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public String login(String username, String password) throws GatewayException {
+    try {
+      return getTokenApi().login(username, password).getAccessToken();
+    } catch (ApiException e) {
+      throw new GatewayException(e.getMessage(), e.getCode());
+    }
+  }
+
+  private TokenApi getTokenApi() throws GatewayException {
```
sqreek: is this necessery to wrap of wrap of wrap object creation?

lib/service-gateway/service-gateway-base/src/main/java/com/wikia/gateway/annotation/ServiceEnvironment.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.google.inject.BindingAnnotation;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@BindingAnnotation
+@Target({FIELD, PARAMETER, METHOD, TYPE})
+@Retention(RUNTIME)
+public @interface ServiceEnvironment {
```
sqreek: :+1:

service/external-auth/src/main/java/com/wikia/services/externalauth/resources/ExternalAuthResource.java
```diff
@@ -15,27 +16,37 @@
...
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 
 @Api(
     basePath = "/",
     value = "/HelloWorld",
     description = "Says hello")
-@Path("/helloworld/{name}")
+@Path("/")
 @Produces(RepresentationFactory.HAL_JSON)
```
sqreek: We should remove RepresentationFactory from this service, and template also. 

but this is not a scope of this PR.

settings.gradle
```diff
@@ -49,14 +53,14 @@ include 'workers:article-snippet-worker'
 include 'service:central-swagger-ui'
 include 'service:clickstream'
 include 'service:clickstream:api'//no tests
-include 'service:community-data'
+//include 'service:community-data'
```
sqreek: what is a reason to disable this service?

settings.gradle
```diff
@@ -49,14 +53,14 @@ include 'workers:article-snippet-worker'
...
 include 'service:dc-file-sync'
 include 'service:discussion'
 include 'service:example'
 include 'service:external-auth'
 include 'service:facade'//no tests
 include 'service:geoip'
-include 'service:job-queue'
+//include 'service:job-queue'
```
sqreek: what is a reason to disable this service?

lib/service-gateway/helios-gateway/helios-swagger-client/resources/swagger.json
nmonterroso: Well after I say that jenkins is having trouble generating the code, so i'll commit it :)

settings.gradle
nmonterroso: unused in production, it was an intern project that isn't deployed to production. its tests are failing, and shouldn't block any merges.

settings.gradle
nmonterroso: the service isn't serving any traffic. it's deployed to production, but none of the 3 instances has received any requests in 3 months. It has failing tests, but since it's unused it shouldn't block merges.

lib/service-gateway/helios-gateway/src/main/java/com/wikia/gateway/helios/SwaggerHeliosGateway.java
nmonterroso: huh?

lib/service-gateway/helios-gateway/src/main/java/com/wikia/gateway/helios/SwaggerHeliosGateway.java
```diff
@@ -0,0 +1,51 @@
...
+      ServiceProvider serviceProvider,
+      @ServiceEnvironment String environment) {
+
+    this.serviceProvider = serviceProvider;
+    this.environment = environment;
+  }
+
+  public String login(String username, String password) throws GatewayException {
+    try {
+      return getTokenApi().login(username, password).getAccessToken();
```
sqreek: I can't find a right word for replacement but TokenApi does not sound as something I can login on it.
tokenApiClient. ?
heliosApiClient?
heliosClient?

lib/service-gateway/helios-gateway/src/main/java/com/wikia/gateway/helios/SwaggerHeliosGateway.java
nmonterroso: I mean I guess I could have the whole class in one function? There are different `*Api` classes, all of which need an `ApiClient`, so it's just to separate responsibilities in code.

lib/service-gateway/helios-gateway/helios-swagger-client/resources/swagger.json
sqreek: :) :+1:

lib/service-gateway/helios-gateway/src/main/java/com/wikia/gateway/helios/SwaggerHeliosGateway.java
sqreek: at least one level of wrapping could be removed
getBasePath have more logic, so it's ok to keep it there, but we can remove getApiClient()
``` private TokenApi getTokenApi() throws GatewayException {
    return new TokenApi(
        new ApiClient()
            .setBasePath(getBasePath())
    );
  }```

settings.gradle
sqreek: :+1:

settings.gradle
sqreek: :+1:

lib/service-gateway/helios-gateway/src/main/java/com/wikia/gateway/helios/SwaggerHeliosGateway.java
nmonterroso: There are different `*Api` classes in the Helios API: `UserApi`, `TokenApi`, `FacebookApi`, `InfoApi`, and `UsernameApi`. It doesn't make sense to do `new ApiClient().setBasePath(...)` for each of them. We should have a consolidated place to create a basic `ApiClient` with common functionality (everything will need the base path), and then each method can modify it as needed (for example, adding authentication).

ArturKlajnerok:
Can we get some tests for `SwaggerHeliosGateway` ? Let's also make sure all tests are passing, other then that LGTM :+1:

nmonterroso:
We can't make sure all tests are passing since they're failing for reasons of their own, not for reasons introduced by this PR. Or do you mean for `SwaggerHeliosGateway`?

nmonterroso:
@ArturKlajnerok we should merge this PR and then I can add tests for the `SwaggerHeliosGateway`, since any push here results in tests being run for the entire repo since there are some changes to `service-core`. cool?

ArturKlajnerok:
@nmonterroso That makes sense :+1:

nmonterroso:
https://wikia-inc.atlassian.net/browse/SERVICES-927 is the ticket for writing tests

PR #528, opened 2015-10-15
DAT-3371 Refactor tests for Template Classification Storage

sqreek:
test refactor


sqreek:
Self :hand: and merge
@idradm aprove :hand:

PR #529, opened 2015-10-15
[SERVICES-902] DB access layer for external auth service

ArturKlajnerok:
This introduces DAO for linked google accounts with interface that enables saving new linked account and getting existing ones based on google account ID. Jooq classes are auto-generated and committed as separate package.

/cc @Wikia/services-team


nmonterroso:
LGTM

PR #531, opened 2015-10-19
DAT-3367: Move "other" template type to constant in classification test

lukaszkonieczny:
Move "other" template type to constant in classification test


sqreek:
LGTM

PR #532, opened 2015-10-19
Split ConsulGuiceApplication into Consul and PandoraGuice Bundles

sqreek:
This PR split and change logic from ConsulGuiceApplication into ConsulBundle and PandoraGuiceBundle.

Example of usage is in https://github.com/Wikia/pandora/pull/533

ping @Wikia/services-team


drsnyder:
Most if not all services extend `ConsulGuiceApplication`. Are we going to change them in the same way that you did the `UserPreferenceApplication` in #533?

sqreek:
yes, this is a plan

sqreek:
also, add a BoilerPlateBundle with BoilerPlateBundleBuilder that contains all common bundles, and options in one place. (Like SwaggerBundle, GuiceBundle, ConsulBundle, ExceptionMapper, and optional JooqInstaller, ApiBundle (if this will exists in a future))

nmonterroso:
It makes sense to split these into modules. It's very nice to inherit from the normal `Application` rather than a home-grown solution. My only concern is that they're separate but we really can't have one without the other, as seen in Wikia/pandora#533

PR #536, opened 2015-10-19
Change the "helios" URI in the swagger docs to /auth

drsnyder:
Change `/helios` to `/auth`. This [PR](https://github.com/Wikia/api-gateway/pull/37) is removing access to `/helios`.

@Wikia/services-team


ArturKlajnerok:
Lgtm :+1:

nmonterroso:
Note that this change will NOT change `central-swagger-ui` to use `/auth`, this change is only for a test. If we want to change the frontend we need to change how we look up services to be configureable, since in Consul the service name is still `helios`, and the logic is currently `return externalBaseUrl + serviceName + "/" + swaggerPath;`, so this would have to change to be configureable (much like the other settings for a service)

See [`SwaggerDocProviderImpl`](https://github.com/Wikia/pandora/blob/master/service/central-swagger-ui/src/main/java/com/wikia/services/centralswaggerui/domain/SwaggerDocProviderImpl.java)

PR #542, opened 2015-10-20
GoogleAuth view for account linking/logging in

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-926

Adding view for allowing users to link their wikia <-> google accounts and also to log in


drsnyder:
lgtm

PR #543, opened 2015-10-20
Allow the service path to be configured in consul

drsnyder:
This allows the service path (services.wikia.com/<path>/) to be configured via consul. This will allow us to alias services that are in transition by pointing the swagger ui to a different service name than what was tagged in consul.

For example, by setting `kv/config/central-swagger-ui/dev/helios/service_path/` to `auth` the swagger UI will look at `http://services.wikia.com/auth` instead of `http://services.wikia.com/helios`.

This is a follow on to the work done [here](https://github.com/Wikia/api-gateway/pull/37) so we can point the helios docs in the swagger UI to auth since helios is blocked.

/cc @nmonterroso


nmonterroso:
LGTM

PR #545, opened 2015-10-21
multibinding

rybmat:
allowing to define multiple bindings for rabbit queue


SebastianMarzjan:
As long as we don't expect semicolons in keys, LGTM :)

idradm:
LGTM

rybmat:
semicolons are just proposal, if you think that some other character would be better just write it here ;)

PR #553, opened 2015-10-22
add empty gradle properties for not block other developers

sqreek:
when new database credential was added into repo, it required to add new entries into gradle.properties.
Added empty ones will stop blocking other developers that don't know about it.

ping @Wikia/services-team


pchojnacki:
does it work ?

pchojnacki:
I mean are you able to override these values ? If so LGTM

sqreek:
https://docs.gradle.org/current/userguide/build_environment.html

tested and works

btw. ping @idradm

nmonterroso:
I hope people don't fill in appropriate values and commit/push them upstream. Can you add a comment to each of the gradle files to indicate that? We're creating more tribal knowledge here, where people have to know "yes these gradle.properties exist but I shouldn't fill them in"

PR #554, opened 2015-10-22
add information as a value for gradle properties mandatory for gradle tasks

sqreek:
ping @nmonterroso


nmonterroso:
I was thinking something like:
```
// these values should be overwritten in a non-version-controlled gradle.properties file
// this file is only here to indicate that this project has a dependency on these gradle variables
centralMysqlDbHost=
centralMysqlUsername=
centralMysqlPassword= 
```

PR #544, opened 2015-10-20
[SERVICES-172] External auth google support

ArturKlajnerok:
First resource for `extarnal auth` that enables login with Google identity provider. Resource is based on google auth service that combines request to `google api`, `db storage` and `helios` to provide access token for user that posses valid `google id token`. Uses google gateway that is responsible for google tokens validation and extraction of authenticated user data. 

- [x] Google api gatway
- [x] Google auth service
- [x] Google token resource
- [x] New general exception for internal server errors
- [x] Response with `access_token` cookie
- [x] Unit and integration tests

/cc @Wikia/services-team

PR #555, opened 2015-10-22
Whoami service scaffolding

drsnyder:
Scaffolding for the whoami service. It currently works when passed the `access_token` cookie or the `X-Wikia-AccessToken` header. It only returns the user id of the logged in user. This is probably sufficient for the current needs.

https://wikia-inc.atlassian.net/browse/SERVICES-930

TODO:
 * [x] add controller tests
 * [x] add service tests

/cc @Wikia/services-team


service/whoami/src/main/java/com/wikia/services/whoami/resources/WhoamiResource.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @GET
+  @Timed
+  @ApiOperation(
+      value = "Returns information about the currently logged in user",
+      notes = "Only successful when the request is authenticated")
+  @ApiResponses(value = {
+      @ApiResponse(code = 200, message = "successfully looked up the currently logged in user", response = User.class)
+  })
+  public Response getCurrentlyLoggedInUser(@Context HttpServletRequest httpRequest) {
```
nmonterroso: Couldn't we leverage existing code here and use `@Auth AuthenticatedUser(required = false)`, and let the api gateway do the validation of the access token (which it does anyway)? We don't return anything besides the user id anyway, so we could remove a lot of code and have it be something like

```java
public Response getCurrentlyLoggedInUser(@Auth(required = false) AuthenticatedUser user) {
  if (user == null) {
     // anon, return 0
  }

  // authenticated, return response with user.Id
}
```

lib/service-gateway/helios-gateway/src/main/java/com/wikia/gateway/helios/HeliosGateway.java
```diff
@@ -1,7 +1,9 @@
 package com.wikia.gateway.helios;
 
 import com.wikia.gateway.GatewayException;
+import com.wikia.services.swagger.helios.client.model.TokenResponse;
 
 public interface HeliosGateway {
   String getUserAccessToken(String userId) throws GatewayException;
+  TokenResponse getTokenInfo(String token) throws GatewayException;
```
nmonterroso: If we go down this path we shouldn't return the `TokenResponse` object since that's Swagger-specific. @ArturKlajnerok created a Wikia domain response object [`AccessData`](https://github.com/Wikia/pandora/pull/544/files#diff-b33725be5031da8592d982dff6bc4b49)

service/whoami/src/main/java/com/wikia/services/whoami/resources/WhoamiResource.java
nmonterroso: Or we could omit the `(required=false)` part and let Dropwizard respond with a `401` for us

drsnyder: Duh. Somehow I completely forgot about that.

lib/service-gateway/helios-gateway/src/main/java/com/wikia/gateway/helios/HeliosGateway.java
drsnyder: Yeah, I saw that. I can yank this out now that we can use the gateway provided header.

service/whoami/src/main/java/com/wikia/services/whoami/resources/WhoamiResource.java
```diff
@@ -0,0 +1,43 @@
...
+      @ApiResponse(code = 401, message = "Unauthorized access"),
+      @ApiResponse(code = 200, message = "user is logged in", response = User.class)
+  })
+  public Response getCurrentlyLoggedInUser(@Auth AuthenticatedUser user) {
+    System.out.println(user);
+    User loggedInUser = User.builder()
+          .userId(user.getUserId().toString())
+          .build();
+    return Response.ok(loggedInUser).build();
+  }
```
drsnyder: @nmonterroso ccfcb9e. I think it's as simple as that for basic functionality.

service/whoami/README.md
```diff
@@ -0,0 +1,25 @@
+# Whoami
+Please enter a short, general description about what this service does
```
sqreek: Please enter a short, general description about what this service does

service/whoami/src/main/java/com/wikia/services/whoami/resources/WhoamiResource.java
```diff
@@ -0,0 +1,44 @@
...
+  @CacheControl(noCache = true, maxAge = 0)
+  @ApiOperation(
+      value = "Returns information about the currently logged in user",
+      notes = "Only successful when the request is authenticated")
+  @ApiResponses(value = {
+      @ApiResponse(code = 401, message = "Unauthorized access"),
+      @ApiResponse(code = 200, message = "user is logged in", response = User.class)
+  })
+  public Response getCurrentlyLoggedInUser(@Auth AuthenticatedUser user) {
+    System.out.println(user);
```
sqreek: hmm

nmonterroso:
Are failed tests related to the changes? if not lgtm

drsnyder:
Looks like some of them are timing out after 10 minutes.

PR #561, opened 2015-10-27
DAT-3418 get providers data from master on save

idradm:
https://wikia-inc.atlassian.net/browse/DAT-3418


idradm:
@lukaszkonieczny @sqreek take a look

SebastianMarzjan:
LGTM

sqreek:
Good enough for merge it, but it is a bit confusing for next resource that will use it, (or next implementation of storage), but...
yea, probably it will never happen...
so LGTM

PR #562, opened 2015-10-27
Global permission resources

nmonterroso:
@Wikia/services-team @jcellary 
https://wikia-inc.atlassian.net/browse/SERVICES-948

Adding resources to query for a user's global permissions, both for listing and for querying if a user has a specific global permission.


service/user-permissions/src/main/java/com/wikia/services/userpermissions/service/MWUserPermissionsService.java
```diff
@@ -19,6 +19,8 @@
 
 public class MWUserPermissionsService implements UserPermissionsService {
 
+  private static final int GLOBAL_PERMS_WIKI_ID = 177;
```
drsnyder: Is there one "city" that has all of the perms?

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
```diff
@@ -164,4 +177,69 @@ public Response getUserPermissions(
...
+  @ApiOperation(
+      value = "determine if a user has a specific global permission",
+      response = UserPermissionCheck.class,
+      produces = MediaType.APPLICATION_JSON
+  )
+  @ApiResponses(value = {
+      @ApiResponse(code = 404, message = "user id or permission not found"),
+      @ApiResponse(code = 500, message = "internal server error")
+  })
+  public Response getUserHasGlobalPermission(
```
drsnyder: Should this be restricted so that you can only see your permissions or is this internal only?

nmonterroso: This is probably a question for @jcellary. I didn't add authentication because none of the other resources had it.

service/user-permissions/src/main/java/com/wikia/services/userpermissions/service/MWUserPermissionsService.java
nmonterroso: No. Permissions are local to each wiki, but community has a user's global perms

drsnyder:
Minor comments and question about whether that or any resource in this service should be exposed. OTW, LGTM

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
jcellary: Yeah, initially I made everything public since MW also exposes all that information by the public api. However we then decided to change the course and mark these methods as requiring authorization. There's already a branch in which I'm working on this, so ideally you can add that directly here. My (unfinished) branch: https://github.com/Wikia/pandora/tree/user-permissions-auth

service/user-permissions/src/main/java/com/wikia/services/userpermissions/service/MWUserPermissionsService.java
jcellary: Yeah, as mentioned in jira ticket we'll probably extend MW to expore an api to only query the "global" table, we should then refactor this part

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
nmonterroso: Ah ok cool. Let's leave auth as part of that branch then and we'll add it to these resources when it's ready, to avoid conflicts

PR #565, opened 2015-10-28
Clearer error messages when linking accounts

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-958

Better error messages when trying to link an account while not logged in to Wikia, or when the Google account is already linked to a different Wikia account


drsnyder:
Looks like the tests are failing because of the MySQL username/password shuffle. Other than that, LGTM

PR #569, opened 2015-10-29
[SERVICES-956] Google access token validation

ArturKlajnerok:
To properly support case when google account rejects our app we need to validate also the access token. When app is removed form google account the google id token is still valid until expiration but the access token is invalidated and we should not authenticate users that already removed our app from their account. 

Now only `access token` will be used to google account authentication.

/cc @Wikia/services-team


service/external-auth/src/main/java/com/wikia/services/externalauth/gateway/GoogleGateway.java
```diff
@@ -24,43 +29,75 @@
...
-    GoogleIdToken idToken = null;
+  public Optional<AuthenticatedAccount> validateTokens(String idToken, String accessToken) {
+
+    GoogleIdToken googleIdToken = validateIdToken(idToken);
+
+    if (!validateAccessToken(accessToken)) {
+      return Optional.empty();
+    }
+
+    if (googleIdToken != null) {
```
nmonterroso: Is there a reason the `googleIdToken` would ever be passed in as `null`?

service/external-auth/src/main/java/com/wikia/services/externalauth/gateway/GoogleGateway.java
```diff
@@ -24,43 +29,75 @@
 
 public class GoogleGateway {
 
-  private static final String GOOGLE_API_UNAVAILABLE_MESSAGE = "Google API is unavailable.";
-
-  private static final String GOOGLE_TOKEN_SECURITY_MESSAGE = "Google token don't meet security rules.";
-
-  private static final String GOOGLE_TOKEN_VALIDATION_FAILED_MESSAGE = "Google token validation failed.";
+  private static final String GOOGLE_API_UNAVAILABLE_MSG = "Google API is unavailable.";
+  private static final String GOOGLE_TOKEN_SECURITY_MSG = "Google token don't meet security rules.";
```
drsnyder: Should this be `"Google token doesn't meet security rules."`?

drsnyder:
Something doesn't seem right about doing this. We are validating 2 tokens now because of a potentially unrealistic expectation or invalid assumption underlying the bug in SERVICES-956. This use case is not documented anywhere in the google docs regarding [backend auth](https://developers.google.com/identity/sign-in/web/backend-auth).

service/external-auth/src/main/java/com/wikia/services/externalauth/gateway/GoogleGateway.java
ArturKlajnerok: It's possible if there will be `GeneralSecurityException` or `Exception` cached in `private GoogleIdToken validateIdToken(String idToken)` then it will return `null` and that basically means that id token validation failed.

service/external-auth/src/main/java/com/wikia/services/externalauth/gateway/GoogleGateway.java
ArturKlajnerok: Fixed in : 327dedf

ArturKlajnerok:
Based on our offline discussion, we decided to use only `access_token` for google account authentication. I removed all the support for google id token.

nmonterroso:
LGTM

PR #571, opened 2015-10-29
Remove explicit, deprecated cors call from template service

armonr:
... In an effort to keep new services up to date.

/CC @nmonterroso @sqreek


nmonterroso:
LGTM

PR #573, opened 2015-10-30
terminating app when error occurs

rybmat:


workers/common/src/main/java/com/wikia/workers/common/AbstractWorker.java
```diff
@@ -82,11 +82,11 @@ public void run() {
         lastProcessed.update();
       } catch (Exception e) {
         LOGGER.error(e.getMessage(), e);
-        rabbitConnection.closeConnection();
-        System.exit(0);
+        break;
```
SebastianMarzjan: could stop() be called here instead to terminate the loop?

idradm:
how about introducing shutdown hooks for connections?

http://stackoverflow.com/questions/3715967/when-should-we-call-system-exit-in-java
http://docs.oracle.com/javase/1.5.0/docs/guide/lang/hook-design.html

workers/common/src/main/java/com/wikia/workers/common/AbstractWorker.java
```diff
@@ -65,6 +65,7 @@ public void stop() {
   }
 
   public void run() {
+    this.shutdownHooks();
```
idradm: Can we change this name to be more appropriate to what it does?

workers/common/src/main/java/com/wikia/workers/common/AbstractWorker.java
```diff
@@ -104,4 +104,15 @@ protected void route(QueueingConsumer.Delivery delivery) throws Exception {
...
     throw new NotImplementedException("this method should be implemented in subclass");
   }
+
+  /**
+   * when overriding, remember to call super
+   */
+  protected void registerShutdownHooks() {
+    Runtime.getRuntime().addShutdownHook(new Thread() {
+      public void run() {
+        rabbitConnection.closeConnection();
```
idradm: "this" ?

rybmat: can not use it

SebastianMarzjan:
Please resolve the conflicts; the current code looks good.

SebastianMarzjan:
Once checks pass LGTM

PR #581, opened 2015-11-02
Fixed user avatar swagger client

jcellary:
Added missing apiparam hidden=true attributes to the put method which made it impossible to generate the swagger client code


harnash:
LGTM

jcellary:
Ok, I'll keep an eye on occurrences of this in other services :)

PR #582, opened 2015-11-02
External-auth code clean up

sqreek:
@Wikia/services-team 
As in PR name.
I do not touched tests yet

Questions:
Is return BadRequest when tried linked google account is correct status code?
Can we merge Service into resource, to get rid off one layer of abstraction?
is function "remove" in GoogleAccountsDAO has correct logic? return always true, and delete whatever user provide ,so you can delete 2 rows, one with user_id, and second with google_id?

Also 
We should change DB_NAME to use config file instead generic jooq, for easier futher testing with different database names


harnash:
LGTM

ArturKlajnerok:
I don't think we should merge the service and the resource implementation, the resource is responsible for http rest functionality and should have as small logic as possible. As for BadRequest if there is something wrong with the incoming request I don't see anything wrong to return that status code.

sqreek:
In this moment bad request can occur only when you fail add new record into database. Maybe 409 Conflict? Bad Request seems strange to me, but I cannot figure out any other code.

service/external-auth/src/main/java/com/wikia/services/externalauth/dao/GoogleAccountsDAO.java
```diff
@@ -34,22 +24,12 @@ public GoogleAccountsDAO(JooqConfigs database) {
...
-      if (e.getCause().getClass().equals(MySQLIntegrityConstraintViolationException.class)) {
-        LOGGER.warn(LINKED_ACCOUNT_ALREADY_EXIST, e);
-        return false;
-      } else {
-        throw new InternalServerErrorException();
-      }
-    }
-
-    return true;
+        .onDuplicateKeyIgnore()
```
ArturKlajnerok: What status code will be returned when someone will try to insert data that already exist ??

ArturKlajnerok: Will this case also be logged, and will this be logged as error or warn ? I think that when someone tries add data that already exist we should log that only for information/debug purposes.

service/external-auth/src/main/java/com/wikia/services/externalauth/gateway/GoogleGateway.java
```diff
@@ -51,9 +51,9 @@ public GoogleGateway(@Named(GOOGLE_CLIENT_ID_KEY) String clientID) {
       LOGGER.error(GOOGLE_API_UNAVAILABLE_MSG, e);
       throw new InternalServerErrorException();
     } catch (Exception e) {
-      LOGGER.warn(GOOGLE_ACCESS_TOKEN_VALIDATION_FAILED_MSG, e);
+      LOGGER.error(GOOGLE_ACCESS_TOKEN_VALIDATION_FAILED_MSG, e);
```
ArturKlajnerok: I don't think this would always be an error, this will be cached for most cases when the token is just invalid.

service/external-auth/src/main/java/com/wikia/services/externalauth/dao/GoogleAccountsDAO.java
sqreek: Difference is this code will not Log warn. or whatever when someone will try break constraints.
It will return false, so Bad Request on top of a stack

service/external-auth/src/main/java/com/wikia/services/externalauth/gateway/GoogleGateway.java
sqreek: generic Exception is Exception U want to throw loudly.
You don't know what it is.
And if it will be a token is invalid, it should return specific Exception, not generic Exception.

nmonterroso:
^ :+1: @sqreek

service/external-auth/src/main/java/com/wikia/services/externalauth/gateway/GoogleGateway.java
nmonterroso: ^ :+1: @sqreek

PR #583, opened 2015-11-02
SOC-1555 Increment the number of discussion service instances

armonr:
/CC @jimmordino
https://wikia-inc.atlassian.net/browse/SOC-1555


service/discussion/build.gradle
```diff
@@ -50,7 +50,7 @@ deployments {
         resources {
             cpus = 1
             mem = 800
-            instances = 1
+            instances = 2
```
jimmordino: I would have gone for 3, but 2 is good as well.  You can temporarily scale to 2 and it should stick until this is deployed (unless some restarts).

jimmordino:
:shipit:

service/discussion/build.gradle
```diff
@@ -50,7 +50,7 @@ deployments {
         resources {
             cpus = 1
             mem = 800
```
jsutterfield: This is unrelated, but I noticed there's a disconnect btw the amount of memory we're asking for from Mesos here, and the amount of memory we're reserving for the JVM line 28 (`-Xmx300m`). So Mesos is reserving 800MB for each instance, but the JVM for each instance is capped at 300MB (you can verify by looking at `jvm.memory.heap.max` [here](http://mesos-s1:31874/metrics?pretty=true)).

We should probably update either one of those so they're consistent.

jsutterfield:
1 unrelated issue. Other than that, LGTM.

service/discussion/build.gradle
armonr: Good observation, @jsutterfield. I am not sure why this is; it could have originally been copied from the "example" service since a few other services look similar. I wont change that hear, but will follow up with the services team about this.

PR #589, opened 2015-11-03
[SERVICES-883] Removed application/v2+json header requirement

ArturKlajnerok:
/cc @Wikia/services-team


nmonterroso:
LGTM - let's make sure on the order of deployments between this, the wikia app, and helios to make sure we don't break anything :D

ArturKlajnerok:
That's my goal :smile_cat:  :+1:

PR #590, opened 2015-11-04
[SERVICES-883] Added support for old user preference put

ArturKlajnerok:
Added support for additional put resource for user preferences with `application/v2+json` header for the time of migration.

/cc @Wikia/services-team


nmonterroso:
LGTM

PR #593, opened 2015-11-04
CE-2187: Return data as an array

lukaszkonieczny:
Fix response type to return array while swagger-codegen generate incorrect code

@sqreek @harnash


harnash:
LGTM.
Did you tried leaving the 200 response definition without type specified?

lukaszkonieczny:
Yes, I tried. It doesn't work.

PR #595, opened 2015-11-04
SOC-1573 Update ja starwars forum name

armonr:
/CC @jimmordino


armonr:
Test is timing out, but this change wont affect any code.

jimmordino:
:+1:  This was best guess from google translate - hopefully the old values weren't offensive :-)

PR #596, opened 2015-11-04
Environment Configuration Overrides

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-933

Allow for environment variables to take priority over Consul variables. This is a step towards removing our reliance on Consul (at least for configuration, service-service TBD). Initially I wanted to move the `KeyValueConfig` interface, and the `CustomKeyValueConfig`, `ConsulKeyValueConfig` classes into the `dropwizard` project, but fwiw these configuration classes are being used somewhere outside of pandora (is this still correct @pchojnacki?)


pchojnacki:
@nmonterroso I had some plans to use those classes in deployment plugin but haven't got around to do that. so feel free to do whatever you want with it.

nmonterroso:
Looks like `:lib:consul-config:test` is failing because it detects a file difference with `master` (correctly), but the difference is a deletion, so the tests can't complete. We should ignore this failure.

nmonterroso:
Looks like `:lib:facade-gateway:test` is failing because http://pl.testmediawikigateway.wikia.com is closed

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/config/ServiceVariableLookup.java
```diff
@@ -10,15 +10,15 @@
 
 import javax.annotation.Nonnull;
 
-public class ConsulVariableLookup extends StrLookup {
+public class ServiceVariableLookup extends StrLookup {
 
   public static final Pattern VARIABLE_PATTERN = Pattern.compile("([\\w_]+)(:(.+))?");
-  public static final String SUBSTITUTOR_PREFIX = "${consul:";
+  public static final String SUBSTITUTOR_PREFIX = "${";
```
drsnyder: Should this constant be moved to `VariableInterpolationBundle`? I don't see it used in this class.

lib/service-config/src/main/java/com/wikia/config/env/CustomKeyValueConfig.java
```diff
@@ -0,0 +1,25 @@
+package com.wikia.config.env;
+
+import com.wikia.config.KeyValueConfig;
+
+import java.util.Map;
+import java.util.Optional;
+
+public class CustomKeyValueConfig implements KeyValueConfig {
```
drsnyder: Minor: should this be `KeyValueConfigMap`? `MapKeyValueConfig`? Something to reflect that it's a map.

drsnyder:
Minor comments. LGTM

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/config/ServiceVariableLookup.java
nmonterroso: Good catch 8150196

lib/service-config/src/main/java/com/wikia/config/env/CustomKeyValueConfig.java
nmonterroso: Yah I was trying to figure out what to call this thing. I wasn't happy with `CustomKeyValueConfig`. Changed to `MapKeyValueConfig` in 4f83e19

PR #597, opened 2015-11-05
SOC-1517 Notification service - notify resource

armonr:
* Add a new resource where notification requests can be accepted
* Introduce domain classes to model notifications
* Service specific exception mappers

/CC @garthwebb @jsutterfield 

https://wikia-inc.atlassian.net/browse/SOC-1517


service/notification/src/main/java/com/wikia/services/notification/NotificationApplication.java
```diff
@@ -3,15 +3,22 @@
...
 import io.dropwizard.Application;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class NotificationApplication extends Application<NotificationConfiguration> {
 
+  public static final Logger LOGGER =
+      LoggerFactory.getLogger(NotificationApplication.class); //NOSONAR
```
jsutterfield: Is there a reason to use this central logger instead of having each class define it's own?

service/notification/src/test/resources/fixtures/invalidUserNotification.json
```diff
@@ -0,0 +1,10 @@
+{
```
garthwebb: Is this used anywhere?

service/notification/src/main/java/com/wikia/services/notification/NotificationApplication.java
armonr: This is being used for exception classes, which probably having logger for could get a bit hairy.

service/notification/src/test/resources/fixtures/invalidUserNotification.json
armonr: No, I'll remove it.
This was going to be used for more tests for failure cases; I did not include the tests until I understand why exception mappers are not getting plugged in correctly in the tests.

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
```diff
@@ -8,42 +13,58 @@
...
+      @ApiResponse(code = 401, message = "Unauthorized", response = Problem.class),
+      @ApiResponse(code = 403, message = "Forbidden", response = Problem.class),
+      @ApiResponse(code = 500, message = "Failed to send notification", response = Problem.class)
   })
-  public Response getHelloWorld(
-      @ApiParam("name to say hello to") @PathParam("name") String name) {
+  public Response sendNotification(
+      @ApiParam(value = "Notification input json", required = true, name = "notificationInput")
+      @Valid NotificationInput notificationInput,
+      @Context UriInfo uriInfo) {
```
jsutterfield: Is this being used for anything?

service/notification/src/main/java/com/wikia/services/notification/domain/hal/input/NotificationInput.java
```diff
@@ -0,0 +1,109 @@
+package com.wikia.services.notification.domain.hal.input;
+
+import com.wikia.services.notification.domain.Notification;
+
+import com.wordnik.swagger.annotations.ApiModelProperty;
+
+import java.time.Instant;
+
+public class NotificationInput {
```
jsutterfield: I'm not sure I understand the purpose of this class. Why can't we just use `Notification.java` class by itself? Why do we have to wrap it?

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
```diff
@@ -8,42 +13,58 @@
...
+
+    Notification notification = notificationInput.getNotification();
+    Preconditions.checkArgument(notification.getTargetUser() > 0 &&
```
jsutterfield: These are the kind of things that Hibernate Validator can handle for us automatically.

If we end up just passing the `Notification` class in here directly, we can remove L56-63 here just by updating our annotations inside of Notification.java.
```java
// Notification.java

@Data
public class Notification {

@NotNull
@Size(min=1)
private Integer targetUser;

@NotNull
@Size(min=1, max=100. message="message must be non-empty and up to 100 characters")
private String message;
...
```
No need to set a max for targetUser, since Jackson will validate the type and if it's larger than an Integer will throw an error.

service/notification/src/main/java/com/wikia/services/notification/exceptions/IllegalArgumentExceptionMapper.java
```diff
@@ -0,0 +1,40 @@
...
+import java.net.URI;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+
+@Data
+public class IllegalArgumentExceptionMapper implements ExceptionMapper<IllegalArgumentException> {
```
jsutterfield: This is another one I don't really understand. I see you guys are using it in the Discussion service, but we never did anything like in the attribute service.

What's the idea behind these custom `ExceptoinMapper`s?

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
armonr: This is very useful to have in general for logging, errors, and debugging as well.

service/notification/src/main/java/com/wikia/services/notification/domain/hal/input/NotificationInput.java
armonr: This class adds an abstraction layer on the Notification class. This way, we can for instance hide internal fields that end user should not be concerned with.

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
armonr: I like the idea of having the validation on the class; the validation syntax is quite neat as well. However, it is much better to fail as soon as the input is known to be bad. This, IMO, allows for better use of resources and quicker responses.
I can add those attributes to the class as well though if that sounds good?

service/notification/src/main/java/com/wikia/services/notification/exceptions/IllegalArgumentExceptionMapper.java
armonr: Dropwizard by default returns 500 on a lot of different types of exceptions thrown while processing a request. This is achieved by a somewhat catch-all exception mapper whose job it is to essentially route an exception type (and its subtypes) to a handler to decides what to return as response. Mappers are a way to overwrite the default wiring and have custom handlers and therefore more sensible return codes and responses on certain types of exceptions.

jsutterfield: Gotcha. Thanks for the clarification!

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
armonr: @jsutterfield: I removed the input class to use dropwizard validation directly as discussed: [Here](https://github.com/Wikia/pandora/commit/a94cef6f7b9427bf1d7b3ed9af93c99b572b27e1). There is one big catch now: "[If the deserialized Notification isn’t valid, Dropwizard returns a 422 Unprocessable Entity response to the client.](https://dropwizard.github.io/dropwizard/0.6.2/manual/core.html#request-entities)". I'm not a fan of 422 and there seems to be no way to customize this, so I'm inclined to keep the input class because there is more control over response. Would you agree?

jsutterfield: Personally I think 422 is appropriate in that case. It looks like a 400 is used for 
> The request could not be understood by the server due to malformed syntax...

It's not that the request has malformed syntax, it's just missing required fields. A 422 on the other hand is used for:
>The server understands the content type of the request entity...and the syntax of the request entity is correct...but was unable to process the contained instructions.

I think in this case the "contained instructions" could be interpreted as "use these fields to perform some action". If there aren't enough fields, the instructions are essentially incomplete.

That being said, I'm not super wedded to 422. I'd be okay with updating it to a 400. I do think we should still avoid the `NotificationInput` class and let Dropwizard handle the validation for us. It looks we can customize this by intercepting [ConstraintViolationExceptionMapper](https://github.com/dropwizard/dropwizard/blob/release/0.8.x/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapper.java) responses like you're doing with `IllegalArgumentExceptionMapper`. In that mapper we can update the response code to a 400.

Also, since we're using Dropwizard 0.8, apparently we need to add `registerDefaultExceptionMappers: false` under `server` in our yaml file in order to use these custom exception mappers. See [here](http://stackoverflow.com/questions/28806959/override-dropwizard-constraintviolation-message/29059976#29059976)

armonr: Ah, this server configuration is exactly what was missing! Good find, @jsutterfield! Fix coming...

Also 422 is deprecated and a non-standard response. 400 should be more appropriate by consensus for instance, [here](http://stackoverflow.com/questions/16133923/400-vs-422-response-to-post-of-data).

service/notification/src/main/java/com/wikia/services/notification/domain/Notification.java
```diff
@@ -0,0 +1,62 @@
...
+import java.time.Instant;
+
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
+@EqualsAndHashCode(callSuper = false)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@AllArgsConstructor
+@NoArgsConstructor
```
jsutterfield: Do we need all of these annotations apart from `@Data`? It looks like `@Data` is a shorthand for `@ToString`, `@EqualsAndHashCode`, `@Getter` / `@Setter` and `@RequiredArgsConstructor`. I removed everything but `@Data` and it worked fine. Can we simplify this?

service/notification/src/main/java/com/wikia/services/notification/domain/Notification.java
```diff
@@ -0,0 +1,62 @@
...
+@Data
+@Builder
+public class Notification {
+
+  @JsonIgnoreProperties
+  private Long id;
+
+  @JsonProperty
```
jsutterfield: I've seen these annotations used around, but I haven't figured out what exactly they're used for. It seems like they're mostly used when we have a `Notification` object that we want to serialize to JSON. I removed them and Jackson still deserialized the request just fine. Do we need them?

service/notification/src/main/java/com/wikia/services/notification/exceptions/ConstraintViolationExceptionMapper.java
```diff
@@ -0,0 +1,42 @@
...
+import java.net.URI;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.validation.ConstraintViolationException;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+
+@Data
```
jsutterfield: Do we need to add this `@Data` annotation? It works fine without it. I think we just need to override the `toResponse` method, right?

service/notification/src/main/java/com/wikia/services/notification/exceptions/ConstraintViolationExceptionMapper.java
```diff
@@ -0,0 +1,42 @@
...
+public class ConstraintViolationExceptionMapper
+    implements ExceptionMapper<ConstraintViolationException> {
+
+  @Context
+  private HttpServletRequest request;
+
+  @Override
+  public Response toResponse(ConstraintViolationException exception) {
+    String errorMessage = String.format("400 Bad Request - %s", exception.getMessage());
```
jsutterfield: So an invalid response returned by this Mapper looks like this

```JSON
{
  "title": "400 Bad Request - The request entity had the following errors:",
  "instance": "/notify"
}
```
What it's missing is those "following errors". The [implementation from dropwizard](https://github.com/dropwizard/dropwizard/blob/release/0.8.x/dropwizard-jersey/src/main/java/io/dropwizard/jersey/validation/ConstraintViolationExceptionMapper.java) does this by passing `new ValidationErrorMessage(exception.getConstraintViolations());` to a Response object. This tells the client what is actually wrong, like 
```JSON
{
  "errors": [
    "targetUser may not be null (was null)"
  ]
}
```

We should probably return those specific errors as well.

service/notification/src/test/resources/notification-test.yaml
```diff
@@ -0,0 +1,47 @@
...
+        - jvm\.memory\.total\.committed
+server:
+  type: default
+  applicationConnectors:
+  - type: http
+    port: 8080
+  adminConnectors:
+  - type: http
+    port: 8081
+  registerDefaultExceptionMappers: false
```
jsutterfield: So it looks like there is a consequence to using this option. When this is set to false, jetty disables _all_ off the default mappers, not just the ones that we override.

What that means is that while we gain greater control over the responses we choose to override, we lose the handlers which already exist.

For example, here's what I get with `registerDefaultExceptionMappers` set to `true,`:
```bash
$ curl -X POST -H "Content-Type: application/json" -d '{"message": 1,  "targetUser": "foo" }' 'http://localhost:8080/notify' | python -m json.tool
{
  "code": 400,
  "message": "Unable to process JSON"
}
```

But if I issue that same request with `registerDefaultExceptionMappers` set to `false`, I get a generic 500 HTML error from dropwizard.

The reason is that jackson is throwing an `InvalidFormatException`, which we haven't created a mapper for. So the question is how we want to handle that. I think we should either create another mapper to handle  `InvalidFormatException` , or remove this option and just use the default mappers. I'm leaning towards the latter since we don't know what other exceptions could be thrown, but any that we miss seem to become generic 500s.

What do you think?

armonr: Several other services use this option. The benefit of giving the service more control over what it returns for different error scenarios far outweighs taking advantage of using out of the box responses just for the sake of using them.
Honestly, I am not too worried about all the edge-casey response codes. We can definitely add mappers to handle different errors, but since this is not open to the world, I think we can balance returning generic 500s for edge cases and have better control over responding properly to more common cases.

service/notification/src/main/java/com/wikia/services/notification/domain/Notification.java
armonr: Sure, it will work, but they cover properties/behaviors not so trivial. For instance, `EqualsAndHashCode` (hash and equals) is very important in the Java world when wanting to apply equality operation to the objects of a class, `JsonIgnoreProperties` helps with fields that should or should not be de/serialized, and so on.
So I would keep them :)

service/notification/src/main/java/com/wikia/services/notification/domain/Notification.java
armonr: [This](https://github.com/FasterXML/jackson-annotations/wiki/Jackson-Annotations#property-naming) should be a good documentation to look at. I personally like to use these and be consistent with the rest of services in pandora.

service/notification/src/main/java/com/wikia/services/notification/exceptions/ConstraintViolationExceptionMapper.java
armonr: This also generates setters/getters and a good toString() which is a very good practice in Java to have.

service/notification/src/main/java/com/wikia/services/notification/exceptions/ConstraintViolationExceptionMapper.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Context
+  private HttpServletRequest request;
+
+  @Override
+  public Response toResponse(ConstraintViolationException exception) {
+    String messages = exception.getConstraintViolations()
+        .stream()
+        .map(ConstraintViolation::getMessageTemplate)
+        .collect(Collectors.joining(", "));
```
jsutterfield: Cool use of Java's stream API :)

jsutterfield:
LGTM!

service/notification/src/main/java/com/wikia/services/notification/domain/Notification.java
```diff
@@ -0,0 +1,62 @@
...
+
+@EqualsAndHashCode(callSuper = false)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@AllArgsConstructor
+@NoArgsConstructor
+@Data
+@Builder
+public class Notification {
+
+  @JsonIgnoreProperties
```
garthwebb: From what I can tell `@JsonIgnoreProperties` is a class level annotation.  For member level annotations, you can use `@JsonIgnore`

service/notification/src/main/java/com/wikia/services/notification/domain/Notification.java
garthwebb: But it sounds like we can get rid of `EqualsAndHashCode` though right?  The `@Data` annotation covers that.

service/notification/src/main/java/com/wikia/services/notification/domain/Notification.java
```diff
@@ -0,0 +1,62 @@
...
+  @NotNull(message = "targetUser must not be null")
+  @Min(value = 1, message = "targetUser must be a valid positive integer")
+  private Integer targetUser;
+
+  @JsonProperty
+  @NotNull(message = "message must not be null")
+  @Size(min = 1, max = 100, message = "message must be non-empty and up to 100 characters")
+  private String message;
+
+  @JsonIgnoreProperties
```
garthwebb: Should be `@JsonIgnore` here and on L58

PR #599, opened 2015-11-05
User registration

nmonterroso:
@Wikia/services-team 

Adding user-registration service from `:tools:create-new-service:run`. 

Also adding `ConsulBundle` and the service's `Module` to the generated `Application`


service/user-registration/README.md
```diff
@@ -0,0 +1,25 @@
+# UserRegistration
+Please enter a short, general description about what this service does
```
ArturKlajnerok: Can you add some basic information to readme ?

ArturKlajnerok:
Other than one small comment, LGTM :+1:

service/user-registration/README.md
nmonterroso: d79831f

drsnyder:
:+1:

PR #606, opened 2015-11-09
Services 982 connect to flow

sqreek:
ping @ArturKlajnerok 
This one connect Storage into main flow

Tests was failing because there is no key values in consul yet


sqreek:
wrong merge branch choice

PR #611, opened 2015-11-09
[SERVICES-983] Dao for user registration

ArturKlajnerok:
Storage layer for user registration with the integration test.

/cc @Wikia/services-team


service/user-registration/src/main/java/com/wikia/services/userregistration/dao/StoredUsersDAO.java
```diff
@@ -0,0 +1,52 @@
...
+
+  private JooqConfigs database;
+
+  @Inject
+  public StoredUsersDAO(JooqConfigs database) {
+    this.database = database;
+  }
+
+  @Override
+  public Long add(User user) {
```
sqreek: after todays talk, we should take small steps into change userId from number into Strings

service/user-registration/src/test/java/com/wikia/services/userregistration/resources/UserRegistrationResourceTest.java
```diff
@@ -47,15 +51,30 @@ public void createValidUserShouldReturnOkWithToken(UserRegistrationResource reso
...
       UserRegistrationConfiguration configuration = new UserRegistrationConfiguration();
       configuration.setCookieDomain("test.something.com");
+      configuration.setHashPrefix(":B:");
+      configuration.setExpirationTime(3600);
+      configuration.setSaltSize(8);
       return configuration;
     }
   }
+
+  public static class TestBindingOverrideDefault extends AbstractModule {
```
nmonterroso: Why do we need this module?

service/user-registration/src/main/java/com/wikia/services/userregistration/dao/StoredUsersDAO.java
ArturKlajnerok: changed in: 957f80c

PR #614, opened 2015-11-10
Provide a service library for hashcash verification

drsnyder:
This PR adds the basics for hashcash verification of [hashcash version one](http://hashcash.org/docs/hashcash.html#stamp_format__version_1_).

https://wikia-inc.atlassian.net/browse/SERVICES-1007

TODO:
 * [x] Add tests for the double spend prevention branch

@Wikia/services-team


lib/hashcash/src/main/java/com/wikia/hashcash/HashcashVersion1.java
```diff
@@ -0,0 +1,43 @@
...
+
+    if (this.doubleSpendPrevention.hasBeenSpent(stampString)) {
+      return false;
+    }
+
+    return true;
+  }
+
+  private static Boolean partialPreimageMatch(String stampString, Integer bits) {
+    String zeros = StringUtils.repeat("0", bits);
```
ArturKlajnerok: Can we assign `0` to a constant ?

lib/hashcash/src/main/java/com/wikia/hashcash/StampGenerator.java
```diff
@@ -0,0 +1,27 @@
...
+import org.apache.commons.lang.StringUtils;
+
+import java.util.stream.IntStream;
+
+@Value
+@Builder
+public class StampGenerator {
+
+  public static String mint(String challenge, Integer bits) {
+    String zeros = StringUtils.repeat("0", bits);
```
ArturKlajnerok: `0` could be a constant.

lib/hashcash/src/main/java/com/wikia/hashcash/StampValue.java
```diff
@@ -0,0 +1,100 @@
...
+    if (parts.length != STAMP_ELEMENT_STRINGS.length) {
+      throw new IllegalArgumentException("Error, malformed stamp: " + stamp);
+    }
+
+    IntStream.range(0, STAMP_ELEMENT_STRINGS.length).forEach(
+        index -> stampValues.put(STAMP_ELEMENT_STRINGS[index], parts[index])
+    );
+
+    return StampValue.builder()
+        .version(stampValues.get("version"))
```
ArturKlajnerok: Could we use elements from `STAMP_ELEMENT_STRINGS` instead of strings when getting values ?

ArturKlajnerok:
Other than some small comments, LGTM :+1:

lib/hashcash/src/main/java/com/wikia/hashcash/StampValue.java
drsnyder: I was a little torn about this. Part of the reason that I converted it to a map is so I wouldn't have to `[0]`, `[1]`, etc. I always find that hard to read. If I then went and used the positions in `STAMP_ELEMENT_STRINGS` I would be back to what I was trying to avoid and could skip the array -> map translation altogether.

What I really wanted was something like

```
version, bits, date, resource, extension, rand, counter = stamp.split(":")
```

But you can't do that in :java:.

lib/hashcash/src/main/java/com/wikia/hashcash/StampGenerator.java
drsnyder: ab6d23e

PR #616, opened 2015-11-10
SERVICES-1008 add asynchronous calls to other services in UserRegistr…

sqreek:
This PR provide Asynchronous calls in UserRegistrationResource for postcreation commands.
Those calls are:
Call to helios for token
Call to SendConfirmationEmailService
Call to update antispoof
ping @Wikia/services-team


service/user-registration/src/main/java/com/wikia/services/userregistration/service/PhalanxService.java
```diff
@@ -4,7 +4,7 @@
 
 public class PhalanxService implements AntiSpoofService{
```
nmonterroso: I think these are two distinct things. Phalanx is the Scala app, but AntiSpoof happens inside MediaWiki.

nmonterroso:
Should updating `user-preferences` be part of this?

Also I wonder if we need to explicitly block user registration when updating anti spoof, user-preferences, or sending a confirmation email fails. To me, these are important parts of user-creation, but shouldn't block a user from registering and using Wikia.

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/UserRegistrationResource.java
```diff
@@ -59,9 +69,30 @@ public Response users(RegistrationData registrationData) {
...
 
-    return Response.ok(response).cookie(cookie).build();
+    try {
+      isAntiSpoofServiceUpdated.get();
+      isConfirmationEmailSend.get();
+      return Optional.of(token.get());
+    } catch (InterruptedException | ExecutionException e) {
+      LOGGER.error("Post creation commands occur exception", e);
+      return Optional.empty();
+    }
```
drsnyder: Is there any cleanup required of the thread pool?

ArturKlajnerok:
I think that for this use case we should use a solid async http client like: https://github.com/AsyncHttpClient/async-http-client instead of introducing concurrency logic in the registration resource.

PR #618, opened 2015-11-10
Support Multiple Gateway Dependencies

nmonterroso:
@Wikia/services-team 

Using `@Provides` in `ServiceProviderModule` did not allow for multiple installations of the `ServiceProviderModule`, which is what we want to do in the `user-registration` service since it has dependencies on multiple gateways. I chose to change this to use a `Provider` class rather than pushing the dependency on `ServiceGatewayModule` onto the applications, since that shouldn't be their concern.


ArturKlajnerok:
Ltgm :+1:

PR #619, opened 2015-11-11
[SERVICES-990] Set preferences for new user.

ArturKlajnerok:
Added `RegistrationService` that should combine all logic connected to registration process, also sets preferences for new user. Also did additional small refactor.

/cc @Wikia/services-team


service/user-registration/build.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
     compile project(':lib:service-gateway:user-preference-gateway')
     compile project(':lib:dropwizard:databases:db-jooq')
     compile project(":lib:jooq:jooq-wikicities")
+    compile project(':lib:service-gateway:user-preference-gateway')
```
nmonterroso: This is already added above, so we can remove this redundant line :)

service/user-registration/src/main/java/com/wikia/services/userregistration/guice/UserRegistrationModule.java
```diff
@@ -27,9 +27,9 @@ protected void configure() {
     bind(ConfirmationEmailService.class).to(AppEmailService.class);
     bind(AntiSpoofService.class).to(PhalanxService.class);
     bind(TokenService.class).to(HeliosService.class);
-    bind(UsersStorageService.class).to(MwUsersStorageService.class);
     bind(StoredUsers.class).to(StoredUsersDAO.class);
     bind(PasswordHashService.class).to(MwPasswordHashService.class);
+    bind(UserPreferenceGateway.class).to(UserPreferenceGatewayImpl.class);
```
nmonterroso: I think the service doesn't need to be concerned with doing this explicit binding, since the implementation is hidden behind the `UserPreferenceGateway` interface that this specific application shouldn't have to know. 

Instead we should add an instance of `UserPreferenceGatewayModule` to the `Application` class's list of installed modules. There's another dependency (`ServiceGatewayModule`) that `UserPreferenceGatewayModule` installs that's required for the service resolution to work.

service/user-registration/src/main/java/com/wikia/services/userregistration/service/RegistrationService.java
```diff
@@ -0,0 +1,114 @@
...
+import javax.inject.Inject;
+import javax.ws.rs.InternalServerErrorException;
+
+public class RegistrationService {
+
+  private static final String PREFERENCES_API_BAD_REQUEST_MSG = "Bad request to user preferences API.";
+  private static final String PREFERENCES_API_UNAVAILABLE_MSG = "User preferences API is unavailable.";
+  private static final Logger LOGGER = LoggerFactory.getLogger(RegistrationService.class);
+
+  private static final String LANGUAGE_CODE_DEFAULT = "en";
```
nmonterroso: Hm, a lot of these constants exist in `RegistrationData` and `Fields` right now. We should consolidate on where we reference them from.

service/user-registration/src/main/java/com/wikia/services/userregistration/service/RegistrationService.java
```diff
@@ -0,0 +1,114 @@
...
+      LOGGER.error(PREFERENCES_API_BAD_REQUEST_MSG, e);
+      throw new InternalServerErrorException();
+    } catch (GatewayException e) {
+      LOGGER.error(PREFERENCES_API_UNAVAILABLE_MSG, e);
+      throw new InternalServerErrorException();
+    }
+  }
+
+  private String getLanguageCodeFromData(RegistrationData registrationData) {
+    if (registrationData.getLangCode().isEmpty()) {
```
nmonterroso: It should be impossible for the language code to be empty. It's validated by `@NotEmpty` and it has a default of `en`

service/user-registration/src/main/java/com/wikia/services/userregistration/service/RegistrationService.java
```diff
@@ -0,0 +1,114 @@
...
+      return LANGUAGE_CODE_DEFAULT;
+    }
+    return registrationData.getLangCode();
+  }
+
+  private String getMarketingAllowedFromData(RegistrationData registrationData) {
```
nmonterroso: the bean validation takes care of ensuring that this is either `"1"` or `"0"`, so we shouldn't have to do any additional validation here.

service/user-registration/src/test/java/com/wikia/services/userregistration/helpers/TestHelper.java
```diff
@@ -5,23 +5,41 @@
 
 import org.joda.time.DateTime;
 import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
 
 import java.sql.Date;
 
 public class TestHelper {
 
+  public static final DateTimeFormatter FMT = DateTimeFormat.forPattern("yyyyMMddHHmmss");
```
nmonterroso: Since this is repeated in `RegistrationService. getUserFromRegistrationData`, should this be a constant in that class?

nmonterroso:
some comments otw lgtm

service/user-registration/src/test/java/com/wikia/services/userregistration/helpers/TestHelper.java
ArturKlajnerok: Moved in : 1c86167

service/user-registration/src/main/java/com/wikia/services/userregistration/service/RegistrationService.java
ArturKlajnerok: Removed : 452196b

service/user-registration/src/main/java/com/wikia/services/userregistration/service/RegistrationService.java
ArturKlajnerok: Removed : 452196b

service/user-registration/build.gradle
ArturKlajnerok: Removed : 452196b

service/user-registration/src/main/java/com/wikia/services/userregistration/service/RegistrationService.java
ArturKlajnerok: Removed : c336ee2

service/user-registration/src/main/java/com/wikia/services/userregistration/guice/UserRegistrationModule.java
ArturKlajnerok: Removed: e0074f1

service/user-registration/src/main/java/com/wikia/services/userregistration/service/RegistrationService.java
```diff
@@ -0,0 +1,95 @@
...
+    return User.builder()
+        .name(registrationData.getUsername())
+        .birthDate(new Date(registrationData.getBirthdate().getMillis()))
+        .email(registrationData.getEmail())
+        .passwordHash(hashService.hash(registrationData.getPassword()))
+        .registrationTime(DATE_TIME_FMT.print(DateTime.now()))
+        .build();
+  }
+
+  private void setUserPreferences(String userId, RegistrationData registrationData) {
```
sqreek: Is this should be separated and async?
ping @Wikia/services-team

nmonterroso: I don't think we should manage our own concurrency here. If we want to do this concurrently, it should be part of the generated client. We have a ticket to do that here: https://wikia-inc.atlassian.net/browse/SERVICES-920

sqreek: ok, but I think logic about user-preference request should be separated from logic of create new user in database

PR #620, opened 2015-11-12
UserRegistrationData validation

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-989

Provide a way to validate registration data input from the user using the [Hibernate Validator](http://hibernate.org/validator/). Most constraints are out-of-the-box, but I had to introduce constraints for being less than 13 years old, having the username be equal to the password, and having a string not match a pattern (for username matching an IP address or having invalid characters)


nmonterroso:
Tests for `:service:user-registration` pass, but more projects will be tested since the check is done against `master`. since this merge is only to the `SERVICES-973_User_Registration_Service` branch we shouldn't block merging if other tests fail.

service/user-registration/src/main/java/com/wikia/services/userregistration/domain/RegistrationData.java
```diff
@@ -1,26 +1,108 @@
...
-  private String langCode;
+  @Length(max= MAX_LANGUAGE_CODE_LENGTH, message = LANGUAGE_CODE_TOO_LONG)
+  @JsonProperty(value = Fields.LANGUAGE_CODE, defaultValue = Fields.LANGUAGE_CODE_DEFAULT)
+  private String langCode = Fields.LANGUAGE_CODE_DEFAULT;
 
-  @JsonProperty("marketingallowed")
-  private String marketingAllowed;
+  @OneOf(value = {"0", "1"}, message = MARKETING_ALLOWED_INVALID)
+  @JsonProperty(value = Fields.MARKETING_ALLOWED, defaultValue = Fields.MARKETING_ALLOWED_DEFAULT)
+  private String marketingAllowed = Fields.MARKETING_ALLOWED_DEFAULT;
```
drsnyder: :+1: declarative

nmonterroso:
Tests have timed out ~5 times in a row now. They pass locally.

service/user-registration/src/main/resources/banner.txt
```diff
@@ -1,8 +1,19 @@
...
+        ( //) |        \///      |     \     \
+      ( / /) _|_ /   )  //       |      \     _\
+    ( // /) '/,_ _ _/  ( ; -.    |    _ _\.-~        .-~~~^-.
+  (( / / )) ,-{        _      `-.|.-~-.           .~         `.
+ (( // / ))  '/\      /                 ~-. _ .-~      .-~^-.  \
+ (( /// ))      `.   {            }                   /      \  \
+  (( / ))     .----~-.\        \-'                 .~         \  `. \^-.
+             ///.----..>        \             _ -~             `.  ^-`  ^-_
+               ///-._ _ _ _ _ _ _}^ - - - - ~                     ~-- ,.-~
+                                                                  /.-~
```
drsnyder: WHOA!

drsnyder:
LGTM
It's nice to be able to do this declaratively. Even though the parameter list is so few it's still nice to see the constraints in one place.

service/user-registration/src/main/resources/banner.txt
nmonterroso: YES WE ARE IN THE FUTURE

PR #622, opened 2015-11-13
SOC-1518 Use Parse to send push notifications

jsutterfield:
This PR creates a Parse client which is used to send push notifications. The client is injected into the `NotificationResource`, and used by the `POST` endpoint.

The [updated README](https://github.com/Wikia/pandora/compare/SOC-1518?expand=1&name=SOC-1518&short_path=01dc88c#diff-01dc88ccd87414c197fd5a7b7a601ff9) gives a more in-depth explanation about what's needed to send these notifications via Parse and how this PR implements those requirements.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-1518


service/notification/src/main/java/com/wikia/services/notification/domain/ParseKey.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class ParseKey {
+
+  public static final String APP_ID_HEADER = "X-Parse-Application-Id";
+  public static final String API_KEY_HEADER = "X-Parse-REST-API-Key";
+
+  public final String apiKey;
+  public final String appId;
+
+  public ParseKey(String apiKey, String appId) {
```
jsutterfield: I made these resemble C-style structs rather than JavaBean because of [this recommendation](https://github.com/cxxr/better-java#structs_

service/notification/src/test/java/com/wikia/services/notification/resources/NotificationResourceTest.java
```diff
@@ -1,78 +1,105 @@
...
+    Response response = RULE.client().target(NOTIFY_PATH).request().post(
+        Entity.entity(fixture(FIXTURES_INVALID_SOURCE_ID), MediaType.APPLICATION_JSON_TYPE)
+    );
+
+    assertThat(response.getStatusInfo().getStatusCode())
+        .isEqualTo(Response.Status.BAD_REQUEST.getStatusCode());
+  }
 
   @Test
-  public void notifySuccess() {
```
jsutterfield: Since these original `notifySuccess` and `notifyValidationFail` are integration tests, I moved them into `NotificationResourceIntegrationTest`.

service/notification/src/test/resources/fixtures/trivialNotification.json
```diff
@@ -1,9 +1,9 @@
 {
-  "sourceId": "0",
+  "sourceId": 0,
   "sourceType": "",
-  "objectId": "0",
+  "objectId": 0,
   "objectType": "",
   "action": "",
-  "targetUser": "1",
+  "targetUser": 27204647,
```
jsutterfield: This is the test user I created: NotificationServiceTester.

armonr: Let's try to use strings and not numbers here: https://one.wikia-inc.com/wiki/User_blog:Charles_Miller/Don%27t_use_numbers_as_IDs_in_your_API

service/notification/src/main/resources/notification.yaml
```diff
@@ -45,3 +45,15 @@ swaggerConfiguration:
...
+  connectionTimeout: 2s
+  timeToLive: 1h
+  cookiesEnabled: false
+  maxConnections: 1024
+  maxConnectionsPerRoute: 1024
+  keepAlive: 0ms
+  retries: 0
+  userAgent: notification (dropwizard)
+parseApiUrl: ${PARSE_API_URL}
+parseKeys: ${PARSE_KEYS}
```
armonr: Missing default values

service/notification/src/main/java/com/wikia/services/notification/client/ParseClient.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  public boolean sendPush(Notification notification, ParseKey parseKey) throws IOException {
+    boolean success = true;
+    HttpResponse response = httpClient.execute(getPreparedPost(notification, parseKey));
+
+    int statusCode = response.getStatusLine().getStatusCode();
+    if (statusCode != Response.Status.OK.getStatusCode()) {
+      success = false;
+      LOGGER.warn(String.format(FAILED_PUSH_MESSAGE, notification.getMessage(),
```
armonr: Is this just a warning or an error if a notification fails?

service/notification/src/main/java/com/wikia/services/notification/client/ParseClient.java
```diff
@@ -0,0 +1,67 @@
...
+import javax.ws.rs.core.Response;
+
+public class ParseClient {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ParseClient.class);
+
+  private final static String FAILED_PUSH_MESSAGE =
+      "Failed to send push notification. Message: \"%s\", user_id: %d, ParseResponse: \"%s\"";
+  private final static String PAYLOAD_TEMPLATE =
+      "{\"where\": {\"user_id\": \"%d\"}, \"data\": {\"alert\": \"%s\"}}";
```
armonr: This is simple enough right now, but later should probably be a json template.

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
```diff
@@ -47,11 +60,19 @@
   public Response sendNotification(
       @ApiParam(value = "Notification json", required = true, name = "notification")
       @Valid Notification notification,
-      @Context UriInfo uriInfo) {
+      @Context UriInfo uriInfo) throws IOException {
+
+    Optional<ParseKey> parseKey = parseKeyService.get(notification.getSourceId());
+
+    if (!parseKey.isPresent()) {
```
armonr: It's better to use map/OrElse

service/notification/src/test/java/com/wikia/services/notification/client/ParseClientTest.java
```diff
@@ -0,0 +1,74 @@
...
+  private static final HttpEntity ENTITY = mock(HttpEntity.class);
+  private static final String PARSE_API_URL = "http://localhost";
+  private static final ParseClient PARSE_CLIENT = new ParseClient(HTTP_CLIENT, PARSE_API_URL);
+
+  private static final String NOTIFICATION_FIXTURE = "fixtures/trivialNotification.json";
+  private static final ParseKey PARSE_KEY = new ParseKey("AppId", "dApiKey");
+
+  private Notification notification;
+
+  private Notification getNotificationInputFromFixture()
```
armonr: This can get the fixture name as a parameter and each test can call it separately rather than done in setup. If a test needs to use a different fixture, this cannot be used.

service/notification/src/test/resources/fixtures/invalidLongMessageNotification.json
```diff
@@ -0,0 +1,10 @@
+{
+  "id": 0,
+  "sourceId": 0,
+  "sourceType": "",
+  "objectId": 0,
+  "objectType": "",
+  "action": "",
+  "targetUser": 0,
+  "message": "This message is too long. The limit is 100 characters. This one is 165 characters. Clearly this is a rambly message, which wouldn't work very well as a notification."
```
armonr: @garthwebb: Is 100 a good limit? Originally, I picked this number arbitrarily :game_die:

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
garthwebb: In this case though, wouldn't it be better to *not* use `Optional` since parseKey doesn't seem to be optional (e.g. we return an error if its not present)?  Seems like we should just do:
```java
ParseKey parseKey = parseKeyService.get(notification.getSourceId());
if (parseKey == null) {
...
```

service/notification/README.md
```diff
@@ -1,14 +1,65 @@
 # Notification
-The notification service will act as a gateway for internal services that want to send push notifications or emails. The first stage will focus specifically on push notifications. For more information, see the [planning docs](https://one.wikia-inc.com/wiki/Engineering/Services/Notification/Planning) 
+The notification service will act as a gateway for internal services that want to send push
+notifications or emails. The first stage will focus specifically on push notifications.
+For more information, see the [planning docs](https://one.wikia-inc.com/wiki/Engineering/Services/Notification/Planning) 
 
 # API Docs
-Please add information about resources and what they provide
+The best way to explore the API is to visit the swagger interactive documentation. To access that 
+documentation, first start the server (see "To run the server" below), then browse to [localhost:8080/swagger]
```
garthwebb: I think this link needs a URL portion with a protocol to be linked, e.g., 
```
[localhost:8080/swagger](http://localhost:8080/swagger)
```

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
armonr: [`Optional`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html)'s job is just that: to eliminate the need to explicitly compare things to null and at the same time provides a "functional" interface which is nice to use.

service/notification/src/test/resources/fixtures/invalidLongMessageNotification.json
garthwebb: Maybe up it a bit.  I can see on my iPhone that the notification center seems to show up to about 150 characters before it shows ellipsis. Additionally, the internets say (http://stackoverflow.com/questions/9487223/what-is-the-maximum-number-of-characters-that-can-be-displayed-in-android-push-n) that iPhone has a limit of about 2k bytes and Andriod is 4k bytes.

Also, not sure if your check is looking at bytes or characters, but should make sure to look at characters.  Given the above I'd say maybe 250 characters (up to 500 bytes for wide characters)

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
garthwebb: I'm not sure about that.  `Optional` is to help manage values that aren't required for the code to continue; it allows chaining together data processing commands or allows more readable ways to say what should happen if there happens to be a value.

Here not having `parseKey` is an error condition; there's no advantage to wrapping it in an optional just to switch a `== null` for `isPresesnt()`.  Where as it eliminates boilerplate code for optional values, it adds boiler plate for values that are required.

Note what the Oracle docs say in the conclusion on their article about `Optional`:
```
The purpose of Optional is not to replace every single null reference in your codebase but rather to help design better APIs in which—just by reading the signature of a method—users can tell whether to expect an optional value.
```

service/notification/src/main/java/com/wikia/services/notification/client/ParseClient.java
```diff
@@ -0,0 +1,67 @@
...
+    this.httpClient = httpClient;
+    this.parseApiUrl = parseApiUrl;
+  }
+
+  public boolean sendPush(Notification notification, ParseKey parseKey) throws IOException {
+    boolean success = true;
+    HttpResponse response = httpClient.execute(getPreparedPost(notification, parseKey));
+
+    int statusCode = response.getStatusLine().getStatusCode();
+    if (statusCode != Response.Status.OK.getStatusCode()) {
```
garthwebb: Would it be possible to simplify this to:
```java
if (response.getStatusLine() != Response.Status.OK) {
    ...
}
```
It doesn't look like they're the same type of thing though... I'll just note that it makes me super sad for some reason to see all of these superfluous `getStatusCode()` calls that don't really seem to add much value.

service/notification/src/test/resources/fixtures/invalidLongMessageNotification.json
armonr: Cool. And [`@Size#max`](http://docs.oracle.com/javaee/6/api/javax/validation/constraints/Size.html#max()) should be by character count.

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
armonr: Sure, it would not be of much use to introduce an Optional in the same scope that it gets inspected, for example, but it still adds value if the variable in question is getting passed around, IMO. Having said that, feel free to remove it here.

service/notification/src/main/java/com/wikia/services/notification/client/ParseClient.java
garthwebb: I think this is fine as a warning until this service matures.  At some point it should be the Notification services' job to re-queue this notification if it fails but right now the infrastructure doesn't exist to handle this.  Once we get some kind of queue up in here, we can add retry logic.

In any case however, its not the client's problem (the discussion service in this case) if something goes wrong. It simply needs to emit events and let them find their way to the user.  We should only really return Server errors from the notification service itself, not one of its dependent services.

armonr: Agreed. Just to clarify, I meant to suggest using `LOGGER.error()` instead.

service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
garthwebb: Actually I think there are great uses of `Optional` in the same scope that you've already shown in other code, e.g. `.map().orElse()` for getting default values.  I just wanted to point out that declaring something Optional that's actually required is not very semantic, and doesn't give you the benefits that you'd get if it were really optional. Anyway time to stop beating a dead :horse:

service/notification/src/main/java/com/wikia/services/notification/client/ParseClient.java
garthwebb: Sounds good.  Only other thing to note is that changing this to `error()` fires up the P3 generator if we get here.  Not saying that's good or bad, but something to consider :)

armonr: I was not aware of P3 generation. Sure, warning it is, until failure tolerance is lowered to the P3 level :smile_cat:

garthwebb:
Other than a change to the Markdown and a few notes, LGTM

service/notification/src/main/resources/notification.yaml
jsutterfield: I don't think these actually need default values. They're required fields. The service shouldn't run without them.

service/notification/src/test/resources/fixtures/trivialNotification.json
jsutterfield: :+1:

armonr:
LGTM!

PR #623, opened 2015-11-13
SOC-1506 Block check from permissions to before permission and roles are added in post response

armonr:
This is a temporary solution; as permission service evolves, this check will not remain clunky and inefficient.
* Adds functionality in the service to be able to check block status
* Utilizes the function in one place to limit permission emission in the response to the unblocked
* More places will need to use this soon...

/CC @garthwebb  @jsutterfield 

https://wikia-inc.atlassian.net/browse/SOC-1506


service/discussion/src/main/java/com/wikia/discussionservice/services/UserPermissionServiceImpl.java
```diff
@@ -49,6 +54,37 @@ public UserPermissionServiceImpl(
...
+      httpGet.addHeader(GATEWAY_USER_ID_HEADER, userId);
+      httpClient.execute(httpGet);
+      response = httpClient.execute(httpGet);
+
+      int statusCode = response.getStatusLine().getStatusCode();
+      if (statusCode == Response.Status.NOT_FOUND.getStatusCode()) {
+        return Boolean.TRUE;
+      } else if (statusCode == Response.Status.OK.getStatusCode()) {
+        Optional<ClientUserPermission> permissionOptional = createPermissionFromResponse(response);
+        return permissionOptional.map(ClientUserPermission::getUserIsBlocked)
```
garthwebb: Where is `getUserIsBlocked` defined?

service/discussion/src/main/java/com/wikia/discussionservice/services/UserPermissionServiceImpl.java
```diff
@@ -94,10 +130,31 @@ private String getPermissionURL(String userId, String wikiId) {
...
+   * @param response CloseableHttpClient
+   * @return Optional <ClientUserPermission>
+   * @throws IOException
+   */
+  private Optional<ClientUserPermission> createPermissionFromResponse(
+      CloseableHttpResponse response)
+      throws IOException {
+    String responseContent = EntityUtils.toString(response.getEntity());
+    ObjectMapper mapper = new ObjectMapper();
+    return Optional.of(mapper.readValue(responseContent, ClientUserPermission.class));
```
jsutterfield: Does this have to return an `Optional`? We can either successful deserialize a `CilentUserPermission` object, or this throws an IOException, right? Is it possible for this method to return an empty `Optional`?

service/discussion/src/main/java/com/wikia/discussionservice/services/UserPermissionServiceImpl.java
armonr: Lambok's [`@Data`](https://projectlombok.org/features/Data.html) takes care of adding boilerplate getters and setters.

service/discussion/src/main/java/com/wikia/discussionservice/services/UserPermissionServiceImpl.java
armonr: It is always possible there maybe empty or missing fields as far as this code is concerned. I don't think that quite fits in `IOException` more than an invalid API response.

armonr:
Closing. Need to retarget...

armonr:
#645

PR #626, opened 2015-11-13
Phalanx gateway and validation

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-994

Adding `BlockedContentGateway` which uses `Phalanx` to determine if certain content should be blocked. Added an annotation constraint and added it to RegistrationData so that it completes along with other validation logic. 

Also piggybacking in changes that centralize logic for generating a swagger client from a json file, since `blocked-content-gateway` would have been the 3rd gateway that does this. This is now in `gradle/swagger/generate_from_json.gradle`. Also moved `generate_spec_and_code.gradle` to `generate_from_project.gradle` to be consistent.

Also adding the ability to have validators used with `Hibernate` to receive injection, since we needed this to be able to have `@NotBlocked` use the expected `BlockedContentGateway`


nmonterroso:
The test builds keep timing out. The tests pass locally.

ArturKlajnerok:
Lgtm :+1:

PR #627, opened 2015-11-14
Proof of work service

drsnyder:
This PR provides a proof of work service. The proof of work service provides two endpoints `/challenge` and `/verify`. The `/challenge` endpoint will provide a map with elements `"bits"` and `"challenge"` upon posting a string representing the resource to it. The `/verify` service responds with a 204 when a valid and verified hashcash stamp is `POST`ed to it.

The service also provides `/test/challenge` for testing purposes. You can enter a resource and see the request response flow.

https://wikia-inc.atlassian.net/browse/SERVICES-1003
https://wikia-inc.atlassian.net/browse/SERVICES-1004 (the javascript "proof" function)

@Wikia/services-team


lib/dropwizard/base/build.gradle
```diff
@@ -7,7 +7,10 @@ dependencies {
     compile 'com.netflix.governator:governator:1.3.3'
 
     compile "io.dropwizard:dropwizard-core:$dropwizardVersion"
-    compile "io.dropwizard:dropwizard-client:$dropwizardVersion"
+    compile("io.dropwizard:dropwizard-client:$dropwizardVersion") {
+      exclude group: "commons-codec"
+    }
+    compile "commons-codec:commons-codec:1.10"
```
drsnyder: Is this the right place or `services-core` below?

service/proof-of-work/src/main/java/com/wikia/services/proofofwork/domain/Challenge.java
```diff
@@ -0,0 +1,84 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import lombok.Value;
+import net.logstash.logback.encoder.org.apache.commons.lang.RandomStringUtils;
+
+import java.time.Clock;
+import java.time.LocalDateTime;
+
+@Value
+@JsonDeserialize(builder=Challenge.Builder.class)
```
drsnyder: I did this and the `Builder` below so that it can be deserialized in the tests. Jackson has trouble with deserializing `@Value` objects without the constructor or annotation hints.

service/proof-of-work/src/main/java/com/wikia/services/proofofwork/domain/ChallengeSignature.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public static Boolean verifySignature(String stamp, String key) {
+    return verifySignature(StampValue.fromString(stamp), key);
+  }
+
+  public static Boolean verifySignature(StampValue stamp, String key) {
+    ChallengeSignature signatureFromStamp = ChallengeSignature.builder()
+        .date(stamp.getDate())
+        .random(stamp.getRand())
+        .build();
```
drsnyder: Using an object for this originally felt like overkill but then I felt like generating for and extracting from a stamp would always be consistent and less error prone.

lib/dropwizard/base/build.gradle
ArturKlajnerok: Do we need to add that dependency to `dropwizard /base` isn't it specific only for `hashcash` service ??

lib/hashcash/src/main/java/com/wikia/hashcash/HashcashVersion1.java
```diff
@@ -17,26 +17,18 @@ public HashcashVersion1(DoubleSpendPrevention doubleSpendPrevention) {
...
+    return stamp.validateFormat() &&
+           stamp.getVersion().equals(STAMP_VERSION) &&
+           partialPreimageMatch(stampString, stamp.getBits()) &&
+           stamp.isDateWithinGracePeriod(gracePeriodInSeconds) &&
+           !this.doubleSpendPrevention.hasBeenSpent(stampString);
   }
 
   private static Boolean partialPreimageMatch(String stampString, Integer bits) {
-    String zeros = StringUtils.repeat(StampValue.ZERO_STRING, bits);
+    String zeros = StringUtils.repeat(StampValue.ZERO_STRING, (int)Math.floor(bits / 4));
```
ArturKlajnerok: Why do we need to `Math.floor()` here ?

lib/dropwizard/base/build.gradle
drsnyder: It's included in `dropwizard-client` (for what I don't know). However, the version included with `dropwizard-client` is 1.6 and I need something greater than or equal to 1.7. I thought it's best to first exclude it from `dropwizard-client` then re-add the replacement immediately following. That way it should be somewhat more clear what's happening.

lib/hashcash/src/test/java/com/wikia/hashcash/HashcashVersion1Test.java
```diff
@@ -15,20 +15,22 @@
 @RunWith(JUnitParamsRunner.class)
 public class HashcashVersion1Test {
 
+  private static Integer BITS = 4;
+  private static Long gracePeriod = 600L;
```
ArturKlajnerok: Const uppercasse notation

ArturKlajnerok: Can we also test the use-case when solution was delivered after the `gracePeriod` ??

lib/hashcash/src/main/java/com/wikia/hashcash/HashcashVersion1.java
drsnyder: This is being slightly more liberal in what we accept. The `bits` notion in hashcash is a little misleading. It really only steps in units of 4 (4, 8, 12, etc) because each hex character represents 4 bits and the comparison is done on the hex output of sha1.

That way, if going out you said this was 11 bits but the client rounded down (which it probably should) then you would accept it because you would only look for 2 leading zeros instead of 3. I imagine bits was used to allow for more granularity in the future.

lib/hashcash/src/test/java/com/wikia/hashcash/HashcashVersion1Test.java
drsnyder: The bounds checking tests are done [here](https://github.com/Wikia/pandora/blob/SERVICES-1003/lib/hashcash/src/test/java/com/wikia/hashcash/StampValueTest.java#L53). Is that ok?

drsnyder: The case change was fixed in c707f72.

service/proof-of-work/src/main/java/com/wikia/services/proofofwork/ProofOfWorkConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+public class ProofOfWorkConfiguration
+    extends Configuration
+    implements SwaggerConfigurationProvider {
+
+  @JsonProperty
+  private SwaggerBundleConfiguration swaggerConfiguration;
+
+  @NotNull
+  @Valid
+  @JsonProperty("bits")
```
ArturKlajnerok: I think that `"bits"` in annotation is not needed as far as it's the same with variable name.

lib/hashcash/src/main/java/com/wikia/hashcash/HashcashVersion1.java
ArturKlajnerok: Got it :+1:

ArturKlajnerok:
Other then one small comment at the end, LGTM :+1:

drsnyder:
Waiting on the tests to clear up....

PR #629, opened 2015-11-16
Registration validation exception

nmonterroso:
@Wikia/services-team 

Adding exception and exception mapper for registration data issues.


lib/service-gateway/blocked-content-gateway/src/test/java/com/wikia/services/blockedcontent/constraint/NotBlockedValidatorTest.java
```diff
@@ -51,11 +51,11 @@ public void nullOrEmptyContentShouldValidate() {
   }
 
   @Test
-  public void gatewayExceptionShouldFail() throws Exception {
+  public void gatewayExceptionShouldValidate() throws Exception {
```
ArturKlajnerok: I'm not sure that if the request to phalanx fails we should assume that content is valid.

ArturKlajnerok:
Other than one small comment, Lgtm :+1:

lib/service-gateway/blocked-content-gateway/src/test/java/com/wikia/services/blockedcontent/constraint/NotBlockedValidatorTest.java
nmonterroso: I committed this when registration requests were failing because the phalanx service couldn't be resolved. It made sense to me to make this the default, but I don't have a strong opinion either way. I'll revert this.

nmonterroso: 6431fca

PR #630, opened 2015-11-17
Google Registration

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-985
https://wikia-inc.atlassian.net/browse/SERVICES-988

Creating `GoogleRegistrationService` and `GoogleRegistrationResource`, which are responsible for creating new accounts when using a Google `access_token`. 

Also fixing some bugs with authenticating service calls made through the gateways.

- [x] tests


lib/service-gateway/user-preference-gateway/src/main/java/com/wikia/userpreference/gateway/UserPreferenceGatewayImpl.java
```diff
@@ -32,6 +33,15 @@ public UserPreferencesApi userPreferencesApi() throws GatewayException {
   }
 
   @Override
+  public UserPreferencesApi userPreferencesApi(String userId) throws GatewayException {
+    ApiClient apiClient = getApiClient();
+    apiClient.addDefaultHeader("X-Wikia-UserId", userId);
+    ((ApiKeyAuth) apiClient.getAuthentication("user_id")).setApiKey(userId);
```
nmonterroso: This is pretty ghetto, but the Swagger code generation will only generate code for a single `ApiKey` (whichever it finds first). Since we have two authentication mechanisms depending on whether or not you come through the gateway, we need to set the userid header directly in case it's not the first api key found, and as an `ApiKeyAuth` in case it is :\

lib/service-gateway/external-auth-gateway/src/main/java/com/wikia/services/externalauth/gateway/ExternalAuthGatewayImpl.java
```diff
@@ -30,6 +31,15 @@ public GoogleApi googleApi() throws GatewayException {
     return new GoogleApi(getApiClient());
   }
 
+  @Override
+  public GoogleApi googleApi(String userId) throws GatewayException {
+    ApiClient apiClient = getApiClient();
+    apiClient.addDefaultHeader("X-Wikia-UserId", userId);
```
drsnyder: Should we use [`SwaggerBundleProvider.GATEWAY_USER_ID_HEADER`](https://github.com/Wikia/pandora/blob/aacfd76fd5f906d006dfa9075f8279cbb315ab1e/lib/swagger/src/main/java/com/wikia/pandora/core/swagger/SwaggerBundleProvider.java#L19)?

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/GoogleRegistrationResource.java
```diff
@@ -0,0 +1,64 @@
...
+
+@Api(
+    basePath = "/",
+    value = "Register and link account to Google",
+    produces = MediaType.APPLICATION_JSON,
+    tags = {"google"}
+)
+@Path("/google")
+@Produces(MediaType.APPLICATION_JSON)
+@RequiredArgsConstructor(onConstructor =@__({@Inject}))
```
drsnyder: What does the `@__({@Inject})` magic do?

sqreek: This is annotation that add annotation to Constructor.
In some moment we will have annotation like `@ServiceThatProductWant` and thats all

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/GoogleRegistrationResource.java
```diff
@@ -0,0 +1,64 @@
...
+  public Response register(
+      @ApiParam(value = "The google access token")
+      @QueryParam("access_token") String accessToken,
+      RegistrationData registrationData) throws Exception {
+
+    try {
+      UserCreateResponse response = registrationService.registerUser(registrationData, accessToken);
+      return Response.ok(response).build(); // TODO: add cookie
+    } catch (GatewayException e) {
+      return StandardResponses.buildResponse(Response.Status.fromStatusCode(e.getStatus()), e.getMessage());
```
sqreek: hmm
proxy inside service looks strange ... 
can we do something with this?

sqreek: Maybe handling by generic GatewayExceptionMapper?

service/user-registration/src/main/java/com/wikia/services/userregistration/service/GoogleRegistrationService.java
```diff
@@ -0,0 +1,52 @@
...
+
+  private final ExternalAuthGateway externalAuthGateway;
+
+  private final AuthenticationGateway authenticationGateway;
+
+  public UserCreateResponse registerUser(
+      RegistrationData data,
+      String googleAccessToken) throws Exception {
+
+    try {
```
sqreek: This try is a little too big

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/GoogleRegistrationResource.java
nmonterroso: From the Lombok docs:

```java
	/**
	 * Any annotations listed here are put on the generated constructor. The syntax for this feature is: {@code @RequiredArgsConstructor(onConstructor=@__({@AnnotationsGoHere}))}
	 */
	AnyAnnotation[] onConstructor() default {};
```

drsnyder: Mind blown.

lib/service-gateway/external-auth-gateway/src/main/java/com/wikia/services/externalauth/gateway/ExternalAuthGatewayImpl.java
nmonterroso: This library is not dependent on dropwizard, so it doesn't have access to that constant. It probably should be a shared constant somewhere that has nothing but constants, since the external service gateway libraries shouldn't be dependent on dropwizard.

service/external-auth/src/main/java/com/wikia/services/externalauth/resources/GoogleAuthResource.java
```diff
@@ -84,7 +84,7 @@ public GoogleAuthResource(
       @ApiResponse(code = 500, message = INTERNAL_SERVER_ERROR_MSG)
   })
   @Timed
-  public Response googleToken(
+  public Response login(
```
ArturKlajnerok: Can this be named 'token'?

nmonterroso: I think `login` is more indicative of what it does. It doesn't only get the token, it sets a cookie which logs you in to wikia.

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/GoogleRegistrationResource.java
nmonterroso: I initially thought the ExceptionMapper interface was part of dropwizard, which is why I didn't do this. Your comment made me dig further though, and it's actually part of the jersey framework. I added a `GatewayExceptionMapper` to `service-gateway-base` in c5ed41d

lib/service-gateway/external-auth-gateway/src/main/java/com/wikia/services/externalauth/gateway/ExternalAuthGatewayImpl.java
drsnyder: Oh, right I see regarding the dependency. It is still greppable so changing it should be straight forward should the need arise.

ArturKlajnerok: Should we then move that to a const in the class ??

lib/service-gateway/external-auth-gateway/src/main/java/com/wikia/services/externalauth/gateway/ExternalAuthGatewayImpl.java
```diff
@@ -30,6 +31,15 @@ public GoogleApi googleApi() throws GatewayException {
...
+    ((ApiKeyAuth) apiClient.getAuthentication("user_id")).setApiKey(userId);
```
ArturKlajnerok: Would be good to move `user_id` to const.

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/GoogleRegistrationResource.java
```diff
@@ -0,0 +1,58 @@
...
+  )
+  @ApiResponses(value = {
+      @ApiResponse(code = 500, message = "internal server error"),
+      @ApiResponse(code = 400, message = "bad request. either registration data is bad or Google access token invalid")
+  })
+  @Timed
+  public Response register(
+      @ApiParam(value = "The google access token")
+      @QueryParam("access_token") String accessToken,
+      RegistrationData registrationData) throws Exception {
```
ArturKlajnerok: Can we also add `QueryParam` support for other params that are part of `RegistrationData` ?

nmonterroso: I think that should be a separate resource that accepts data as `FormParam`

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/GoogleRegistrationResource.java
nmonterroso:
```
 ____________ 
< the future >
 ------------ 
     \
      \
          oO)-.                       .-(Oo
         /__  _\                     /_  __\
         \  \(  |     ()~()         |  )/  /
          \__|\ |    (-___-)        | /|__/
          '  '--'    ==`-'==        '--'  '
```

lib/service-gateway/external-auth-gateway/src/main/java/com/wikia/services/externalauth/gateway/ExternalAuthGatewayImpl.java
nmonterroso: 3a2a1ed

lib/service-gateway/external-auth-gateway/src/main/java/com/wikia/services/externalauth/gateway/ExternalAuthGatewayImpl.java
nmonterroso: 3a2a1ed

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/GoogleRegistrationResource.java
```diff
@@ -0,0 +1,58 @@
...
+
+    UserCreateResponse response = registrationService.registerUser(registrationData, accessToken);
+    return Response.ok(response).build(); // TODO: add cookie
```
nmonterroso: Waiting until cookie creation is centralized before adding this.

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/GoogleRegistrationResource.java
nmonterroso: 4def35e753a343da3250492d1264b5598da618f1

ArturKlajnerok:
Lgtm :+1:

PR #631, opened 2015-11-17
SOC-1480: Changes to avoid slave deadlocks

jimmordino:
We use declarative transactions in the DIS service.  The `readonly` attribute of the `@Transactional` annotation tells the method interceptor whether to route the transaction to the slave or the master, so we need to continue to use these annotations.  However, for readonly transactions we were setting the isolation level to `SERIALIZABLE`, which is generally thought to be the safest isolation level.  According the [MySQL docs](https://dev.mysql.com/doc/refman/5.7/en/innodb-locks-set.html), using this level with InnoDB can lock index gaps:

> SELECT ... FROM is a consistent read, reading a snapshot of the database and setting no locks unless the transaction isolation level is set to SERIALIZABLE. For SERIALIZABLE level, the search sets shared next-key locks on the index records it encounters.

These locks may well be what's causing deadlocks on the slave while it attempts to replicate master transactions.  To avoid such locks, this pull request makes the following changes:

- [x] Change all readonly transactions to use isolation level `READ_COMMITTED` instead of `SERIALIZABLE`, which should not lock anything on the slave
- [x] Allow `autocommit` on the slave, which should be safe because there are no writes sent to the slave by the service.  If the driver does not issue an explicit `BEGIN` or `START TRANSACTION` command then all resources should always be released at the end of every statement.

@jogura, @armonr


jimmordino:
Tests pass on my Mac and my devbox, but always time out on Jenkins.

jogura:
LGTM

PR #633, opened 2015-11-17
Services-991: helios auth for external user registration

harnash:
Adding Helios token to the user-registration response.

Additionally using JFairy for fuzzy test data generation.
@Wikia/services-team


service/user-registration/src/test/java/com/wikia/services/userregistration/service/RegistrationServiceTest.java
```diff
@@ -1,26 +1,26 @@
 package com.wikia.services.userregistration.service;
 
-import static com.wikia.services.userregistration.helpers.TestHelper.REGISTRATION_DATA;
-import static com.wikia.services.userregistration.helpers.TestHelper.USER;
-import static com.wikia.services.userregistration.helpers.TestHelper.USER_ID;
-import static com.wikia.services.userregistration.helpers.TestHelper.USER_PASSWORD;
-import static com.wikia.services.userregistration.helpers.TestHelper.USER_PASSWORD_HASH;
-import static com.wikia.services.userregistration.helpers.TestHelper.USER_REGISTRATION_TIME_MILLIS;
-import static org.junit.Assert.assertEquals;
+import static groovy.util.GroovyTestCase.assertEquals;
```
drsnyder: Is this the right include?

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/UserRegistrationResource.java
```diff
@@ -20,47 +28,70 @@
...
+              userId), e);
+    }
+
+    Response.ResponseBuilder responseBuilder = Response.ok(createResponseBuilder.build());
+
+    if (cookie.isPresent()) {
+        responseBuilder.cookie(cookie.get());
+    }
+
+    return responseBuilder.build();
```
drsnyder: This function has gotten really long. Can some of the related blocks be extracted out into function? For example, the try catch along with the  `sendConfirmationEmail` can probably be moved to a helper function locally.

drsnyder:
Minor comments, otw LGTM.

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/UserRegistrationResource.java
nmonterroso: the code should move into the `RegistrationService` since Google registration will have to walk through the same codepath.

service/user-registration/src/main/java/com/wikia/services/userregistration/guice/UserRegistrationModule.java
```diff
@@ -20,22 +21,16 @@
 
   @Override
   protected void configure() {
+    install(new AuthenticationGatewayModule());
```
nmonterroso: This is installed by [`UserRegistrationApplication`](https://github.com/Wikia/pandora/blob/master/service/user-registration/src/main/java/com/wikia/services/userregistration/UserRegistrationApplication.java#L42)

service/user-registration/src/main/java/com/wikia/services/userregistration/service/AuthService.java
```diff
@@ -0,0 +1,8 @@
+package com.wikia.services.userregistration.service;
+
+import com.wikia.gateway.GatewayException;
+import com.wikia.services.authentication.gateway.domain.AccessData;
+
+public interface AuthService {
```
nmonterroso: Do we need this interface? This is identical to the [`AuthenticationGateway`](https://github.com/Wikia/pandora/blob/master/lib/service-gateway/authentication-gateway/src/main/java/com/wikia/services/authentication/gateway/domain/AccessData.java) interface. I don't think we need to put another layer of abstraction on top of the existing one that we created, especially since this uses the domain object (`AccessData`) provided by the gateway.

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/UserRegistrationResource.java
```diff
@@ -20,47 +28,70 @@
...
   }
 
-  private NewCookie createCookie(String token) {
+  private NewCookie createCookie(String token, Integer expireTime) {
```
nmonterroso: Should this functionality move to the `AuthenticationGateway`? I think that we'll need a standard way of creating an access token cookie across services.

service/user-registration/src/test/java/com/wikia/services/userregistration/service/RegistrationServiceTest.java
harnash: Fixed in 100cc18

service/user-registration/src/main/java/com/wikia/services/userregistration/service/RegistrationService.java
```diff
@@ -38,18 +42,23 @@
...
+                              UserPreferenceGateway preferenceGateway,
+                              ConfirmationEmailService confirmationEmailService,
+                              AuthenticationGateway authorizationGateway,
+                              RegistrationDataValidator validator
+  ) {
     this.storage = storage;
     this.hashService = hashService;
     this.preferenceGateway = preferenceGateway;
+    this.confirmationEmailService = confirmationEmailService;
+    this.authorizationGateway = authorizationGateway;
```
nmonterroso: we should rename this to `authenticationGateway`, since it's for authentication not authorization

nmonterroso:
one last comment from me otherwise lgtm

nmonterroso:
LGTM

ArturKlajnerok:
Test timed out on Jenkins but are passing locally, LGTM :+1:

PR #634, opened 2015-11-17
SOC-1640 - Consume HttpEntity to close resources for all code paths

jogura:
Experiencing high amounts of ConnectionPoolTimeouts when trying to get user permissions in PROD.

To mitigate this only call user permissions for logged in users and consume the HttpEntity to ensure that resource is cleaned properly. 

Additional information in the ticket.

@armonr @jimmordino @garthwebb @jsutterfield


service/discussion/src/main/java/com/wikia/discussionservice/services/UserPermissionServiceImpl.java
```diff
@@ -51,25 +52,34 @@ public UserPermissionServiceImpl(
...
       httpClient.execute(httpGet);
       response = httpClient.execute(httpGet);
+      responseEntity = response.getEntity();
 
       int statusCode = response.getStatusLine().getStatusCode();
       if (statusCode == Response.Status.NOT_FOUND.getStatusCode()) {
         return Optional.empty();
       } else if (statusCode == Response.Status.OK.getStatusCode()) {
-        return createPermissionsFromResponse(response);
+        return createPermissionsFromResponse(responseEntity);
```
jimmordino: Minor:  I've never been a fan of  early returns in a try block, especially when there's a finally block.  It's too easy when reading the code to ignore the important things that happen below even when you return here.

jimmordino:
LGTM, if you can get the tests to pass

PR #637, opened 2015-11-17
Wire up the anti-spoof update call after the user is registered

drsnyder:
This PR adds the `updateAntiSpoof` call to the `RegistrationService` after the user is created. See also #635.

https://wikia-inc.atlassian.net/browse/SERVICES-996

@Wikia/services-team


ArturKlajnerok:
Lgtm :+1:

PR #638, opened 2015-11-18
[DIS] Allows deletion of the only thread in a given forum

jimmordino:
Ensures that post count is never NULL.

@jogura @armonr

Migration script for personal DBs: [update-stats-procs.sql.txt](https://github.com/Wikia/pandora/files/37624/update-stats-procs.sql.txt)


armonr:
:shipit: 
Thanks @jimmordino

PR #640, opened 2015-11-18
Remove old create-new-service tool

sqreek:
as in description. 
new one name is "create-new-service-tool"
ping @Wikia/services-team


nmonterroso:
LGTM - let's make sure to send out an email to everyone or post in #services that we're making this change. Also should we rename the new tool to `create-new-service`? feels kind of weird to have a task named `:tools:create-new-service-tool`, since "tool" is repeated. Not super important though.

PR #643, opened 2015-11-19
User Permissions - Added roles, fixed swagger return type, added block info

jcellary:
Four tickets in one:
https://wikia-inc.atlassian.net/browse/PLATFORM-1661
https://wikia-inc.atlassian.net/browse/PLATFORM-1660
https://wikia-inc.atlassian.net/browse/PLATFORM-1657
https://wikia-inc.atlassian.net/browse/PLATFORM-1675

- Added roles (for now a single role named staff mapped 1/1 to the staff group)
- Added Block Info to list of permissions response
- Added optional user ip header to include in block checking
- Fixed swagger annotations to correctly specify return model for 200 response

pinging @sqreek


service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
```diff
@@ -45,13 +53,26 @@
...
   @Inject
   public UserPermissionsResource(
-      @Named("user-permissions-service") UserPermissionsService userPermissionsService) {
+      @Named("user-permissions-service") UserPermissionsService userPermissionsService,
+      @Named("user-ip-header") String userIpHeader) {
     this.userPermissionsService = userPermissionsService;
+    this.userIpHeader = userIpHeader;
+
+    groupToRolesMappings = new HashMap();
+    groupToRolesMappings.put("staff", "staff");
```
sqreek: staff can be extracted to const

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
```diff
@@ -45,13 +53,26 @@
...
   }
 
-  private InetAddress parseInetAddress(String userIp) {
+  private Optional<InetAddress> parseInetAddress(String userIp) {
+    if (StringUtils.isEmpty(userIp)) {
```
sqreek: Strings.isNullOrEmpty? or isEmpty is equivalent for this? or you don't need to check is it null

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
```diff
@@ -88,7 +95,7 @@ private UserPermissionCheck createUserPermissionCheckResponse(
         .userId(userPermissionInfo.getUserId())
         .permissionName(requestedPermission)
         .userHasPermission(userHasPermission)
-        .build();
+        .userIsBlocked(userPermissionInfo.isBlocked()).build();
```
sqreek: .build can be in next line

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
```diff
@@ -96,19 +103,35 @@ private UserPermissions createGetUserPermissionsResponse(
 
     return UserPermissions.builder()
         .userId(userPermissionInfo.getUserId())
-        .permissions(userPermissionInfo.getPermissions()).build();
+        .permissions(userPermissionInfo.getPermissions())
+        .roles(mapGroupsToRoles(userPermissionInfo.getGroups()))
+        .userIsBlocked(userPermissionInfo.isBlocked()).build();
```
sqreek: build can be in next line

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
```diff
@@ -96,19 +103,35 @@ private UserPermissions createGetUserPermissionsResponse(
...
+  }
+
+  private Collection<String> mapGroupsToRoles(Collection<String> groups) {
```
sqreek: Why this need Collection?

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
```diff
@@ -96,19 +103,35 @@ private UserPermissions createGetUserPermissionsResponse(
...
+      }
+    }
+
+    return result;
   }
 
   @GET
   @Timed
-  @Path("/user/{userId}/wiki/{wikiId}/permission/{permissionName}/userIp/{userIp}")
+  @Path("/user/{userId}/wiki/{wikiId}/permission/{permissionName}")
```
sqreek: Is this change not break api for some client?
Or this is not in use in this moment?

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
jcellary: I believe there is no such function as isNullOrEmpty, there is only isEmpty and it checks both - if is null and if equals ""

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
```diff
@@ -118,24 +141,21 @@ public Response getUserHasPermission(
       @PathParam("userId") int userId,
       @PathParam("wikiId") int wikiId,
       @PathParam("permissionName") String permissionName,
-      @PathParam("userIp") String userIpAddress) {
+      @Context HttpHeaders headers) {
 
     LOGGER.info(String.format(
-        "Received request getUserHasPermission with user id %d, wiki id %d, permission %s, ip %s",
+        "Received request getUserHasPermission with user id %d, wiki id %d, permission %s",
```
sqreek: Please use markers instead putting variables into message

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
jcellary: What do you mean exactly? Why is it of type Collection and not something else?

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
jcellary: It will, but I spent my morning wandering around the office and asking teams if someone uses it :) It shouldn't cause problems for anyone, user permissions service is not used in production yet apart from the global permission endpoint used by helios

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
jcellary: :+1:

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
jcellary: :+1:

service/user-permissions/src/test/java/com/wikia/services/userpermissions/resources/UserPermissionsResourceTest.java
```diff
@@ -117,11 +113,11 @@ public void testGetUserHasPermissionOrIsBlocked() throws IOException, UserPermis
...
                                                        UserPermissionsServiceException {
     when(USER_PERMISSIONS_SERVICE.getUserPermissions(
-        TEST_WIKI_ID, TEST_USER_ID)).thenReturn(
+        eq(TEST_WIKI_ID), eq(TEST_USER_ID), any())).thenReturn(
         Optional.of(UserPermissionsService.UserPermissionInfo.builder()
-                        .userId(TEST_USER_ID)
-                        .groups(TEST_GROUPS)
-                        .permissions(TEST_PERMISSIONS).build())
+                        .userId(Integer.toString(TEST_USER_ID))
+                                    .groups(TEST_GROUPS)
```
sqreek: formating?

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
sqreek: Strings.isNullOrEmpty, I checked and have same code.

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
sqreek: yes, Collection instead List for example.
It's actually ok, but why? :)

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
sqreek: wandering around the office should be put into our process as a mandatory step before api change :)

service/user-permissions/src/main/java/com/wikia/services/userpermissions/resources/UserPermissionsResource.java
jcellary: I just assumed it's better to be as general as possible - the function doesn't need specifically a list, it just needs something iterable, when operating on streams it's often better to use more general interfaces to avoid casts to list, but in this context it doesn't really change a lot

jcellary:
I added usage of markers. I assumed user id and wiki id should always be string. We'll see if the logs are properly generated in ES.

sqreek:
I talked with paweł about it, and index is create every day.
So if we have problem with field that was index as Number, and should be as string, it can cost us less than a day of logs that using this field. Should not be a problem

jcellary:
good to hear that

sqreek:
LGTM

PR #644, opened 2015-11-19
Add ServiceException into Service-core

sqreek:
As requested in [SERVICES-1014](https://wikia-inc.atlassian.net/browse/SERVICES-1014) this add universal that can replace WebApplicationException, returning status code, description, and details if provided in json.


harnash:
LGTM

PR #646, opened 2015-11-19
Removing unused community data service

ArturKlajnerok:
This service is unused and there is no active development in the core code, it only forces us to keep it compatible with any pandora wide services changes.

I hope there are no objections for removal ?

/cc @Wikia/services-team @kenkouot @garthwebb  @kvas-damian


nmonterroso:
LGTM

garthwebb:
No clue what this service is but looks like Nelson does and is on board.  LGTM

ArturKlajnerok:
@garthwebb As far I know there was some old plan to use it in mercury, but seems no longer valid, so I've attached few more people here just to double check.

PR #649, opened 2015-11-20
SERVICES-1014: handling duplicate key

harnash:
Simple task not so simple solution. To properly handle key duplicates we had to add custom exception and a mapper to properly return 409 status code.
Additionally it there is some not-so-pretty try-catch code which needs to unwrap the proper exception from `DataAccessException` since we want to handle just one specific case.


service/user-registration/src/main/java/com/wikia/services/userregistration/bundles/UserRegistrationExceptionMapperBundle.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class UserRegistrationExceptionMapperBundle<T extends Configuration> implements ConfiguredBundle<T> {
+    @Override
+    public void initialize(Bootstrap<?> bootstrap) {
+      // nothing to initialize in this bundle
+    }
+
+    @Override
+    public void run(T configuration, Environment environment) {
+      environment.jersey().register(JaxRsHalBuilderSupport.class);
```
sqreek: a) this should not be in Exception Mapper
b) this should not be in our service because we are not using Hal here

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/UserRegistrationResource.java
```diff
@@ -36,14 +41,18 @@
   @POST
   @Path("/users")
   @Produces(MediaType.APPLICATION_JSON)
-  public Response users(RegistrationData registrationData) {
+  public Response users(RegistrationData registrationData) throws Exception {
```
sqreek: why Exception?

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/UserRegistrationResource.java
```diff
@@ -36,14 +41,18 @@
...
     UserCreateResponse.UserCreateResponseBuilder createResponseBuilder = UserCreateResponse.builder();
     String userId;
 
     try {
       userId = registrationService.registerUser(registrationData);
-    } catch (Exception e) {
-      throw new WebApplicationException(e);
+    } catch (DataAccessException e) {
```
sqreek: This check catch error and wrap it or put it throu. It wraps only a Datastorage exception, so we should put this wrapper in that level, not in Resource

harnash: Alright. Good point! I've moved it to `UserRegistrationService` since I think this is a better place since putting this in `Storage` layer seems to deep.

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/UserRegistrationResource.java
```diff
@@ -27,6 +27,7 @@
 public class UserRegistrationResource {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(UserRegistrationResource.class);
+  public static final String USER_ALREADY_EXISTS_MSG = "User already exists";
```
sqreek: Is this in use in other place than tests?
If it only in tests remove this from here, and use message from service in tests

harnash: @sqreek done in 77b5b72

sqreek:
LGTM

PR #641, opened 2015-11-18
[SERVICES-993] Validate username with antispoof

ArturKlajnerok:
Added `NotSpoof` annotation for username that validates it with MW AntiSpoof Extension. Also added additional registration resource that works with url query params.

/cc @Wikia/services-team

PR #654, opened 2015-11-20
Services 993 validate username with antispoof pr

sqreek:
ping @ArturKlajnerok, @Wikia/services-team 
This is https://github.com/Wikia/pandora/pull/641 PR with fix tests and resolve merge conflicts


service/user-registration/src/main/java/com/wikia/services/userregistration/gateway/WikiaHelperGateway.java
```diff
@@ -7,6 +7,7 @@
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.utils.URIBuilder;
 import org.apache.http.impl.client.CloseableHttpClient;
+import org.codehaus.jackson.type.TypeReference;
```
ArturKlajnerok: Do we need this import ?

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/UserRegistrationResource.java
```diff
@@ -47,16 +71,65 @@ public Response users(RegistrationData registrationData) throws ValidationExcept
...
+      @FormParam(Fields.EMAIL) String email,
+      @ApiParam(required = true)
+      @FormParam(Fields.PASSWORD) String password,
+      @ApiParam(required = true)
+      @FormParam(Fields.BIRTHDATE) String birthDate,
+      @FormParam(Fields.MARKETING_ALLOWED) String marketingAllowed,
+      @FormParam(Fields.LANGUAGE_CODE) String languageCode) throws Exception {
+
+    Optional.ofNullable(marketingAllowed).orElse(Fields.MARKETING_ALLOWED_DEFAULT);
+    languageCode = languageCode == null ? Fields.LANGUAGE_CODE_DEFAULT : languageCode;
```
ArturKlajnerok: Should we use the same pattern here to be consistent ?

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/UserRegistrationResource.java
```diff
@@ -47,16 +71,65 @@ public Response users(RegistrationData registrationData) throws ValidationExcept
...
+      @ApiParam(required = true)
+      @FormParam(Fields.EMAIL) String email,
+      @ApiParam(required = true)
+      @FormParam(Fields.PASSWORD) String password,
+      @ApiParam(required = true)
+      @FormParam(Fields.BIRTHDATE) String birthDate,
+      @FormParam(Fields.MARKETING_ALLOWED) String marketingAllowed,
+      @FormParam(Fields.LANGUAGE_CODE) String languageCode) throws Exception {
+
+    Optional.ofNullable(marketingAllowed).orElse(Fields.MARKETING_ALLOWED_DEFAULT);
```
nmonterroso: Don't you still have to do the assignment? This is creating the new string but without assigning it it's basically a noop, right?

```java
marketingAllowed = Optional.ofNullable(marketingAllowed).orElse(Fields.MARKETING_ALLOWED_DEFAULT);
```

or am i missing something here?

service/user-registration/src/test/java/com/wikia/services/userregistration/validation/RegistrationDataValidatorTest.java
```diff
@@ -145,6 +158,38 @@ private void assertErrorsExist(List<ValidationViolation> validationViolations, S
     List<String> errorSearch = new ArrayList<>(Arrays.asList(expectedMessages));
     validationViolations.forEach(v -> errorSearch.remove(v.getDescription()));
 
-    assertTrue("missing errors: "+errorSearch.toString(), errorSearch.isEmpty());
+    assertTrue("missing errors: " + errorSearch.toString(), errorSearch.isEmpty());
+  }
+
+  public static class NoAntiSpoofValidationModule extends AbstractModule {
```
nmonterroso: What's the reasoning for adding this module to the test? It seems like a lot of setup code that shouldn't be needed. If it's just to get past the failures when `NotSpoofed.Validator.isValid` runs (which looks like it is, because the mock is hard-coded to always allow any value), then we can/should remove it because in a separate commit @ArturKlajnerok changed the default behavior of `@NotSpoofed`. 

We have separate tests for `@NotSpoofed` and `@NotBlocked` to ensure that they work properly, so I don't think we need to introduce this module to run the tests.

service/user-registration/src/main/java/com/wikia/services/userregistration/gateway/WikiaHelperGateway.java
ArturKlajnerok: Removed in : 0c8a188

service/user-registration/src/main/java/com/wikia/services/userregistration/resources/UserRegistrationResource.java
ArturKlajnerok: Yes, without assignment this doesn't do anything, but I guess we need also need to cover empty string default so I used `StringUtils.isBlank()` for this here: 0fc1edd

service/user-registration/src/test/java/com/wikia/services/userregistration/validation/RegistrationDataValidatorTest.java
ArturKlajnerok: Let's resolve this outside this PR and then we can address it in other PR, I need this merged for now to proceed with other work.

ArturKlajnerok:
From my side LGTM :+1:

PR #655, opened 2015-11-20
MOB-5963 - Handle User Permissions API changes which includes userIsBlocked and roles

jogura:
* Add the BANNED/STAFF roles based on `userIsBlocked` and `roles` from the User Perms Service
* Update the AuthorizationService to throw a ForbiddenException if a banned user attempts to create a thread/post 
* Add unit test to gracefully handle additional API changes in the JSON response
* Add unit test to test AuthorizationService handling of banned users

RFCR @garthwebb @armonr @jsutterfield @jimmordino


service/discussion/src/main/java/com/wikia/discussionservice/domain/ClientUserPermission.java
```diff
@@ -1,8 +1,10 @@
 package com.wikia.discussionservice.domain;
```
armonr: This file can be removed now actually. I created this because there was no way to know block status in the original API point, but now there is.

service/discussion/src/main/java/com/wikia/discussionservice/domain/ClientUserPermissions.java
```diff
@@ -1,10 +1,13 @@
 package com.wikia.discussionservice.domain;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import lombok.Data;
+import lombok.Getter;
 
 import java.util.List;
 
 @Data
+@JsonIgnoreProperties(ignoreUnknown = true)
```
armonr: Is this added to keep things backward compatible?

jogura: Yes, exactly.

service/discussion/src/main/java/com/wikia/discussionservice/auth/DiscussionServiceRealm.java
```diff
@@ -74,25 +76,47 @@ public DiscussionServiceRealm(AuthorizationRoleProvider authorizationRoleProvide
...
-      Optional<ClientUserPermissions> userPermissionsOptional =
-          userPermissionService.getUserPermission(userId, siteId);
+      if (siteId != null && !siteId.isEmpty()) {
+        Optional<ClientUserPermissions> userPermissionsOptional =
+            userPermissionService.getUserPermission(userId, siteId);
 
-      userPermissionsOptional.ifPresent(userPermissions ->
-                                            localPermissions.addAll(
-                                                userPermissions.getPermissions()));
+        userPermissions = userPermissionsOptional.get();
```
garthwebb: AFAIK a `get()` will throw `NoSuchElementException` if there is no value present.  Maybe it would be better to carry this Optional down to L92, do an `isPresent` there and then do the `get()` once its known there is a value?

service/discussion/src/main/java/com/wikia/discussionservice/services/AuthorizationService.java
```diff
@@ -63,6 +63,13 @@ private AuthorizationService() {
   };
 
   public static final Authorizable authorizeCreateThread = (user, siteId, forumId) -> {
+    if (user.isBanned()) {
+      LOGGER.info(
+          String.format("Banned user %s attempt to create thread in forum %s, site %s denied",
+                        user.getUserId(), forumId, siteId));
+      throw new ForbiddenException("This user is not authorized to create the thread.");
```
armonr: Could this be a more generic `assertNotBlocked()` to be called from all `authorizeCreate*`, `authorizeUpdate*`, `authorizeDelete*`, and any other write action check?

service/discussion/src/main/java/com/wikia/discussionservice/auth/DiscussionServiceRealm.java
jogura: You're right, plus early return caused other unforeseen side-effects e14b907

armonr:
LGTM, will address anything else in another PR

service/discussion/src/main/java/com/wikia/discussionservice/auth/DiscussionServiceRealm.java
```diff
@@ -74,25 +76,44 @@ public DiscussionServiceRealm(AuthorizationRoleProvider authorizationRoleProvide
...
+          userPermissions = userPermissionsOptional.get();
+        }
+      }
+    }
 
+    localPermissions.addAll(userPermissions.getPermissions());
+    if (userPermissions.isUserIsBlocked()) {
+      // Return early if the user is blocked with the role of BANNED
+      userInfo.put(ServiceRealm.PERMISSIONS_KEY, new HashSet<>());
+      userInfo.put(ServiceRealm.ROLES_KEY, Sets.newHashSet(Role.BANNED.toString()));
```
jsutterfield: Wha are thoughts on updating the enum `Role.BANNED` to `Role.BLOCKED`? That seems to be the term used elsewhere (in MW and in the response from the Permission service). It would be one less mental translation to worry about.

service/discussion/src/main/java/com/wikia/discussionservice/domain/ClientUserPermissions.java
```diff
@@ -1,10 +1,14 @@
 package com.wikia.discussionservice.domain;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.google.common.collect.Lists;
 import lombok.Data;
+import lombok.Getter;
```
jsutterfield: Looks like this isn't being used?

service/discussion/src/main/java/com/wikia/discussionservice/auth/DiscussionServiceRealm.java
armonr: Yes, I will prepare a another PR for that.

armonr: #656

PR #659, opened 2015-11-23
[PLATFORM-1609] Added possiblity to overwrite mw domain

jcellary:
When deploying instance of user-permission to dev-mesos I needed to create a dedicated mw dev instance. Unfortunately the dev instance when resolving wiki id to domain returns a production domain, however we want to continue using our dev box for further api calls. Because of that I added the possiblity to force a dev domain in place of standard wikia.com through configuration.

Pinging @sqreek


sqreek:
after offline talk:
add provider for wiki url's

jcellary:
Refactored, pls take another look

jcellary:
Per offline conversation merging as is and we can introduce changes to guice modules in a separate PR.

PR #662, opened 2015-11-24
Concept of binding proper guice module depends on environment

sqreek:
For this moment this is a concept, not a solution.

Problem:
User-permision work different on dev, and prod.
dev user-permision after resolve URI for wiki, it always get production one.
for dev/test purpouse its need to change this uri for dev one.
Code for handling this issue need to be put in some place, and a question is where.
* It can be resolved on the lowest level as a big IF environment == ... :-1: 
* it can be handled as provider that contains logic to provide proper wrapper :+1: :-1: (guice says that logic in modules is not so good, didn't say anything about providers)
* Create separate application for dev with different guice module (so this is actualy recomended solution, but I cannot see that in practice) :+1: :-1: 
d) Solution Below, that scan package for Modules and use only proper for environment
It creates  another level of magic so I don;t like it so much, but it is some kind of solution for a problem  :+1: :-1: 

What do you think? @pchojnacki @jcellary @Wikia/services-team


lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/MagicPandoraGuiceBundle.java
```diff
@@ -0,0 +1,58 @@
...
+import ru.vyarus.dropwizard.guice.module.installer.scanner.ClassVisitor;
+import ru.vyarus.dropwizard.guice.module.installer.scanner.ClasspathScanner;
+
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Consumer;
+
+public class MagicPandoraGuiceBundle<T extends Configuration> extends PandoraGuiceBundle<T> {
```
jcellary: Not sure if the word "Magic" is very descriptive :)

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/MagicPandoraGuiceBundle.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(MagicPandoraGuiceBundle.class);
+
+  public MagicPandoraGuiceBundle(Class applicationClass, Consumer<GuiceBundle.Builder> beforeBuild) {
+    super(applicationClass, (b -> {
```
jcellary: Formatting seems to be of, where does this function end? I believe such long lambda functions are not very readable, I would create a normal method for this.

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/MagicPandoraGuiceBundle.java
```diff
@@ -0,0 +1,58 @@
...
+          b.modules(
+              modules.stream()
+                  .filter(module -> Optional.ofNullable(module.getAnnotation(Environment.class))
+                      .map(x -> Objects.equals(x.value(), System.getenv("ENVIRONMENT")))
+                      .orElse(false))
+                  .map(module -> {
+                    try {
+                      return module.newInstance();
+                    } catch (InstantiationException | IllegalAccessException e) {
+                      LOGGER.error("magic doesn't work");
```
jcellary: Not very informative :)

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/MagicPandoraGuiceBundle.java
```diff
@@ -0,0 +1,58 @@
...
+            if (Module.class.isAssignableFrom(type)) {
+              modules.add((Class<? extends Module>) type);
+            }
+          });
+//          Reflections reflections = new Reflections(applicationClass.getPackage().getName());
+//          Set<Class<? extends Module>> modules = reflections.getSubTypesOf(Module.class);
+
+          b.modules(
+              modules.stream()
+                  .filter(module -> Optional.ofNullable(module.getAnnotation(Environment.class))
```
jcellary: There's no default fallback. I would recommend using prod if there's no matching module. At least that's how it was done in the past in mobile applications

lib/dropwizard/consul/src/main/java/com/wikia/dropwizard/consul/MagicPandoraGuiceBundle.java
```diff
@@ -0,0 +1,58 @@
...
+                      .map(x -> Objects.equals(x.value(), System.getenv("ENVIRONMENT")))
+                      .orElse(false))
+                  .map(module -> {
```
jcellary: Also too much logic nested into this stream processing imho, I would split that for readability's sake

jcellary:
I don't have any strong feelings, I think it's ok to do it this way. The code itself needs refinement before going to production, but I'm assuming that's because it's just a proof of concept.

pchojnacki:
Will all modules now be autoloadeed? 
Will you propagate this change to all services? So the repo is not left in two states like with ConsulGuiceApplication.

What about modules  I don't want to autoload?

pchojnacki:
Generally
1 I would avoid autoloading modules.
2. You can bind different implementations based on config

I think the concept of different modules bound based on environment hasn't worked that well for mobile-configurations. Maybe we could try something else like 2.

sqreek:
@pchojnacki:
In this state all modules with proper annotation is loaded.
For this moment I will not propagate it anywhere.
I agree that magic is not so good
about Mobile-(config(uration)|aplication) there is still stinky code that I want to remove it, and looking for replacement.

If this is in provider, and you have to create more than one environment specific bindings, it will have a lot of duplicate logic in code. 
Combine it into module seems more clean.

pchojnacki:
I don't think you have to create modules for each environment
its enough if you bind different classes based on configuration options
This imho would be more generic and not constrained to artificially named environment.

PR #665, opened 2015-11-25
CE-3095: Add possibility to prioritize classification by provider

lukaszkonieczny:
Add possibility to prioritize template classification by provider

@idradm @sqreek


service/template-classification-storage/src/main/java/com/wikia/services/templateclassificationstorage/domain/TemplateTypeProvider.java
```diff
@@ -19,6 +19,14 @@
 @NoArgsConstructor
 public class TemplateTypeProvider {
 
+  public static final String USER_PROVIDER_NAME = "user";
+  public static final String AUTO_PROVIDER_NAME = "auto_type_matcher";
+  public static final String USAGE_PROVIDER_NAME = "usage_classifier";
+
+  public static final int USER_PROVIDER_PRIORITY = 10;
+  public static final int AUTO_PROVIDER_PRIORITY = 5;
+  public static final int USAGE_PROVIDER_PRIORITY = 2;
```
idradm: In that case shouldn't provider be a new resource? Strongly typed and stored in db?

sqreek: Template provider is only one place that will talk to db with template classifications.
So we don't need to store this there.
Also we don't need to put gap between values, because we can change it everytime. (beacuse it's not stored in DB).

when we got a case where we don't know provider value, and we get it from outside in request. then we can think about put this value in db

service/template-classification-storage/src/test/java/com/wikia/services/templateclassificationstorage/helper/IntegrationTestsHelper.java
```diff
@@ -58,6 +60,11 @@ public static TemplateTypeProvider getAutomaticTemplateTypeProvider(Set<String>
         .types(templateType).build();
   }
 
+  public static TemplateTypeProvider getUsageTemplateTypeProvider(Set<String> templateType) {
```
sqreek: Where this is in use?
And I mean in production code, not in tests?
it seems to be only in tests so put it there

lukaszkonieczny: I think it is in tests :)

idradm:
let it go :snowflake:

sqreek:
https://github.com/Wikia/pandora/pull/665#discussion-diff-45894166R63 this should be moved into test before merge @idradm

sqreek:
nevermind. I'm blind. everything is ok ! :+1: :see_no_evil:

PR #666, opened 2015-11-25
Minor fix in static-assets config and added validation in dc-file-sync

jcellary:
Renamed SJC_CEPH_* to LOCAL_CEPH_* to be able to deploy static assets to reston. Also added one check in dc-file-sync to avoid copying files from one place to the same one.

Pinging @sqreek


sqreek:
LG

PR #667, opened 2015-11-25
Update some swagger annotations for the notification service

jsutterfield:


service/notification/src/main/java/com/wikia/services/notification/resources/NotificationResource.java
```diff
@@ -48,13 +48,11 @@ public NotificationResource(ParseClient parseClient, ParseKeyService parseKeySer
...
   @Path("/notify")
   @Timed
-  @ApiOperation(value = "Sends a notification to a user", notes = "Not Implemented")
+  @ApiOperation(value = "Sends a notification to a user")
   @ApiResponses(value = {
       @ApiResponse(code = 202, message = "Successfully sent a notification",
           response = Representation.class),
       @ApiResponse(code = 400, message = "Bad request", response = Problem.class),
-      @ApiResponse(code = 401, message = "Unauthorized", response = Problem.class),
-      @ApiResponse(code = 403, message = "Forbidden", response = Problem.class),
```
jsutterfield: We currently don't return these responses, we can add these back when we add auth.

service/notification/src/main/java/com/wikia/services/notification/domain/Notification.java
```diff
@@ -24,27 +24,34 @@
...
+  @ApiModelProperty(hidden = true)
   private Integer objectId;
 
-  @JsonProperty
+  @JsonIgnoreProperties
+  @ApiModelProperty(hidden = true)
   private String objectType;
 
-  @JsonProperty
+  @JsonIgnoreProperties
```
armonr: These should be `JsonIgnore`

PR #669, opened 2015-11-25
SOC-1513 Gateway for notification service

armonr:
This is a gateway library for notification service. It exposes one API point to other services.

/CC @garthwebb @jsutterfield 

https://wikia-inc.atlassian.net/browse/SOC-1513


lib/service-gateway/notification-gateway/src/main/java/com/wikia/notification/gateway/NotificationGatewayImpl.java
```diff
@@ -0,0 +1,62 @@
...
+import com.wikia.services.notification.client.model.Notification;
+import com.wikia.services.notification.client.model.Representation;
+
+import com.google.inject.Inject;
+
+public class NotificationGatewayImpl implements NotificationGateway {
+
+  public static final String SERVICE_NAME = "notification";
+
+  private static final String AUTH_HEADER_NAME = "X-Wikia-UserId";
```
jsutterfield: We're not actually using the X-Wikia-UserId header right now. Can we remove it?

garthwebb:
LGTM

PR #671, opened 2015-11-26
Move logic about default marketing and language

sqreek:
Registration Data should has this logic, not external class.


pchojnacki:
LGTM

PR #672, opened 2015-11-26
in user registration change logs data from message to markers

sqreek:
as in PR name


service/user-registration/src/main/java/com/wikia/services/userregistration/gateway/WikiaHelperGateway.java
```diff
@@ -1,5 +1,7 @@
 package com.wikia.services.userregistration.gateway;
 
+import static net.logstash.logback.marker.Markers.append;
```
harnash: Please unify the imports. Sometimes it is `net.logstash.logback.marker.Markers.append` and sometimes `net.logstash.logback.marker.Markers`

harnash:
LGTM

PR #673, opened 2015-11-26
add Logger bundle that add user_id for request that contains authHeader

sqreek:
using MDC we can add user_id from auth header to every log in current thread.
Remember that creating new thread inside request process will omit this functionality(different thread, different MDC)

ping @Wikia/services-team @pchojnacki


pchojnacki:
Have you tested it under latest Jetty ?
As I mentioned to you in private conversation recently Jetty went fully async. Which means that your request can be paused in some conditions and threads can change/start handling different requests at the same time.

Have you tried:
https://github.com/google/guice/wiki/Scopes

sqreek:
I tested on dev env, and put some traffic there.
Yes, we need to check how it works on new jetty, but I don't think that this will change a lot. because if it is, many filter will fail also.
Filter is register once per application, and run in every request by doFiltering, so I don't know how guice scopes can help it here

PR #674, opened 2015-11-27
CE-3095: Filter other type on classification

lukaszkonieczny:
ping @idradm @sqreek


service/template-classification-storage/src/main/java/com/wikia/services/templateclassificationstorage/classification/UserOverrideClassification.java
```diff
@@ -13,8 +13,9 @@
   @Override
   public String classify(TemplateClassificationData classificationData) {
     return classificationData.getProviders().stream()
-        // filter providers containing more than one type
-        .filter(provider -> provider.getTypes().size() == 1)
+        // filter providers containing more than one type or type other
+        .filter(provider -> provider.getTypes().size() == 1
+                            && !provider.getTypes().contains(OTHER_TYPE))
```
idradm: You can add separate filter method for that, should look nicer

idradm:
What about unit tests for logic? Could you add a case for what you are changing?

service/template-classification-storage/src/main/java/com/wikia/services/templateclassificationstorage/classification/UserOverrideClassification.java
sqreek: :+1:

idradm:
awesomely awesome!

PR #677, opened 2015-11-30
SERVICES-1008 In registration service add Async Calls

sqreek:
Async calls to:
- SendConfirmation email
- Update AntiSpoof
- Set user preference

Also refactored WikiaHelperGateway

ping @Wikia/services-team


service/user-registration/src/main/java/com/wikia/services/userregistration/gateway/WikiaHelperGateway.java
```diff
@@ -32,127 +36,107 @@
...
-      if (payload.containsKey(ALLOW_KEY)) {
-        return payload.get(ALLOW_KEY);
-      }
-    } catch (IOException e) {
-      LOGGER.error(REQUEST_FAILED_MSG, e);
-      throw new GatewayException(e.getMessage());
+      return requestPost(METHOD_CHECK_SPOOF, name).allow;
+    } catch (GatewayException e) {
+      LOGGER.error(Markers.append("username", name), "Error while checking antispoof", e);
+      return false;
```
sqreek: Is this should be a true when Gateway has some problems?

service/user-registration/src/main/java/com/wikia/services/userregistration/gateway/WikiaHelperGateway.java
```diff
@@ -32,127 +36,107 @@
...
-    );
+  @Data
+  public static class WikiaHelperResponse {
+    @JsonProperty
+    private Boolean allow;
+    @JsonProperty
+    private String message;
+    @JsonProperty
+    private Boolean success;
   }
```
drsnyder: I know it's trivial, but I think this should be in a separate file.

service/user-registration/src/main/java/com/wikia/services/userregistration/service/RegistrationService.java
```diff
@@ -93,13 +95,20 @@ public String registerUser(RegistrationData registrationData) throws ValidationE
...
   }
 
+  private void postRegistrationCommands(String userId,
+                                        String username,
+                                        RegistrationData registrationData) {
+
+    executorService.submit(() -> setUserPreferences(userId, registrationData));
+    executorService.submit(() -> sendConfirmationEmail(username, registrationData.getLangCode()));
+    executorService.submit(() -> updateAntiSpoof(username));
+  }
```
drsnyder: What happens if one or more of these throws an exception?

service/user-registration/src/main/java/com/wikia/services/userregistration/gateway/WikiaHelperGateway.java
sqreek: It's like 50/50 points for keep it here or in separate file.
If none will make vote on it, I will put it in separate file

service/user-registration/src/main/java/com/wikia/services/userregistration/service/RegistrationService.java
sqreek: Exactly nothing :)
So this is a little tricky.
Expected exceptions are handled inside methods, so we don't need to worry about them.

ArturKlajnerok:
Lgtm :+1:

PR #678, opened 2015-11-30
SOC-1646 Tenants for new communities

armonr:
/CC @jimmordino @jsutterfield 

https://wikia-inc.atlassian.net/browse/SOC-1646


jimmordino:
Can we do some shard testing before we enable these sites?  If we open up this many sites without a backup plan we are asking for trouble.

armonr:
@jimmordino: Good point! I will work on setting up another one and see how things work.

jsutterfield:
LGTM

armonr:
Per [OPS-7722](https://wikia-inc.atlassian.net/browse/OPS-7722), let's add these until we reach a higher load level. We'll still prepare for multiple shards in the meantime.

PR #679, opened 2015-12-01
in MobileConfigurations remove wrong pattern about guice module

sqreek:
https://wikia-inc.atlassian.net/browse/SERVICES-1030
ping @Wikia/services-team


jogura:
LGTM, thanks!

service/mobile-applications/src/main/java/com/wikia/services/mobileapplications/guicemodule/MobileApplicationsModule.java
```diff
@@ -0,0 +1,87 @@
...
+        "vignetteimage_set");
+  }
+
+  @Provides
+  public AppsListService provideAppListService(
+      MobileApplicationsConfiguration configuration,
+      @Named("apps-deployer-list-service") CloseableHttpClient httpClient,
+      @Named("mobile-applications-service-apps-list") CephGateway cephGateway
+  ) {
+    String stage = System.getenv(ENVIRONMENT);
```
pchojnacki: I think it should be renamed to Environment, at least it would look more consistent. I think.

pchojnacki:
I minor (ignorable) comment Otherwise LGTM

PR #681, opened 2015-12-01
Grafana Dashboard Generation

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-1018

Adds a new `grafanaDashbaord` gradle task to each service that will create a dashboard in grafana, with count and timing information for every resource. This utilizes Dropwizard [commands](http://www.dropwizard.io/0.8.1/docs/manual/core.html#commands) to spin up an instance of the service (without HTTP functionality) to inspect the available resources and generate valid Grafana dashboard JSON. It then sends a `PUT` request to Grafana to create or (with the overwrite flag set) update the dashboard.


ArturKlajnerok:
Lgtm :+1:

lib/dropwizard/grafana-dashboard-generation/src/main/java/com/wikia/grafana/dashboardgeneration/domain/Resource.java
```diff
@@ -0,0 +1,91 @@
...
+          case SUB_RESOURCE_LOCATOR:
+            addMethods(builder, path, method.getInvocable().getRawResponseType(), true);
+            break;
+        }
+      }
+    }
+  }
+
+  private static String normalizePath(String basePath, String path) {
+    if (basePath.endsWith("/")) {
```
sqreek: This should be changed to use `Paths.get()` for normalize

lib/dropwizard/grafana-dashboard-generation/src/main/java/com/wikia/grafana/dashboardgeneration/util/Uploader.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  private boolean dashboardExists() {
+    Response response = ClientBuilder
+        .newClient()
+        .target(getDashboardUrl())
+        .request()
+        .get();
+
+    return response.getStatus() == HttpStatus.OK_200;
```
sqreek: This actually says not about dashboardexist, but about I can connect and dashboard exist.
I can connect can make some difference sometimes

sqreek:
What about monitoring custom metrics? Can we easy extend this script?
For example: InstrumentedServiceExecutor for monitoring threads

nmonterroso:
If a service has custom metrics they want to have on their dashboard, they'll have to add that to their individual dashboards. If there's something universal that every service would want, then it can be added to the dashboard json template.

PR #683, opened 2015-12-02
Added timestamp to username while generating test user data

ludwikkazmierczak:
Added timestamp to username while generating test user data. This one is needed for our test to work stable, cause otherwise names are chosen from a set, and with each test execution there is a bigger chance of collision.


sqreek:
LGTM, remember to put ping @ Wikia/services-team next time

PR #685, opened 2015-12-02
refactor sonar properties

sqreek:
Added missing projects and remove not existing once
ping @Wikia/services-team


nmonterroso:
LGTM - what's the reasoning in the new order of the projects in `settings.gradle`? it doesn't seem to be alphabetical

sqreek:
It is alphabetic except io-swagger.

PR #686, opened 2015-12-02
Only fetch distinct wiki ids

drsnyder:
Only fetch distinct wiki ids in `findWikisWithLocalPreferenceValue`. This change replicates the behavior that [previously existed](https://github.com/Wikia/app/commit/9df2b1008cff36907d0d7cbe2e626d7d0594794f#diff-77c4dcf675364821f0c6836036116805L57) in MW.

https://wikia-inc.atlassian.net/browse/SERVICES-1010

@Wikia/services-team


nmonterroso:
can we add some tests to make sure that we don't return duplicateS? other than that lgtm

